PR #3, opened 2020-01-31
Proposal for refactored command support

davidfestal:
Signed-off-by: David Festal <dfestal@redhat.com>

### What does this PR do?

This PR adds a proposal a refactored design of command syntax, leveraging what the devfile `command`s were, but trying to fix a number of identified problems like:
- limitations in the existing devfile syntax,
- inconsistencies between commands in the existing devfile and commands in the meta.yaml syntax of the che plugin registry.


pkg/apis/workspaces/v1alpha1/commands.go
```diff
@@ -1,16 +1,114 @@
...
+import runtime "k8s.io/apimachinery/pkg/runtime"
+
+// CommandType describes the type of command.
+// Only one of the following command type may be specified.
+// +kubebuilder:validation:Enum=Exec;VscodeTask;VscodeLaunch;Custom
+type CommandType string
+
+const (
+	ExecCommandType         CommandType = "Exec"
+	VscodeTaskCommandType   CommandType = "VscodeTask"
```
sleshchenko: Also, it makes sense to consider renaming constant in upper case way, like `VSCODE_TASK_COMMAND` or `VSCODE_TASK_COMMAND_TYPE`
https://golang.org/doc/effective_go.html#constants

devfiles/simple-devfile.yaml
```diff
@@ -4,11 +4,25 @@ projects:
...
         branch: "master"
 commands:
-  - name: Create json schema from Go API
-    actions:
-      - type: exec
-        command: "./buildSchema.sh"
-        component: build-tools
+  - exec:
+      commandLine: "./buildSchema.sh"
+      component: build-tools
```
sleshchenko: Well, a component may have multiple containers.
Maybe it makes sense to consider improving format to solve this issue. Maybe it would be better/easier at this stage just to postpone it.

pkg/apis/workspaces/v1alpha1/commands.go
```diff
@@ -1,16 +1,114 @@
...
+	Url string `json:"url,omitempty"`
+
+	// Embedded content of the vscode configuration file
+	// +optional
+	Inlined string `json:"inlined,omitempty"`
+}
+
+type VscodeConfigurationCommand struct {
+	BaseCommand `json:",inline"`
+	Location    VscodeConfigurationCommandLocation `json:",inline"`
```
sleshchenko:
```suggestion
	Location    VscodeConfigurationCommandLocation `json:"location,inline"`
```

Did you miss json field name by purpose?

sleshchenko: Is location the same as `reference` which we have in the current Devfile model?

Issue #31, opened 2020-03-24
Better plugin mechanism


l0rd:
We want to replace `chePlugin` and `cheEditor` with a  new `plugin` component type that:
- can be specified using the devfile syntax (opposed to 1.0.0 `meta.yaml`, a devfile-like format) 
- is not Che specific, plugins can be specified for `odo` or even desktop IDEs 

### Use cases

These are the use cases of devfile plugins (basically Che use cases):

- Reference one or more vs code extensions and the sidecar container that will host them (e.g. the [java plugin](https://github.com/eclipse/che-plugin-registry/blob/master/v3/plugins/redhat/java/0.57.0/meta.yaml))
- Reference a container component that is re-used in multiple devfiles/stacks ([che-theia is an example](https://github.com/eclipse/che-plugin-registry/blob/master/v3/plugins/eclipse/che-theia/7.10.0/meta.yaml)) 

### Proposal

```yaml
devfile.yaml
---
components:
  - plugin:
      name: 'java language server'
      id: redhat/dependency-analytics/latest  # <= reference plugins published
                                              #    in a plugin registry
      registry: https://plugin-registry.io/
  - plugin:
      name: 'k8s workspace plugin'
      kubernetes:
        namespace: mynamespace
        selector: mykey=myvalue               # <== reference a kubernetes
                                              #     devworkspacetemplate resource
  - plugin:
      name: ...
      uri: https://..../java8-plugin.yaml     # <== reference external raw devfile
      components:                             # <== customise plugin's configuration
        - kubernetes:
            name: ...
            reference: ...
        - container:
            name: vscode-java
            memoryLimit: 2Gi
        - volume:
            name: m2
            size: 2G
```

```yaml
java8-plugin.yaml
---
schemaVersion: 2.0.0
metadata:
  publisher: redhat
  name: java8
  version: 0.57.0
  displayName: Language Support for Java 8
  title: Language Support for Java(TM) by Red Hat
  description: Java Linting, Intellisense, (...)
  icon: https://(..)/logo.svg
  repository: https://github.com/redhat-developer/vscode-java
  category: Language
  firstPublicationDate: "2020-02-20"
  pluginType: che-theia-vsx    # <== mandatory for plugins. Valid types
                               #     are che-theia-vsx, che-editor, che-theia-plugin,
                               #     che-theia-extension, generic-service and 
                               #     generic-ui
parent:
  id: redhat/theia-vsx-stack/latest
  components:
    - container:
[...]

PR #41, opened 2020-04-24
Implement proposal #31 for better plugin mechanism and cleanup Json schema

davidfestal:
### What does this PR do?

This PR implements in the proposal #31 that was agreed on,
especially the overriding mechanism for parent devfiles and plugins
based on strategic merge patch (a-la-kustomize).

This PR also enhances the build mechanism that derives JSON schemas from
the generated K8S CRD OpenApi schema. Especially it:
- removes some K8S-specific fields (union discriminators and kubernetes extensions)
- forbids additional non-validated properties everywhere, unless it is explicitly requested by the source model
- makes required the fields that should be mandatory but are set optional in the K8S model in order to allow easier overriding 

As part of this enhancement, this PR also upgrades the version of the `operator-sdk` CLI tools used to generate the CRDs to `0.17.0`
 
This PR implements issue #31

Yes


davidfestal:
After confirmation from @l0rd, let me merge this PR and provide remaining changes in new PRs.

PR #81, opened 2020-06-16
Fix the devfile to bypass a GO LS error...

davidfestal:
### What does this PR do?

Fix the devfile to bypass a GO LS error that prevents the IDE loading when preferences are provided.

Also  update the global theia settings.json file only when needed.

No related issue, but this is a followup of PR https://github.com/devfile/kubernetes-api/pull/79

Yes, on che.openshift.io


devfile.yaml
```diff
@@ -27,14 +23,14 @@ components:
       - '-c'
       - >-
         while [ ! -f
-        ${CHE_PROJECTS_ROOT}/src/github.com/devfile/kubernetes-api/.theia/settings.json
-        -o  ! -f ${CHE_PROJECTS_ROOT}/.theia/settings.json ]; do echo "Waiting
-        for the clone end" ; sleep 2; done; sleep 2 ; echo "Updating the yaml
-        schema bindings" ; sed -e
+        ${CHE_PROJECTS_ROOT}/src/github.com/devfile/kubernetes-api/.theia/settings.json ];
```
sleshchenko, line 26: Can we remove this busy box container by moving this command to tools container?

Issue #33, opened 2020-03-24
Containers endpoints (routes/ingresses)


l0rd:
Current devfile 1.0 container.endpoints syntax ([link to the doc](https://www.eclipse.org/che/docs/che-7/making-a-workspace-portable-using-a-devfile/#component-type-dockerimage_making-a-workspace-portable-using-a-devfile)):

```yaml
components:
  - alias: maven
    type: dockerimage
    image: eclipe/maven-jdk8:latest
    endpoints:
      - name: maven-server
        port: 3101
        attributes:
          protocol: http
          secure: 'true'
          public: 'true'
          discoverable: 'false'
```

Current syntax of che plugins ([link to the doc](https://github.com/eclipse/che-plugin-registry#plugin-meta-yaml-structure)):

```yaml
spec:
  endpoints:
   -  name: "theia"
      public: true
      targetPort: 3100
      attributes:
        protocol: http
        type: ide
        secure: true
        cookiesAuthEnabled: true
        discoverable: false
  containers:
   - name: theia-ide
     image: "quay.io/eclipse/che-theia:7.10.0"
     ports:
         - exposedPort: 3100
```

l0rd:
cc @davidfestal @kadel

davidfestal:
In the existing initial proposal of devfile 2.0, I already tried to harmonize these differences, based on the definitions in the Che plugins `meta.yaml` schema, which seems to be more precise and complete, especially providing real `boolean` fields, instead of non-validated strings for everything.

I had also added a slight change (mentioned below), to distinguish URL scheme protocol (`http`, `https`, `ws`...), which I renamed `scheme` and the transport protocol (`tcp`, `udp`, ...) which I named `protocol`.

In K8S, container port protocol is the fact the transport protocol.

However this is subject to approval / alternate proposals. For example to maintain compatibility with the 1.0 definition, maybe we could also have `protocol` and `transportProtocol` ? 
``` yaml
          endpoints:
            - name: term-websockets
              targetPort: 4000
              configuration:    <-- change from the devfile 1.0, to separate the schema-validated configuration from additional free-form *attributes*
                cookiesAuthEnabled: false
                protocol: tcp     <--- change from the dev 1.0, to distinguish between URI protocol and transport protocol 
                scheme: ws     <--- change from the dev 1.0, to distinguish between URI protocol and transport protocol 
                type: terminal
                discoverable: false
                path: "/apath"
[...]

PR #84, opened 2020-06-19
Finalize endpoints

davidfestal:
### What does this PR do?

This PR implements the agreement about how container endpoints are managed in Devfile 2.0, as expressed in this comment: https://github.com/devfile/kubernetes-api/issues/33#issuecomment-629345779

https://github.com/devfile/kubernetes-api/issues/33#issuecomment-629345779

Yes, in `che.openshift.io`


davidfestal:
I have a doubt about the `path` field.
In the comment I followed (https://github.com/devfile/kubernetes-api/issues/33#issuecomment-629345779), it appears as a real specified field, but I seem to remember that at some point we had somehow discussed about putting it into the implementation-dependent `attributes`
@kadel @l0rd do you remember ?

I assume it would make sense since the scheme (also used to build the final URL), is already in the implementation-dependent attributes.

davidfestal:
Another question: in devfile 1.0, it seems endpoints are also available for other compoents, mainly kubernetes and openshift. Is it something we also want in devfile 2.0 ?
@l0rd wdyt ?

l0rd:
> In the comment I followed ([#33 (comment)](https://github.com/devfile/kubernetes-api/issues/33#issuecomment-629345779)), it appears as a real specified field, but I seem to remember that at some point we had somehow discussed about putting it into the implementation-dependent `attributes`
> @kadel @l0rd do you remember ?

@davidfestal sorry but I don't remember. And I would rather use the format specified in the comment you have linked: `path` is in the ingress spec already, `scheme` is not. Another consideration is that we may merge `schema` and `protocol` as `http` or `ws` will always translate to `tpc` in the corresponding k8s ingress no?



> Another question: in devfile 1.0, it seems endpoints are also available for other compoents, mainly kubernetes and openshift. Is it something we also want in devfile 2.0 ?
> @l0rd wdyt ?

That would be useful yes.

davidfestal:
> > In the comment I followed ([#33 (comment)](https://github.com/devfile/kubernetes-api/issues/33#issuecomment-629345779)), it appears as a real specified field, but I seem to remember that at some point we had somehow discussed about putting it into the implementation-dependent `attributes`
> > @kadel @l0rd do you remember ?
> 
> @davidfestal sorry but I don't remember. And I would rather use the format specified in the comment you have linked: `path` is in the ingress spec already, `scheme` is not.

OK, so maybe I mis-remembered. Let's keep it in the Spec part.

> Another consideration is that we may merge `scheme` and `protocol` as `http` or `ws` will always translate to `tpc` in the corresponding k8s ingress no?
> 

I like this idea

> > Another question: in devfile 1.0, it seems endpoints are also available for other compoents, mainly kubernetes and openshift. Is it something we also want in devfile 2.0 ?
> > @l0rd wdyt ?
> 
> That would be useful yes.

OK, I will update it.

> 
> EDIT: Some idea about how to solve the `scheme`/`protocol` debate can be found on https://github.com/kubernetes-sigs/service-apis and in particular [the API Sketch document](https://docs.google.com/document/d/1BxYbDovMwnEqe8lj8JwHo8YxHAt3oC7ezhlFsG_tyag/preview#)

So looking into this document, and taking in account your previous remark, I assume we could come back to having a single `protocol` field, included in the Specification as well, with the following available values:
- **http**: for `http` or `https` endpoints (according to the `secure` field). Implies the `TCP` K8S service protocol
- **ws**: for `ws` or `wss` endpoints (according to the `secure` field). Implies the `TCP` K8S service protocol
- **tcp**: for tcp endpoints that are not `http` nor `ws` endoints. Implies the `TCP` K8S service protocol
- **udp**: for udp endpoints. Implies the `UDP` K8S service protocol.

@l0rd Does it make sense to you ?

l0rd:
Do we need the `secure` field? I mean if we add `https` and `wss` to the list of protocols we do not need the `secure` field at all right?

davidfestal:
> Do we need the `secure` field? I mean if we add `https` and `wss` to the list of protocols we do not need the `secure` field at all right?

I fact secure is more than simply than adding TLS on `http` and `ws` application protocols.
It is also expected to drive how endpoints are protected by some sort of authentication through workspace routing controllers.

So I assume it still makes sense to *auto-promote* `http` endpoints to `https` when `Secure` is set (as done in Devfile 1.0), since authenticating an endpoint without enabling TLS makes no sense afaict.

OTOH it seems we might also want have `https` endpoints *without* requiring authentication.

So I'd rather consider the `protocol` and the `secure` fields 2 distinct things, and add `https` and `wss` in the list of available protocols.

@l0rd do you agree ?

davidfestal:
Result is this documentation:
image

gorkem:
What does the tool (che/odo) supposed to do when it sees a `secure` field?

davidfestal:
> What does the tool (che/odo) supposed to do when it sees a `secure` field?

In Che, it triggers the setup of some authentication on secured endpoints:
- openshift console proxy based authentication for web terminal (afaict)
- JWT-proxy based authentication in Che / CRW.

Finally, when using the workspace CRD controller, the precise implementation of secured endpoint authentication is implemented by the Workspace Routing controller selected based on the routing class.

pkg/apis/workspaces/v1alpha1/containerComponent.go
```diff
@@ -8,42 +8,62 @@ type ContainerComponent struct {
...
 
+// EndpointExposure describes the way an endpoint is exposed on the network.
+// Only one of the following exposures may be specified: public, internal, none.
+// +kubebuilder:validation:Enum=public;internal;none
+type EndpointExposure string
+
+const (
+	// Endpoint will be exposed on the public network, typically through
+	// a K8S ingress or an OpenShift route 
+	PublicEndpointExposure  EndpointExposure = "public"
```
amisevsk: It looks like the codebase needs a `go fmt`

```suggestion
	PublicEndpointExposure EndpointExposure = "public"
```

pkg/apis/workspaces/v1alpha1/endpoint.go
```diff
@@ -6,12 +6,16 @@ package v1alpha1
...
 	HTTPEndpointProtocol  EndpointProtocol = "http"
-	// Endpoint will have either `ws` or `wss` traffic (according to the `Secure` field),
-	// typically on a TCP connection
+	// Endpoint will have `https` traffic, typically on a TCP connection
+	HTTPSEndpointProtocol  EndpointProtocol = "https"
+	// Endpoint will have `ws` traffic, typically on a TCP connection
+	// It will be automaticaly promoted to `wss` when the `secure` field is set to `true`
 	WSEndpointProtocol  EndpointProtocol = "ws"
+	// Endpoint will have `wss` traffic, typically on a TCP connection
+	WSSEndpointProtocol  EndpointProtocol = "wss"
```
amisevsk: According to https://github.com/devfile/kubernetes-api/issues/33#issuecomment-629345779, this field should refer to transport protocol (i.e. TCP/UDP), and the URI protocol selection should be under `.attributes.scheme`.

Also: It's not clear to me why we need separate settings for `https/wss`, given that there is also a `secure` field. What happens if you specify `https` and `secure: false`? What if `scheme: http` and `secure: true`?

It's not clear to me why we have https/wss when there is also a `secure` attribute.

davidfestal: > According to #33 (comment), this field should refer to transport protocol 

see this comment from @l0rd: https://github.com/devfile/kubernetes-api/pull/84#issuecomment-647348320

> It's not clear to me why we have https/wss when there is also a secure attribute.

see this comment https://github.com/devfile/kubernetes-api/pull/84#issuecomment-647472428 and the previous ones. Does it answer your question ?

> What if scheme: http and secure: true?

the `http` would be *auto-promoted* to `https` I assume, since TLS is a sort of logical pre-requisite for endpoint authentication. (mentioned in the schema doc)

amisevsk: Yeah sorry I missed the earlier discussion.

pkg/apis/workspaces/v1alpha1/containerComponent.go
davidfestal: Fixed

pkg/apis/workspaces/v1alpha1/endpoint.go
davidfestal: No problem :-)

l0rd:
> So I'd rather consider the `protocol` and the `secure` fields 2 distinct things, and add `https` and `wss` in the list of available protocols.
> 
> @l0rd do you agree ?

@davidfestal ok I think we can live with that right now but we may need to review that because it's mainly used for Che plugins hence not sure it should be part of the spec of the devfile. It should be a Che configuration.

davidfestal:
> > So I'd rather consider the `protocol` and the `secure` fields 2 distinct things, and add `https` and `wss` in the list of available protocols.
> > @l0rd do you agree ?
> 
> @davidfestal ok I think we can live with that right now but we may need to review that because it's mainly used for Che plugins hence not sure it should be part of the spec of the devfile. It should be a Che configuration.

So do you mean I'll let it like that for now and we'll open a distinct issue to discuss the specific case of the `secure` field ?

l0rd:
> So do you mean I'll merge the current PR as is for now, and we'll open a distinct issue to discuss the specific case of the `secure` field ?

yes

davidfestal:
@l0rd Could you please approve this PR then, it seems I cannot merge without at least one approval :-)

PR #92, opened 2020-06-30
add starterProject fields

kadel:
### What does this PR do?

adds `starterProject` field

fixes https://github.com/devfile/kubernetes-api/issues/42


kadel:
This is my first time contributing to this repo. 
@davidfestal Please, let me know if there is something more that I need to do in this PR.

kadel:
> In the future, could you please separate out the auto-generated files into a separate commit to ease review?

👍 Sure. I will make sure to do that.

davidfestal:
I'm reviewing it.

pkg/apis/workspaces/v1alpha1/projects.go
```diff
@@ -13,6 +13,16 @@ type Project struct {
 	ProjectSource `json:",inline"`
 }
 
+type StarterProject struct {
+	Name string `json:"name"`
+
```
davidfestal, line 18: @kadel Is there a reason when you omit the `clonePath` which is a field common to all types of project sources, and instructs about where to clone the project ?

Why not simply have `starterProjects` be a list of `Project` structures ? This would avoid duplicating code, no ?

schemas/devfile.json
```diff
@@ -3131,6 +3131,105 @@
         "registryUrl": {
           "type": "string"
         },
+        "starterProjects": {
+          "description": "StarterProjects is a project that can be used as a starting point when bootstraping new projects",
```
scottkurz, line 3135: bootstraping  => bootstrapping (throughout)

pkg/apis/workspaces/v1alpha1/projects.go
kadel: forgot about `clonePath` :-( will fix that

> Why not simply have starterProjects be a list of Project structures ? This would avoid duplicating code, no ?

I can't do that as there is no Description field in `Project` but I can embed whole `Project` in `StarterProject`

kadel: @davidfestal I've updated this to reuse existing `Project` struct

davidfestal: Thanks !

PR #96, opened 2020-07-14
add `hotReloadCapable` to exec commands

kadel:
### What does this PR do?
Add ~~two new options~~ a new option (~~`background`~~ and `hotReloadCapable`) to control how commands are executed.
More detailed description is in https://github.com/devfile/kubernetes-api/issues/64#issuecomment-646115779

fix https://github.com/devfile/kubernetes-api/issues/64#issuecomment-646115779

### DIFF
--- a/pkg/apis/workspaces/v1alpha1/commands.go
+++ b/pkg/apis/workspaces/v1alpha1/commands.go
@@ -117,12 +117,13 @@ type ExecCommand struct {
 	// Working directory where the command should be executed
 	WorkingDir string `json:"workingDir,omitempty"`
 
+	// +optional
 	// Optional list of environment variables that have to be set
 	// before running the command
 	Env []EnvVar `json:"env,omitempty"`
 
 	// +optional
-	// Whatever the command it capable to reload itself when source code changes.
+	// Whether the command it capable to reload itself when source code changes.
 	// If set to `true` the command won't be restarted and it is expected to handle file changes on its own.
 	//
 	// Default value is `false`
--- a/pkg/apis/workspaces/v1alpha1/commands.go
+++ b/pkg/apis/workspaces/v1alpha1/commands.go
@@ -123,16 +123,7 @@ type ExecCommand struct {
 	Env []EnvVar `json:"env,omitempty"`
 
 	// +optional
-	// Whatever the command needs should be executed on background.
-	// If "false" the controller should wait for command to finish and report its status.
-	// If "true" the controller won't wait for it to finish (command can run indefinitely or until it is manually stopped).
-	// Default value is "true" for group.kind: run, debug.
-	// Default value is "false" for group.kind: build, test.
-	// Default value is "false" for all commands without group.kind
-	Background bool `json:"background,omitempty"`
-
-	// +optional
-	// Whatever the command needs should be restarted when files changed.
+	// Whatever the command needs to be restarted when files changed.
 	// It should be set to "false" if command is capable to do automatic hotreload.
 	// Default value is "true" for group.kind: build, test.
 	// Default value is "false" for group.kind: run, debug.
--- a/pkg/apis/workspaces/v1alpha1/commands.go
+++ b/pkg/apis/workspaces/v1alpha1/commands.go
@@ -117,7 +117,6 @@ type ExecCommand struct {
 	// Working directory where the command should be executed
 	WorkingDir string `json:"workingDir,omitempty"`
 
-	// +optional
 	// Optional list of environment variables that have to be set
 	// before running the command
 	Env []EnvVar `json:"env,omitempty"`
@@ -125,9 +124,14 @@ type ExecCommand struct {
 	// +optional
 	// Whatever the command needs to be restarted when files changed.
 	// It should be set to "false" if command is capable to do automatic hotreload.
-	// Default value is "true" for group.kind: build, test.
-	// Default value is "false" for group.kind: run, debug.
-	// Default value is "false" for all commands without group.kind
+	//
+	// Default value:
+	//
+	//   - "true" for group.kind: build, test.
+	//
+	//   - "false" for group.kind: run, debug.
+	//
+	//   - "false" for all commands without group.kind
 	RestartOnChange bool `json:"restartOnChange,omitempty"`
 }
 
--- a/pkg/apis/workspaces/v1alpha1/commands.go
+++ b/pkg/apis/workspaces/v1alpha1/commands.go
@@ -121,6 +121,23 @@ type ExecCommand struct {
 	// Optional list of environment variables that have to be set
 	// before running the command
 	Env []EnvVar `json:"env,omitempty"`
+
+	// +optional
+	// Whatever the command needs should be executed on background.
+	// If "false" the controller should wait for command to finish and report its status.
+	// If "false" the controller won't wait for it to finish (command can run indefinitely or until it is manually stopped).
+	// Default value is "true" for group.kind: run, debug.
+	// Default value is "false" for group.kind: build, test.
+	// Default value is "false" for all commands without group.kind
+	Background bool `json:"background,omitempty"`
+
+	// +optional
+	// Whatever the command needs be restarted when files changed.
+	// It should be set to "false" if command is capable to do automatic hotreload.
+	// Default value is "true" for group.kind: build, test.
+	// Default value is "false" for group.kind: run, debug.
+	// Default value is "false" for all commands without group.kind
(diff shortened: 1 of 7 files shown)


pkg/apis/workspaces/v1alpha1/commands.go
```diff
@@ -121,6 +121,23 @@ type ExecCommand struct {
 	// Optional list of environment variables that have to be set
 	// before running the command
 	Env []EnvVar `json:"env,omitempty"`
+
+	// +optional
+	// Whatever the command needs should be executed on background.
+	// If "false" the controller should wait for command to finish and report its status.
+	// If "false" the controller won't wait for it to finish (command can run indefinitely or until it is manually stopped).
```
amisevsk, line 128:
```suggestion
	// Whatever the command needs should be executed on background.
	// If "false" the controller should wait for command to finish and report its status.
	// If "true" the controller won't wait for it to finish (command can run indefinitely or until it is manually stopped).
```

pkg/apis/workspaces/v1alpha1/commands.go
```diff
@@ -121,6 +121,23 @@ type ExecCommand struct {
...
+	// Default value is "true" for group.kind: run, debug.
+	// Default value is "false" for group.kind: build, test.
```
amisevsk, line 130: I worry about setting different defaults depending on `group.kind`. Some builds, for example, are designed to pick up changes and rebuild automatically (e.g. with node). Having an optional value that defaults differently depending on `group.kind` means that the writer needs to cross-reference more fields in the spec to determine behaviour.

pkg/apis/workspaces/v1alpha1/commands.go
```diff
@@ -121,6 +121,23 @@ type ExecCommand struct {
...
+	// Default value is "false" for all commands without group.kind
+	Background bool `json:"background,omitempty"`
+
+	// +optional
+	// Whatever the command needs be restarted when files changed.
```
amisevsk, line 135:
```suggestion
	// Whatever the command needs should be restarted when files changed.
```
To match the phrasing above, though I feel like we could word this more clearly.

pkg/apis/workspaces/v1alpha1/commands.go
kadel: for `restartOnChange` we could define one default value, but it will have to be `true`.

But I'm afraid that for `background` field we can't pick just one default value for all group kind. Whatever value we pick, it won't make sense for part of the commands :-( 
If we pick `false` it will make it  weird for `run` commands (especially in odo)
If we pick `true` it will make it really weird for most `build` commands :-(

amisevsk: Yeah I don't see a good way around this issue. I'm fine with it as-is.

kadel:
I've removed `background` field from this PR.  We should finish discussion in https://github.com/devfile/kubernetes-api/issues/64#issuecomment-658613030 to make sure that we really need this field. I will open another PR once we have resolution on `background`

`restartOnChange` is straight forward I think that this one can go in.

pkg/apis/workspaces/v1alpha1/commands.go
```diff
@@ -121,6 +121,14 @@ type ExecCommand struct {
...
 	// before running the command
 	Env []EnvVar `json:"env,omitempty"`
+
+	// +optional
+	// Whatever the command needs to be restarted when files changed.
+	// It should be set to "false" if command is capable to do automatic hotreload.
+	// Default value is "true" for group.kind: build, test.
+	// Default value is "false" for group.kind: run, debug.
+	// Default value is "false" for all commands without group.kind
+	RestartOnChange bool `json:"restartOnChange,omitempty"`
```
davidfestal, line 131: If the default value is not fixed to `false` , then the Go definition should probably be a pointer (`*bool`) type:

```go
	RestartOnChange *bool `json:"restartOnChange,omitempty"`
```

This is the only way, in GO, to distinguish, for booleans, between:
- no value was provided (=> build a default value according to some logic)
- a `false` value.

(see [this comment](https://github.com/golang/go/issues/13284#issuecomment-158153815))

pkg/apis/workspaces/v1alpha1/commands.go
```diff
@@ -121,6 +121,14 @@ type ExecCommand struct {
 	// Optional list of environment variables that have to be set
 	// before running the command
 	Env []EnvVar `json:"env,omitempty"`
+
+	// +optional
+	// Whatever the command needs to be restarted when files changed.
```
davidfestal, line 126: If you want to have your comment nicely rendered as workdown (with new lines and a markdown list) in the generated documentation + Documentation hovers, you could build your comment this way:

```
	// Whatever the command needs to be restarted when files changed.
	// It should be set to "false" if command is capable to do automatic hotreload.
	//
	// Default value:
	//
	// - `true` for group.kind: build, test.
	//
	// - `false` for group.kind: run, debug.
	//
	// - `false` for all commands without group.kind
```

pkg/apis/workspaces/v1alpha1/commands.go
amisevsk: This is a good point, but will also require nil checks wherever it's used.

kadel: @davidfestal Good point.

I'm still considering if different default values based on command kind is a good idea.
Would it make sense for Che if we always default this to `true`?

davidfestal: @kadel `true` means "no hot reload" => the user (or Che) will have to restart the command on each file change, right ?

kadel: > @kadel `true` means "no hot reload" => the user (or Che) will have to restart the command on each file change, right ?

yes

pkg/apis/workspaces/v1alpha1/commands.go
kadel: This is nice, did not know about this. I've updated the description.

pkg/apis/workspaces/v1alpha1/commands.go
kadel: updated description with information that default value for `restartOnChange` is `true` (for all group.kind)

kadel:
I have renamed  `restartOnChange` to `hotReloadCapable` so we can have a default value of this flag `false`.

Boolean fields in Json files with default value `true` are complicated to handle in Go.

pkg/apis/workspaces/v1alpha1/commands.go
```diff
@@ -117,10 +117,16 @@ type ExecCommand struct {
 	// Working directory where the command should be executed
 	WorkingDir string `json:"workingDir,omitempty"`
 
-	// +optional
```
davidfestal, line 120: Why did you remove `+optional` here ?

pkg/apis/workspaces/v1alpha1/commands.go
```diff
@@ -117,10 +117,16 @@ type ExecCommand struct {
...
 	// Optional list of environment variables that have to be set
 	// before running the command
 	Env []EnvVar `json:"env,omitempty"`
+
+	// +optional
+	// Whatever the command it capable to reload itself when source code changes.
+	// If set to "true" the command won't be restarted and it is expected to handle file changes on its own.
```
davidfestal, line 126:
```suggestion
	// If set to `true` the command won't be restarted and it is expected to handle file changes on its own.
```

pkg/apis/workspaces/v1alpha1/commands.go
```diff
@@ -117,10 +117,16 @@ type ExecCommand struct {
 	// Working directory where the command should be executed
 	WorkingDir string `json:"workingDir,omitempty"`
 
-	// +optional
 	// Optional list of environment variables that have to be set
 	// before running the command
 	Env []EnvVar `json:"env,omitempty"`
+
+	// +optional
+	// Whatever the command it capable to reload itself when source code changes.
```
davidfestal, line 125: > Whatever the command it capable

Did you mean: **_whether the command is capable_** ?

pkg/apis/workspaces/v1alpha1/commands.go
kadel: that is a mistake :-/

pkg/apis/workspaces/v1alpha1/commands.go
kadel: yes 😇 , fixed. thx

Issue #25, opened 2020-02-06
Parent Devfiles


l0rd:
We want to experiment how the `Devfile` format could be used to define a stack.

### Requirements

**1. Parent devfile**

A parent devfile is a devfile published somewhere where `odo` or Che can read 
it. Like a gist, a registry or a kubernetes resource (see `referencing parents` below). 

A parent devfile is a regular devfile and includes components (for build and runtime), one or more code sample projects, events and commands to build and run the sample applications. 

Any devfile that references a parent devfile will inherit its components, events and commands but won't inherit the code sample.

**2. Referencing parents: uri, id or kubernetes**

A parent devfile can be referenced in 3 different ways. Using its **`id`** if it has been published in a registry:

```yaml
parent:
    id: redhat/nodejs/11.6  # <-- the id format is <publisher>/<stack-name>/<version>
    registry: https://devfile-registry.io/
```

Using the **URI** if it has been published on a static http server (like gist or pastebin):

```yaml
parent:
    uri: https://raw.githubusercontent.com/eclipse/che-devfile-
```

Using a **Kubernetes** resource name, namespace if it has been deployed on a Kubernete cluster as a `DevWorkspaceTemplate`:

```yaml
parent:
    kubernetes:
        name: mydevworkspacetemplate
        namespace: mynamespace
```

**3. Customizing parent configuration**

When referencing a parent, a devfile should be able to customize its configuration. The syntax to customize a parent configuration is similar to [kustomize](https://kustomize.io/):  

```yaml
parent:
  id: redhat/theia-vsx-stack/latest     # <--- Parent referenced by id 
  components:                                  # <--- Parent configuration can be customized
    - name: vsx-installer                    # <--- Should match the name of an existing parent component
      container:
         env:
            - name: VSX_LIST
               value: java-dbg.vsix,java.vsix
  commands:
    (...)
components:                          # The main Devfile body provides new elements, not inherited from the parent
  - name: tooling                     # <--- Should *not* match the name of a parent component 
    container:                           # <--- Components are added to parent's components
      image: busybox
events:                                   # <--- Events are only allowed in the main Devfile body,
  (...)                                       #        and the resulting events is the merged of the these events with the events of the parent.  
```
The place where an customizable element (component, command or project) should be defined is clearly driven by the following rule. It is either:
- inside the scope of the `parent` element itself if it is an **override** of an **existing** parent element,
- at the toplevel of the current devfile if it is a **new** element not present in the parent.

As a consequence;
[...]

PR #98, opened 2020-07-17
Implementation of parent and plugin overriding

davidfestal:
### What does this PR do?

This PR implements the parent or plugin overriding.

The overriding of parent devfile or devfile plugins is implemented using the Strategic merge patch K8S library.

#### Implementation entry points;

The implementation is available for import in the [`pkg/utils/overriding`](https://github.com/devfile/kubernetes-api/blob/overriding-implementation/pkg/utils/overriding) package.

Available methods are:
- [`OverrideDevWorkspaceTemplateSpec()`](https://github.com/devfile/kubernetes-api/blob/overriding-implementation/pkg/utils/overriding/overriding.go#L200) to work with GO structs directly
- [`OverrideDevWorkspaceTemplateSpecBytes`](https://github.com/devfile/kubernetes-api/blob/overriding-implementation/pkg/utils/overriding/overriding.go#L106) to work with `json` or `yaml` content.

In both case the result is a `DevWorkspaceTemplateSpecContent` Go struct that corresponds to the code content of a devfile 2.0 (without the top-level `apiVersion` and `metadata` fields).

Note that one can only override an **existing** element of a parent or a plugin.

If a new element is added, it should be added in the main (child) devfile body. And then simple merging of top-level lists is done between the main devfile body, the flattened parent and the flattened plugins. If the key (name of ID) of a element is found several times in the merge, an error is thrown.

The corresponding methods are:
- [`MergeDevWorkspaceTemplateSpec()`](https://github.com/devfile/api/blob/overriding-implementation/pkg/utils/overriding/merging.go#L208) to work with GO structs directly
- [`MergeDevWorkspaceTemplateSpecBytes`](https://github.com/devfile/api/blob/overriding-implementation/pkg/utils/overriding/merging.go#L154) to work with `json` or `yaml` content.

In both case the result is a `DevWorkspaceTemplateSpecContent` Go struct that corresponds to the code content of a devfile 2.0 (without the top-level `apiVersion` and `metadata` fields).

#### Examples of use:

- An example of use with GO structs can be found in the following test: https://github.com/devfile/kubernetes-api/blob/overriding-implementation/pkg/utils/overriding/overriding_test.go#L160
- An example of use with YAML files structs can be found in the following test: https://github.com/devfile/kubernetes-api/blob/overriding-implementation/pkg/utils/overriding/overriding_test.go#L168
- A similar example exists for the merging part.

#### YAML examples:

YAML example files are provided and used in GO tests.

##### Overriding

Overriding examples can be found in the following [`test-fixtures/patches`](https://github.com/devfile/kubernetes-api/tree/overriding-implementation/pkg/utils/overriding/test-fixtures/patches) folder: each sub-folder contains 3 Yaml files:
- `original.yaml`: the original Yaml that should be overridden  
- `patch.yaml`: the Yaml patch that should be used to override the original file
- `result.yaml`: the expected result of the overriding, which is the result of applying strategic merge patch on the original file.  

##### Merging

Merging examples can be found in the following [`test-fixtures/merges`](https://github.com/devfile/kubernetes-api/tree/overriding-implementation/pkg/utils/overriding/test-fixtures/merges) folder: each sub-folder contains up to 4 Yaml files:
- `main.yaml`: the Yaml of the main (child) devfile body content     
- `parent.yaml`: the Yaml of a flattened parent referenced by the main devfile 
- `plugin.yaml`: the Yaml a flattened plugin referenced by the main devfile
- `result.yaml`: the expected result of the merging of all those devfile contents.  

When opened in VSCode or [`Hosted Che`](https://che.openshift.io/f/?url=https://github.com/devfile/kubernetes-api/tree/overriding-implementation), all these example files are validated with appropriate and up-to-date schemas. So they benefit from YAML language services such as completion and doc hovers. 

Contributions are welcomed in the future to further complete the tests (and at the same time documentation) of overriding rules.

#### Noteworthy (and possibly breaking) changes

- [`Events`](https://github.com/devfile/kubernetes-api/pull/98/files?file-filters%5B%5D=.go#diff-80b8e2cc67c3cc3fc4cd18ef906b1316R22):
  - Detailed change: Instead of directly referencing the `Events` structure, the `DevWorkspaceTemplateSpecContent` struct contains a pointer to an `Events` struct.
  - Reason: This make testing easier, since it avoids a number of corner cases around serialization, when an empty `Events` object is created although no `events` field is present in the parsed `json` or `yaml`.
  - Impact: it will require a change in depending code, and a nil-check after json or yaml parsing.

##### Keys <span id="keys"></span>

Currently in the existing model, for some top-level lists (`components`, `commands`), the keys of elements are not directly in the objects stored in the list, but one level lower:

```yaml
commands:
  - exec:
      id: "commandId"  <------ the key of the given command in the list of commands 
```
instead of 
```yaml
commands:
  - id: "commandId"
    exec:
      ....
```

Furthermore for some components as `plugin`s, the key of a plugin in the list of components may be calculated, and may depend on a number of fields (`id` and `registryUrl` for example).
Thus a [`Keyed`](https://github.com/devfile/kubernetes-api/blob/overriding-implementation/pkg/apis/workspaces/v1alpha1/keyed.go#L3) interface is defined and implemented for commands, components and component overrides, and it is then used to setup fixed keys in the `commands` and `components` lists of both the original object and the patch object during overriding.

##### Union normalizing

In the GO API, some model types are defined as [Kubernetes unions](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190325-unions.md) with discriminators.

The API [`Union`](https://github.com/devfile/kubernetes-api/blob/overriding-implementation/pkg/apis/workspaces/v1alpha1/union.go#L6) interface defines helpers to manage correctly manage those unions, especially maintaining the fact that values in a union are mutually-exclusive.

This union management logic is then [used in the overriding implementation](https://github.com/devfile/kubernetes-api/blob/overriding-implementation/pkg/utils/overriding/overriding.go#L147), but is also available in a dedicated [`pkg/utils/unions`](https://github.com/devfile/kubernetes-api/tree/overriding-implementation/pkg/utils/unions) package that allows managing all the unions embedded in a wider object.

This PR references issue https://github.com/devfile/kubernetes-api/issues/91

The overriding logic implemented in this PR has associated tests


kadel:
Going over this PR I noticed something that for some reason I didn't notice before.

The current design looks quite weird, and it is not how Strategic Merge Patch works in Kubernetes.

https://github.com/devfile/kubernetes-api/issues/25#issuecomment-661783929

davidfestal:
> Going over this PR I noticed something that for some reason I didn't notice before.
> 
> The current design looks quite weird, and it is not how Strategic Merge Patch works in Kubernetes.
> 
> More in my comment at [#25 (comment)](https://github.com/devfile/api/issues/25#issuecomment-661783929)

@kadel I implemented the agreement we found (summarized [here](https://github.com/devfile/api/issues/25#issuecomment-668672087) ) in the PR, and updated the PR description to add details about the *merging* side of things (addition of elements that are not present in parents or plugins).

davidfestal:
cc @ranakan19

pkg/utils/overriding/merging.go
```diff
@@ -0,0 +1,229 @@
...
+	"k8s.io/apimachinery/pkg/util/json"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/yaml"
+)
+
+func ensureNoConflicWithParent(mainContent *keyedDevWorkspaceTemplateSpecContent, parentflattenedContent *keyedDevWorkspaceTemplateSpecContent) error {
+	return checkKeys(func(elementType string, keysSets []sets.String) []error {
+		mainKeys := keysSets[0]
+		parentOrPluginKeys := keysSets[1]
+		overridenElementsInMainContent := mainKeys.Intersection(parentOrPluginKeys)
```
amisevsk, line 18:
```suggestion
		overriddenElementsInMainContent := mainKeys.Intersection(parentOrPluginKeys)
```

pkg/utils/overriding/merging.go
```diff
@@ -0,0 +1,229 @@
...
+	},
+		allSpecs...)
+}
+
+func mergeKeyedDevWorkspaceTemplateSpec(mainContent *keyedDevWorkspaceTemplateSpecContent, parentFlattenedContent *keyedDevWorkspaceTemplateSpecContent, pluginFlattenedContents ...*keyedDevWorkspaceTemplateSpecContent) (*keyedDevWorkspaceTemplateSpecContent, error) {
+	allContents := []*keyedDevWorkspaceTemplateSpecContent { parentFlattenedContent }
+	allContents = append(allContents, pluginFlattenedContents...)
+	allContents = append(allContents, mainContent)
+	for _, content := range allContents {
+		addKeys(content)
```
amisevsk, line 73: Ignored error

pkg/utils/overriding/overriding.go
```diff
@@ -0,0 +1,157 @@
+package overriding
+
+import (
+	//	"errors"
```
amisevsk, line 4:
```suggestion
```

pkg/utils/overriding/overriding.go
```diff
@@ -0,0 +1,157 @@
...
+
+	workspaces "github.com/devfile/api/pkg/apis/workspaces/v1alpha1"
+	unions "github.com/devfile/api/pkg/utils/unions"
+	"k8s.io/apimachinery/pkg/util/json"
+	"k8s.io/apimachinery/pkg/util/sets"
+	strategicpatch "k8s.io/apimachinery/pkg/util/strategicpatch"
+	"k8s.io/apimachinery/pkg/util/yaml"
+)
+
+func ensureOnlyExistingElementsAreOverriden(keyedSpec *keyedDevWorkspaceTemplateSpecContent, keyedOverlay *keyedDevWorkspaceTemplateSpecContent) error {
```
amisevsk, line 16:
```suggestion
func ensureOnlyExistingElementsAreOverridden(keyedSpec *keyedDevWorkspaceTemplateSpecContent, keyedOverlay *keyedDevWorkspaceTemplateSpecContent) error {
```

amisevsk:
I had to rebase on master as there was a conflict. Commit https://github.com/devfile/api/pull/98/commits/13fbbc9d6b9f9e5c19feee1678885e83492b700a resolves the minor comments from my review.

davidfestal:
Thanks @amisevsk for this deep review and very-valuable fixes.

PR #99, opened 2020-07-18
Add devworkspace condition for failed startup

amisevsk:
### What does this PR do?
Add a condition that can be used to communicate to the user reasons why a devworkspace cannot start (if e.g. a plugin cannot be found, the controller is misconfigured, etc.)

Required for https://github.com/eclipse/che/issues/17326

With changes, a failed workspace can be investigated, e.g.:
```
❯ kc get devworkspace cloud-shell -o yaml | yq -Y '.status'
conditions:
  - lastTransitionTime: "2020-07-18T01:02:10Z"
    message: Devworkspace was created without creator ID label. It must be recreated
      to resolve the issue
    status: "True"
    type: FailedStart
phase: Failed
workspaceId: workspace818ff97fbf7741fb
```

#### Docs PR
Unsure if docs are required for this, will provide if needed.

### DIFF
--- a/pkg/apis/workspaces/v1alpha1/devworkspace_types.go
+++ b/pkg/apis/workspaces/v1alpha1/devworkspace_types.go
@@ -59,6 +59,7 @@ const (
 	WorkspaceRoutingReady        WorkspaceConditionType = "RoutingReady"
 	WorkspaceServiceAccountReady WorkspaceConditionType = "ServiceAccountReady"
 	WorkspaceReady               WorkspaceConditionType = "Ready"
+	WorkspaceFailedStart         WorkspaceConditionType = "FailedStart"
 )
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object


amisevsk:
I'm also wondering if specific conditionType constants should be in this repo at all -- there's no hard-and-fast rule about where they're defined (you can basically supply any string) and so implementations of controllers may want to use their own instead of the ones included here. 

cc: @davidfestal WDYT?

davidfestal:
> I'm also wondering if specific conditionType constants should be in this repo at all -- there's no hard-and-fast rule about where they're defined (you can basically supply any string) and so implementations of controllers may want to use their own instead of the ones included here.
> 
> cc: @davidfestal WDYT?

Conditions are especially useful when using `kubectl wait`, for example in:

```bash
kubectl wait workspace java-mysql --for=condition=Ready
```

So I'm not against providing some flexibility for implementors to add condition types, but I'd probably be against any implementation change that would change typical condition semantics and expectations, and thus might break the way their are useful in either `kubectl wait` or the Openshift console for example.

OTOH condition types somehow define lifecycles step of the workspace, so I still think it might be useful to have them be part of the specification, at least the basic conditions we think will always make sense.

PR #101, opened 2020-07-20
VolumeMount should require Volume component definition

kadel:
### What does this PR do?
`VolumeMount` should require `Volume` component definition.
Otherwise, it doesn't make much sense to add it implicitly. 
One of the reasons is that could lead to errors like this:

```yaml
components:
  - container:
      name: cont1
      volumeMounts:
        - name: myvolume
  - container:
      name: cont2
      volumeMounts:
        # I want to use myvolume, but made a typo
        - name: myvoluem

volumes:
   - name: myvolume
```


maysunfaisal:
@l0rd when and how is the schema here updated? https://devfile.github.io/devfile/_attachments/api-reference.html

PR #106, opened 2020-08-04
Adding dockerfile component

ranakan19:
### What does this PR do?
Adds dockerfile component for devfile 2.1.0

In reference to https://github.com/devfile/parser/issues/26

Not applicable.

### DIFF
--- a/pkg/apis/workspaces/v1alpha1/dockerfileComponent.go
+++ b/pkg/apis/workspaces/v1alpha1/dockerfileComponent.go
@@ -1,0 +1,26 @@
+package v1alpha1
+
+//Dockerfile component of devfile
+type Dockerfile struct {
+	// Mandatory name that allows referencing the Volume component in Container volume mounts or inside a parent
+	Name string `json:"name"`
+
+	// Mandatory path to source code
+	Source *Source `json:"source"`
+
+	// Mandatory path to dockerfile
+	DockerfileLocation string `json:"dockerfileLocation"`
+
+	// Mandatory destination to registry to push built image
+	Destination string `json:"destination,omitempty"`
+}
+
+
+//Source within dockerfile component
+type Source struct {
+	// Mandatory path to local source directory folder
+	SourceDir string `json:"sourceDir"`
+
+	// Mandatory path to source repository hosted locally or on cloud
+	Location string `json:"location"`
+}
--- a/pkg/apis/workspaces/v1alpha1/dockerfileComponent.go
+++ b/pkg/apis/workspaces/v1alpha1/dockerfileComponent.go
@@ -13,6 +13,9 @@ type Dockerfile struct {
 
 	// Mandatory destination to registry to push built image
 	Destination string `json:"destination,omitempty"`
+	
+	// Mandatory whether rootless/unprivileged builder container is required
+	Rootless bool `json:"rootless,omitempty"`
 }
 
 
--- a/pkg/apis/workspaces/v1alpha1/dockerfileComponent.go
+++ b/pkg/apis/workspaces/v1alpha1/dockerfileComponent.go
@@ -5,25 +5,24 @@ type Dockerfile struct {
 	// Mandatory name that allows referencing the Volume component in Container volume mounts or inside a parent
 	Name string `json:"name"`
 
-	// Mandatory path to source code
-	Source *Source `json:"source"`
+	// path to source code, if empty - source is assumed as the directory having devfile
+	Source *Source `json:"source,omitempty"`
 
 	// Mandatory path to dockerfile
 	DockerfileLocation string `json:"dockerfileLocation"`
 
-	// Mandatory destination to registry to push built image
+	// destination to registry to push built image
 	Destination string `json:"destination,omitempty"`
-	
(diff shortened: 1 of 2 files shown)


pkg/apis/workspaces/v1alpha1/dockerfileComponent.go
```diff
@@ -0,0 +1,26 @@
...
+
+	// Mandatory path to source code
+	Source *Source `json:"source"`
+
+	// Mandatory path to dockerfile
+	DockerfileLocation string `json:"dockerfileLocation"`
+
+	// Mandatory destination to registry to push built image
+	Destination string `json:"destination,omitempty"`
+}
```
wtam2018, line 16:
```suggestion
	
	// Mandatory whether rootless/unprivileged builder container is required
	Rootless bool `json:"rootless,omitempty"`
}
```

kadel:
This is targeted to 2.1.0 but we have only a single version of types. How is this versioning going to work?

ranakan19:
> This is targeted to 2.1.0 but we have only a single version of types. How is this versioning going to work?

I'm not sure i understand your question, do you mean the schemaVersion? In the devfile library, devfile 2.0.0 and 2.1.0 are separate schema, but i figured that the kubernetes-api can have all the common structs to be used.

kadel:
> > This is targeted to 2.1.0 but we have only a single version of types. How is this versioning going to work?
> 
> @kadel I'm not sure i understand your question, do you mean the schemaVersion? In the devfile library, devfile 2.0.0 and 2.1.0 are separate schema, but i figured that the kubernetes-api can have all the common structs to be used.

Yes, I'm wondering how is this going to work with `schemaVersion`.

For example, If I have a devfile that has `schemaVersion: 2.0.0` with Dockerfile component.
Would that get parsed or not?

kadel:
btw @ranakan19, you need to run `./docker-run.sh ./build.sh` to generate updated Json Schema and other autogenerated files. Autogenerated files should be included in your PR as a separate commit.

davidfestal:
> > > This is targeted to 2.1.0 but we have only a single version of types. How is this versioning going to work?
> > 
> > 
> > @kadel I'm not sure i understand your question, do you mean the schemaVersion? In the devfile library, devfile 2.0.0 and 2.1.0 are separate schema, but i figured that the kubernetes-api can have all the common structs to be used.
> 
> Yes, I'm wondering how is this going to work with `schemaVersion`.
> 
> For example, If I have a devfile that has `schemaVersion: 2.0.0` with Dockerfile component.
> Would that get parsed or not?

I assume that at some point we would have to keep several versions of the K8S apis (like the v1, v2alpha1, etc, ...), and possibly maintain a mapping between a devfile schema version (semantic-versioning compatible) and the corresponding K8S API version.

That's something that is done for K8S itself for example. But we should obviously discuss how this could be achieved as simply as possible in our case, knowing that we want to avoid in upcoming devfile schema versions, changes that would introduce incompatible changes.

davidfestal:
I created PR https://github.com/ranakan19/kubernetes-api/pull/2 to fix some small errors, and add required code for correct union management.

davidfestal:
I have 2 other questions:

1. In this code block:
```go
//Source within dockerfile component
type Source struct {
	// path to local source directory folder
	SourceDir string `json:"sourceDir,omitempty"`

	// path to source repository hosted locally or on cloud
	Location string `json:"location,omitempty"`
}
```
Are the 2 fields mutually exclusive and can they *both* have a value defined.
If they are mutually exclusive, it could be defined as a union, so that the schema itself would ensure this union-member oneness.

2. Does overriding the dockerimage component of a parent makes sense ?
For now the dockerimage is not added into the overrides. And if yes, does it also make sense to override a dockerfile component that would be encapsulated in a plugin definition ?

wtam2018:
> Are the 2 fields mutually exclusive and can they _both_ have a value defined.
They are mutually exclusive but `SourceDir` does not seem to make sense to be in there.   We can have `SourceLocation` string instead of `Source` struct.

ranakan19:
> I have 2 other questions:
> 
> 1. In this code block:
> 
> ```go
> //Source within dockerfile component
> type Source struct {
> 	// path to local source directory folder
> 	SourceDir string `json:"sourceDir,omitempty"`
> 
> 	// path to source repository hosted locally or on cloud
> 	Location string `json:"location,omitempty"`
> }
> ```
> 
> Are the 2 fields mutually exclusive and can they _both_ have a value defined.
> If they are mutually exclusive, it could be defined as a union, so that the schema itself would ensure this union-member oneness.
> 
> 1. Does overriding the dockerimage component of a parent makes sense ?
>    For now the dockerimage is not added into the overrides. And if yes, does it also make sense to override a dockerfile component that would be encapsulated in a plugin definition ?

for 1, Yes the fields are mutually exclusive. I'm not sure about 2, we should discuss it in the meeting.

wtam2018 (commented):
This PR is superseded by https://github.com/devfile/api/pull/127.   Please close this one.

sleshchenko:
Closing in favor of https://github.com/devfile/api/pull/127

PR #112, opened 2020-08-13
Add minimal CI Github action 

davidfestal:
### What does this PR do?

This PR adds minimal CI Github action to run GO tests.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -22,9 +22,6 @@ jobs:
     # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
     - uses: actions/checkout@v2
 
-    - name: setup-docker
-      uses: docker-practice/actions-setup-docker@0.0.1
-
     - name: Setup Go environment
       uses: actions/setup-go@v2.1.2
       with:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,33 @@
+# This is a basic workflow to help you get started with Actions
+
+name: CI
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -22,12 +22,18 @@ jobs:
     # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
     - uses: actions/checkout@v2
 
+    - name: setup-docker
```
benoitf, line 25: AFAIK docker is already setup on runners, so why do we need it ?

davidfestal: I didn't know that it could be used as a command line. I'll try without it.

benoitf: https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md

benoitf: here are all packages installed. There is also `oc` there :-)

davidfestal: removed. Thanks for the information !

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,41 @@
...
+        # The Go version to download (if necessary) and use. Supports semver spec and ranges.
+        go-version: 1.13
+    
+    - name: Generate CRD and schema
+      run: ./docker-run.sh ./build.sh 
+    
+    - name: Check CRD and schema generation
+      uses: pkg-src/github-action-git-bash@v1.1
+      with:
+        args: bash -c "git diff --exit-code || { echo './docker-run.sh ./build.sh did introduce changes, which should not be  the case.\nPlease run it locally and checkout the results.'; exit 1; }"
```
amisevsk, line 37: Question about GitHub actions more than anything else: are we guaranteed to have these steps run in this sequence? (I believe we are but want to double check)

davidfestal: @benoitf do you have the answer to this question ? It seems I took it for granted ;-)

benoitf: yes, steps are in sequence (if step 1 failed, step2 is ignored and job failed)

it's jobs that are in parallel

benoitf: BTW I don't know why we use a custom git action as git is already installed

davidfestal: because it is not available in a bash command by default.

benoitf: Screenshot 2020-08-13 at 16 37 31
Screenshot 2020-08-13 at 16 36 56

PR #114, opened 2020-08-13
Small wording fix in `ci.yaml`

davidfestal:
### What does this PR do?

Small wording fix in `ci.yaml`

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -34,7 +34,7 @@ jobs:
     - name: Check CRD and schema generation
       uses: pkg-src/github-action-git-bash@v1.1
       with:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -34,7 +34,7 @@ jobs:
     - name: Check CRD and schema generation
       uses: pkg-src/github-action-git-bash@v1.1
       with:
-        args: bash -c "git diff --exit-code || { echo './docker-run.sh ./build.sh did introduce changes, which should not be  the case.\nPlease run it locally and checkout the results.'; exit 1; }"
+        args: bash -c "git diff --exit-code || { echo 'Command `./docker-run.sh ./build.sh` did introduce changes, which should not be the case if it had been run as part of the PR. Please run it locally and checkin the results as part of your PR.'; exit 1; }"
```
amisevsk, line 37:
```suggestion
        args: bash -c "git diff --exit-code || { echo 'Command `./docker-run.sh ./build.sh` did introduce changes, which should not be the case if it had been run as part of the PR. Please run it locally and check in the results as part of your PR.'; exit 1; }"
```

Issue #109, opened 2020-08-10
Parent Devfiles and workingDir vs clonePath issue


kadel:
Currently `clonePath` defaults to the project name. This means that by default source code is cloned into `/project/<project-name>`. This is problematic for Stacks (Parent Devfiles) as they can't know where the project source code will be.

### Example:
How should I set `workingDir` in my example devfile.yaml so It is generic enough that it can be used without everyone overriding `workingDir` for all commands?

```yaml
# parent.yaml
schemaVersion: 2.0.0
metadata:
  name: java
components:
  - container:
      name: tools
      image: quay.io/eclipse/che-java11-maven:nightly
      mountSources: true
commands:
  - exec:
      id: mvn-package
      component: tools
      commandLine: "mvn package"
      # workingDir: ?????
      group:
        kind: build
        isDefault: true
  - exec:
      id: run
      component: tools
      commandLine: "java -jar target/*.jar"
      # workingDir: ?????
      group:
        kind: run
        isDefault: true
```

Now I want to use this in my local devfile.yaml

```yaml
# devfile.yaml
schemaVersion: 2.0.0
metadata:
  name: myproject
projects:
  - name: myproject
      git:
        location: https://github.com/example/example
parent:
  uri: https://registry.example.com/java/parent.yaml
```

This won't work because commands in parent devfile don't specify `workingDir` so it will default to container working directory (in this case `/project`). But I can't set working dir in parent devfile, because I can't know what will be the project name.

johnmcollier:
A similar issue to this, is when the user brings their own project, rather than using one of the ones defined in the devfile.

If the commands in a devfile assume the code is synced to `$PROJECTS_ROOT/$CLONE_PATH`, that'll cause issues for cases like in odo, where users can use their own project/code with the devfile (and so there isn't a possible value for `clonePath`)

kadel:
After a lengthy discussion, we agreed to introduce a new variable `$PROJECT_SOURCE` that will be expanded to a directory with project source for single project Devfiles.

In other words:
`$PROJECT_SOURCE = $PROJECT_ROOT/<project_name>`

TODO @kadel: add more details and update documentation

PR #116, opened 2020-08-17
add PROJECTS_ROOT and PROJECT_SOURCE description to CommandLine and WorkingDir

kadel:
…

### What does this PR do?
Add information about  environment  variables that are available in `commandLine` and `WorkingDir` fields
`$PROJECTS_ROOT`: A path where projects sources are mounted
`$PROJECT_SOURCE`: A path to a project source (`$PROJECTS_ROOT/<project-name>`). 

fixes https://github.com/devfile/api/issues/109

### DIFF
--- a/pkg/apis/workspaces/v1alpha1/commands.go
+++ b/pkg/apis/workspaces/v1alpha1/commands.go
@@ -124,9 +124,9 @@ type ExecCommand struct {
 	//
 	// Special variables that can be used:
 	//
-	//  - `$PROJECTS_ROOT`: A path where projects sources are mounted
+	//  - `${PROJECTS_ROOT}`: A path where projects sources are mounted
 	//
-	//  - `$PROJECT_SOURCE`: A path to a project source ($PROJECTS_ROOT/<project-name>). If there are multiple projects, this will point to the directory of the first one.
+	//  - `${PROJECT_SOURCE}`: A path to a project source (${PROJECTS_ROOT}/<project-name>). If there are multiple projects, this will point to the directory of the first one.
 	WorkingDir string `json:"workingDir,omitempty"`
 
 	// +optional
--- a/pkg/apis/workspaces/v1alpha1/commands.go
+++ b/pkg/apis/workspaces/v1alpha1/commands.go
@@ -112,9 +112,9 @@ type ExecCommand struct {
 	//
 	// Special variables that can be used:
 	//
-	//  - `$PROJECTS_ROOT`: A path path where projects sources are mounted
+	//  - `$PROJECTS_ROOT`: A path where projects sources are mounted
 	//
-	//  - `$PROJECT_SOURCE`: A path to a project source ($PROJECTS_ROOT/<project-name>).  If there are multiple projects, this will point to the directory of the first one.
+	//  - `$PROJECT_SOURCE`: A path to a project source ($PROJECTS_ROOT/<project-name>). If there are multiple projects, this will point to the directory of the first one.
 	CommandLine string `json:"commandLine,omitempty"`
 
 	// Describes component to which given action relates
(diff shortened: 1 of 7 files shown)


pkg/apis/workspaces/v1alpha1/commands.go
```diff
@@ -109,12 +109,24 @@ type ExecCommand struct {
 	LabeledCommand `json:",inline"`
 
 	// The actual command-line string
+	//
+	// Special variables that can be used:
+	//
+	//  - `$PROJECTS_ROOT`: A path path where projects sources are mounted
```
sleshchenko, line 115:
```suggestion
	//  - `$PROJECTS_ROOT`: A path where projects sources are mounted
```
?

pkg/apis/workspaces/v1alpha1/commands.go
```diff
@@ -109,12 +109,24 @@ type ExecCommand struct {
...
 	CommandLine string `json:"commandLine,omitempty"`
 
 	// Describes component to which given action relates
 	Component string `json:"component,omitempty"`
 
 	// Working directory where the command should be executed
+	//
+	// Special variables that can be used:
+	//
+	//  - `$PROJECTS_ROOT`: A path path where projects sources are mounted
```
sleshchenko, line 127:
```suggestion
	//  - `$PROJECTS_ROOT`: A path where projects sources are mounted
```
?

pkg/apis/workspaces/v1alpha1/commands.go
sleshchenko, line 115: Is $VAR_NAME is the only available format? Or `${VAR_NAME}` is available as well?

pkg/apis/workspaces/v1alpha1/commands.go
```diff
@@ -109,12 +109,24 @@ type ExecCommand struct {
 	LabeledCommand `json:",inline"`
 
 	// The actual command-line string
+	//
+	// Special variables that can be used:
+	//
+	//  - `$PROJECTS_ROOT`: A path path where projects sources are mounted
+	//
+	//  - `$PROJECT_SOURCE`: A path to a project source ($PROJECTS_ROOT/<project-name>).  If there are multiple projects, this will point to the directory of the first one.
```
sleshchenko, line 117:
```suggestion
	//  - `$PROJECT_SOURCE`: A path to a project source ($PROJECTS_ROOT/<project-name>). If there are multiple projects, this will point to the directory of the first one.
```

pkg/apis/workspaces/v1alpha1/commands.go
```diff
@@ -109,12 +109,24 @@ type ExecCommand struct {
...
 	// Describes component to which given action relates
 	Component string `json:"component,omitempty"`
 
 	// Working directory where the command should be executed
+	//
+	// Special variables that can be used:
+	//
+	//  - `$PROJECTS_ROOT`: A path path where projects sources are mounted
+	//
+	//  - `$PROJECT_SOURCE`: A path to a project source ($PROJECTS_ROOT/<project-name>).  If there are multiple projects, this will point to the directory of the first one.
```
sleshchenko, line 129:
```suggestion
	//  - `$PROJECT_SOURCE`: A path to a project source ($PROJECTS_ROOT/<project-name>). If there are multiple projects, this will point to the directory of the first one.
```

pkg/apis/workspaces/v1alpha1/commands.go
kadel: `${VAR_NAME}` should work as well. But I guess this will depend on implementation.

kadel:
> Generally LGTM, though I think the default variable format suggested in the docs should be `${VAR}` instead of `$VAR`. We can leave details up to implementation, but the format listed in documentation determines the minimal spec, and in that regard `${VAR}` is more versatile -- e.g. `${PROJECT_SOURCE}/bin` vs `$PROJECT_SOURCE/bin`.

That makes sense.
I'll change it to `${VAR}`
We just need to make sure that we use this format everywhere in our documentation.

PR #119, opened 2020-08-20
Add folders used by idea IDEs to gitignore

amisevsk:
### What does this PR do?
Add `.idea` folder to `.gitignore` in case anyone else opens the repo in Goland.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,3 +75,5 @@ tags
 .vscode/*
 .history
 # End of https://www.gitignore.io/api/go,vim,emacs,visualstudiocode
+# IDE folders
+.idea


.gitignore
```diff
@@ -75,3 +75,5 @@ tags
 .vscode/*
 .history
 # End of https://www.gitignore.io/api/go,vim,emacs,visualstudiocode
+# IDE folders
```
sleshchenko, line 78: `.vscode/*` is also IDE folder but is out of that group... It changes nothing but maybe it makes sense to regroup gitignore

sleshchenko (approved):

.gitignore
amisevsk: It looks like the main block of the gitignore is auto-generated, and I didn't want to mess around with that part :)

I tried using the same site (which has since rebranded to toptal.com) but it spews a ton of lines that seem unnecessary to me.

davidfestal (approved):

Issue #104, opened 2020-07-27
Support multiple remotes for projects to support forking workflow


amisevsk:
### Is your feature request related to a problem? Please describe.
It's not very common to work on git projects with a single remote; at the very least, a common flow is to have 
```
origin: <your fork>
upstream: <upstream fork>
```
but it's further useful to add close collaborators' forks as remotes in your local copy as well.

It would be useful if the devfile supported assigning multiple remotes to a project.

#### Option 1
Add remotes section
```yaml
spec:
  projects:
    - name: project   
      git:
        location: "https://github.com/amisevsk/web-nodejs-sample.git"
      remotes:
        upstream: "https://github.com/che-samples/web-nodejs-sample.git"
        gh-collaborator: "https://github.com/gh-collaborator/web-nodejs-sample.git"        
```

#### Option 2
Support multiple locations instead of a plain string:
```yaml
spec:
  projects:
    - name: project   
      git:
        location: # alt: 'remotes' ?
          origin: "https://github.com/amisevsk/web-nodejs-sample.git"
          upstream: "https://github.com/che-samples/web-nodejs-sample.git"
          gh-collaborator: "https://github.com/gh-collaborator/web-nodejs-sample.git"  
        primary: origin # optional; defines which fork should be checked out from      
```

#### Option 3
(added from https://github.com/devfile/kubernetes-api/issues/104#issuecomment-665045230 below)
```yaml
spec:
  projects:
    - name: project   
      git:
        branch: foo
           remote: upstream # optional, defaults to origin
        remotes:
          origin: "https://github.com/amisevsk/web-nodejs-sample.git"
          upstream: "https://github.com/che-samples/web-nodejs-sample.git"
          gh-collaborator: "https://github.com/gh-collaborator/web-nodejs-sample.git"  
```

### Additional context
Discussion in Eclipse Che issues: https://github.com/eclipse/che/issues/17449

metacosm:
Maybe something like:
```yaml
spec:
  projects:
    - name: project   
      git:
        branch: foo
           remote: upstream # optional, defaults to origin
        remotes:
          origin: "https://github.com/amisevsk/web-nodejs-sample.git"
          upstream: "https://github.com/che-samples/web-nodejs-sample.git"
          gh-collaborator: "https://github.com/gh-collaborator/web-nodejs-sample.git"  
```

l0rd:
I like @metacosm proposal. @amisevsk wdyt?

amisevsk:
[...]

PR #120, opened 2020-08-20
Add an ability to configure multiple remotes for git project

sleshchenko:
### What does this PR do?
The idea of this PR is to introduce multiple remotes for git/github projects:

Before
```yaml
projects:
  - name: "devworkspace-spec"
    git:
        location: "https://github.com/devfile/api"
        branch: "master"
        # // The tag or commit id to reset the checked out branch to	
        startPoint: "v1.0.0"
```
After
```yaml
projects:
  - name: project   
    git:
      remotes:
        origin: "https://github.com/amisevsk/web-nodejs-sample.git"
        upstream: "https://github.com/che-samples/web-nodejs-sample.git"
        gh-collaborator: "https://github.com/gh-collaborator/web-nodejs-sample.git"  
      checkoutFrom: # mandatory if there is more than one remote
        revision: foo # The revision that should be used checked out. May be branch name, tag or commit id.
                      # Default branch should be checked out if missing.
        remote: upstream # The remote name should be used as init. Required if there are more than one remote configured
```

:notebook:  Reviewing PR in a commit-by-commit manner should be the most convenient.

https://github.com/devfile/api/issues/104

Every sample is updated and valid and schema is regenerated. If more testing is needed - let me know.

#### Docs PR
N/A


l0rd:
@sleshchenko we need `startPoint` to have feature parity between devfile 2.0 and devfile 1.0. I would call it `revision` rather than `startPoint` though.

l0rd:
For consistency, if we add the `revision`, we should remove the `branch` under `cloneFrom` (`branch` is a kind of `revision`) and rename `cloneFrom` to `checkoutFrom`:

```yaml
projects:
  - name: project   
    git:
      remotes:
        origin: "https://github.com/amisevsk/web-nodejs-sample.git"
        upstream: "https://github.com/che-samples/web-nodejs-sample.git"
        gh-collaborator: "https://github.com/gh-collaborator/web-nodejs-sample.git"  
      checkoutFrom: # mandatory if there is more than one remote
        remote: upstream # mandatory if there is more than one remote
        revision: master # mandatory
```

amisevsk:
:+1: to Mario's suggestion of removing branch and replacing with revision.

sleshchenko:
In terms of implementation, what should be behavior in case the specified branch/tag/commitId is not found? Should project be:
1. failed to clone :+1: 
2. left at the default branch state :eye: 
  * propagate a warning to an user, if it always possible?
3. created a new branch with revision (whatever it is, feature-branch, or 4.2.3, or 2ksmrh23k2m4h24) :smile: 

?

amisevsk:
If the reference specified does not exist/cannot be found, we should show a warning and just use the default for that repo. Ideally it's exactly the same as automating 
```
git clone <url>
cd <repo-dir>
git checkout <branch/tag/hash>
```
from a local terminal -- if the checkout I specify does not exist it prints a warning and nothing is done:
```
$ git checkout 370e3e88af
error: pathspec '370e3e88af' did not match any file(s) known to git
$ git checkout nobranch
error: pathspec 'nobranch' did not match any file(s) known to git
# git branch is still "master" for this repo
```

pkg/apis/workspaces/v1alpha1/projects.go
```diff
@@ -78,19 +68,37 @@ type CustomProjectSource struct {
...
 
-	// The tag or commit id to reset the checked out branch to
+	// Defines from what the project should be checked out. Required if there are more than one remote configured
 	// +optional
-	StartPoint string `json:"startPoint,omitempty"`
+	CheckoutFrom *CheckoutFrom `json:"checkoutFrom,omitempty"`
+
+	// The remotes map which should be initialized in the git project. Must have at least one remote configured
+	// +required
+	Remotes map[string]string `json:"remotes,omitempty"`
```
davidfestal, line 91: Why did you use `omitempty` *and* the `+required` command tag ?

pkg/apis/workspaces/v1alpha1/projects.go
```diff
@@ -78,19 +68,37 @@ type CustomProjectSource struct {
...
 
 type ZipProjectSource struct {
-	CommonProjectSource `json:",inline"`
+	// Zip project's source location address. Should be file path of the archive, e.g. file://$FILE_PATH
+	// +required
+	Location string `json:"location,omitempty"`
+
+	// Part of project to populate in the working directory.
+	// +optional
+	SparseCheckoutDir string `json:"sparseCheckoutDir,omitempty"`
```
davidfestal, line 77: Why not keeping the `CommonProjectSource` with only `SparseCheckoutDir` inside it, instead of duplicating `SparseCheckoutDir` ?

davidfestal (commented):
I left 2 minor comments.
I'd have to check how this behaves vs overriding (with the strategic merge patch) before approving.

pkg/apis/workspaces/v1alpha1/projects.go
sleshchenko: I'll remove `omitempty` since it's required.

davidfestal:
@sleshchenko What happens if we reference, in the `checkoutFrom` a remote that is not mentioned in the  `remotes` map ?

BTW for such needs usually a list with a name and a value is more typically used K8S-wise than maps. But I'm also OK with using a map if it works correctly with the strategic merge patch.

pkg/apis/workspaces/v1alpha1/projects.go
sleshchenko: Good catch. I don't see any reason not to keep CommonProject Source. I'll revert it soon

pkg/apis/workspaces/v1alpha1/projects.go
sleshchenko: Ah, it must be optional in the schema, otherwise devfiles with a parent may be not valid.
So, I've made it optional.

pkg/apis/workspaces/v1alpha1/projects.go
sleshchenko: Done now.

sleshchenko:
> @sleshchenko What happens if we reference, in the checkoutFrom a remote that is not mentioned in the remotes map ?

@davidfestal I suppose implementation should tell a user that Devfile is not valid. WDYT? Should I declare it somewhere, like field description?

> BTW for such needs usually a list with a name and a value is more typically used K8S-wise than maps. But I'm also OK with using a map if it works correctly with the strategic merge patch.

Let's discuss it on the upcoming call. So,
```yaml
projects:
  - name: project   
    git:
      remotes:
        origin: "https://github.com/amisevsk/web-nodejs-sample.git"
        upstream: "https://github.com/che-samples/web-nodejs-sample.git"
      checkoutFrom: ...
```
vs
```yaml
projects:
  - name: project   
    git:
      remotes:
        - name: origin
          location: https://github.com/amisevsk/web-nodejs-sample.git
        - name: upstream
          location: https://github.com/che-samples/web-nodejs-sample.git
      checkoutFrom: ...
```

davidfestal:
Seems you should have added `provision-remote` in the `patches` sub-folder, and not the `merges`.

- `patches` folder is for strategic merge patches (== parent or plugin overriding of *existing* elements)
- `merges` folder is for testing the merge of flattened parent and plugins with the main child devfile body, checking that this brings no duplicate keys.

devfile-support/samples/sample-devfile.yaml
```diff
@@ -4,8 +4,11 @@ metadata:
 projects:
   - name: "my-project"
     git:
-      location: "https://github.com/che-incubator/devworkspace-api"
-      branch: "master"
+      remotes:
+        origin: "https://github.com/devfile/api"
+      checkoutFrom:
+        revision: "master"
+        remote: origin
```
davidfestal, line 11: Is it mandatory to have the `origin` remote in the `checkoutFrom`, since there is only one origin ?

pkg/apis/workspaces/v1alpha1/projects.go
davidfestal: Ah, I see. So in order to support optional fields only in overrides, but still have them mandatory in the main definition, you would:
- Make it `optional` on the GO struct as you just did
- Still add a `required` constraint in the schema with [a transformation rule similar to this one](https://github.com/sleshchenko/devworkspace-api/blob/multipleRemotes/schema-transformation-rules/devworkspace/addRequiredFields.json#L23)

davidfestal: In the future I plan to generate dedicated Go structures for overrides. But for now that the simplest way.

devfile-support/samples/sample-devfile.yaml
sleshchenko: it's not mandatory, it's just here for sample.

pkg/apis/workspaces/v1alpha1/projects.go
sleshchenko: It's done in https://github.com/devfile/api/pull/120/commits/8e6766364ef24192dfb8a936f549bd6560e94ea6

davidfestal (approved):

Issue #121, opened 2020-08-20
Check samples with JSON schemas during the build


sleshchenko:
**Is your feature request related to a problem? Please describe.**
It might be not obvious which samples must be updated after JSON schema is updated.

**Describe the solution you'd like**
It would be nice if we make checking our samples with JSON schemas during the build, and which is already run as PR check.

**Describe alternatives you've considered**
N/A

**Additional context**
Che Plugin Registry already has such check on their build https://github.com/eclipse/che-plugin-registry/blob/master/build/scripts/check_metas_schema.sh

I've tested my PR https://github.com/devfile/api/pull/120 with jsonschema-cli:
```
export BASE_DIR=.
jsonschema-cli validate "${BASE_DIR}/schemas/devworkspace.json" "${BASE_DIR}/samples/custom.devworkspace.yaml"
jsonschema-cli validate "${BASE_DIR}/schemas/devworkspace.json" "${BASE_DIR}/samples/example.devworkspace.yaml"
jsonschema-cli validate "${BASE_DIR}/schemas/devworkspace.json" "${BASE_DIR}/samples/with-nodejs-parent-devfile.devworkspace.yaml"
jsonschema-cli validate "${BASE_DIR}/schemas/devworkspace.json" "${BASE_DIR}/samples/with-nodejs-parent-template.devworkspace.yaml"

jsonschema-cli validate "${BASE_DIR}/schemas/devworkspace-template.json" "${BASE_DIR}/samples/nodejs.devworkspacetemplate.yaml"

jsonschema-cli validate "${BASE_DIR}/schemas/devfile.json" "${BASE_DIR}/devfile-support/samples/nodejs-stack.devfile.yaml"
jsonschema-cli validate "${BASE_DIR}/schemas/devfile.json" "${BASE_DIR}/devfile-support/samples/sample-devfile.yaml"
jsonschema-cli validate "${BASE_DIR}/schemas/devfile.json" "${BASE_DIR}/devfile-support/samples/job-apply-example.devfile.yaml"
jsonschema-cli validate "${BASE_DIR}/schemas/devfile.json" "${BASE_DIR}/devfile-support/samples/with-nodejs-parent.devfile.yaml"
jsonschema-cli validate "${BASE_DIR}/schemas/devfile.json" "${BASE_DIR}/devfile-support/samples/spring-boot-http-booster-devfile.yaml"
[...]

PR #124, opened 2020-08-24
Check samples with json schemas

davidfestal:
### What does this PR do?

This PR add a script that validates sample files against the corresponding json schemas

#121

Yes, tested locally.

### DIFF
--- a/validate-samples.sh
+++ b/validate-samples.sh
@@ -1,0 +1,71 @@
+#!/bin/bash
+#
+# Copyright (c) 2020 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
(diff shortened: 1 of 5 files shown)


validate-samples.sh
```diff
@@ -0,0 +1,71 @@
...
+fi
+
+BASE_DIR=$(cd "$(dirname "$0")" && pwd)
+
+rm -r validate-output.txt &> /dev/null
+
+for schema in "devfile" "devworkspace" "devworkspace-template"
+do
+  schemaPath="./schemas/${schema}.json"
+  devfiles=$(jq -r ".\"yaml.schemas\".\"${schemaPath}\"[]" .theia/settings.json)
```
amisevsk, line 50: This line is hard to read, can it be simplified?
```bash
devfiles=$(jq -r '."yaml.schemas"."${schemaPath}"[]' --arg schemaPath=$schemaPath .theia/settings.json)
```
or similar?

davidfestal: The proposed change didn't seem to work (using an arg as an object key isn't straightforward).
But I still simplified the expression and removed the `\"` parts which made it hard to read.

Issue #51, opened 2020-05-20
Introduce outerloop 'dockerfile' build & deploy


sbose78:
(EDITED - check edit history for prior revisions and discussion of the opening post) 

For devfile 2.2.0. Check the proposal at https://github.com/devfile/api/blob/main/docs/proposals/devfile/outer-loop-build-and-deploy.md

A stack may require expressing the outer-loop build guidance ( source code to image ) in the form of a Dockerfile build, s2i build, etc.

```
    variables:
      myimage: myimagename
    components:
      - name: mydockerfileimage
        image:
          imageName: {{myimage}}
          dockerfile:
            buildContext: ${PROJECTS_ROOT}/build
            uri: Dockerfile
            args: [ "arg1", "arg2", "arg3" ]
            rootRequired: false

```

The other dockerfile src are available in the proposal doc.

Once the runtime image has been built, we can deploy it via the deploy group command:

```
components:
  - name: outerloop-deploy
    kubernetes:
      uri: deployment-manifest.yaml
commands:
  - id: deployk8s
    apply:
      component: outerloop-deploy
      group:
        kind: deploy
        isDefault: true
```  

The build tool mechanism like `kaniko` or `buildah` will be up to the tools and hence will not be a part of the spec.

EnriqueL8:
Would there be an idea here to add a build image to be used to consume that Dockerfile? I am thinking of a Buildah image or a Kaniko image that would both be used locally and in a pipeline? Would this be something the stack creator should add, or the tool consuming the devfile decide?

l0rd:
In Che, Devfile components are created before the project gets cloned: we do not have access to `${PROJECTS_ROOT}/build/Dockerfile` yet when we need to create the `Dockerfile` component. 

If we want to support this use case we need to agree on the workspace creation flow: cloning of the project should happen before everything else.

On the other hand I think we should consider publishing Dockerfiles and ohter components definitions in a devfile registry so that we may reference as:

```bash
components:
  - id: build-node-app      
    dockerfile:
        id: mycompany/my-node-stack-dockerfile/v2.2.2
        args: []
```

sbose78:
> In Che, Devfile components are created before the project gets cloned: we do not have access to ${PROJECTS_ROOT}/build/Dockerfile yet when we need to create the Dockerfile component.

Should we annotate this component to ensure that building this image is not a pre-req to the workspace starting up in Che?

[...]

PR #127, opened 2020-09-10
Add Build Guidances

wtam2018:
### What does this PR do?
This PR adds `BuildGuidances` which includes `Dockerfile` and `SourceToImage` for build strategies.   It is added to top level so that they can be shared by multiple projects.

#51 

Not applicable.


pkg/apis/workspaces/v1alpha2/buildGuidances.go
```diff
@@ -0,0 +1,51 @@
...
+
+	// +optional
+	// Allows specifying a builder image to initiate s2i (SourceToImage) build
+	SourceToImage *SourceToImage `json:"s2i,omitempty"`
+}
+
+type Dockerfile struct {
+
+	// Name that allows referencing a build guidance
+	Name string `json:"name"`
```
sleshchenko, line 19: In terms of overriding does it makes sense move the name to the upper level, like we do with components:
```
components:
  - name: maven
    container: ....
```
? I'm not sure about just a question

davidfestal: The name is mandatory at the upper-level to identity elements in a list. When the field is the patchMergeKey of the list.
Here there is only 1 build guidance in a project afaik.

However I see you defined a union @wtam2018 Are we sure that the 2 values (dockerfile or s2i) are really mutually-exclusive ?

wtam2018: @davidfestal I'll have union removed dockerfile and s2i are not mutually excluded.   I should probably rename guidance to its plural form.

wtam2018: @davidfestal @sleshchenko Based on the feedbacks from our previous meeting, build guidance should be similar and at the same level as component/command.  I will rework this and take build guidance out of project.

pkg/apis/workspaces/v1alpha2/buildGuidances.go
```diff
@@ -0,0 +1,65 @@
...
+	SourceToImage *SourceToImage `json:"s2i,omitempty"`
+}
+
+type Dockerfile struct {
+	// Dockerfile location which can be an URL or a path relative to buildContext
+	DockerfileLocation string `json:"dockerfileLocation"`
+
+	// +optional
+	// Path of source directory to establish build context.  Default to the top level directory.
+	BuildContext string `json:"buildContext,omitempty"`
```
maysunfaisal, line 44: I am not sure if we want to continue to work on this branch for Build Guidance or open a new PR but for record, I had a question:

What is the difference between this `BuildContext` and the one provided from the Console POC import from devfile form https://github.com/openshift/console/pull/6321

Screen Shot 2020-11-16 at 12 29 19 PM

Issue #128, opened 2020-09-15
remove sparseCheckoutDir


kadel:
Currently, it is defined as

```
             "sparseCheckoutDir": {
                "description": "Part of project to populate in the working directory.",
                "type": "string",
                "markdownDescription": "Part of project to populate in the working directory."
              }
```

As I understand this, if I have a repo with a structure like this.
```
- file0.txt
- /dir1/
    - file1.txt
- /dir2/
   - file2.txt
```
and I do `sparseCheckoutDir: dir2` I should end up with directory with `/file2.txt`.
Problem is that this is not how git sparse checkout works, and as far as I know, you can never clone the repo like this.

The way how [sparse checkout in git](https://github.blog/2020-01-17-bring-your-monorepo-down-to-size-with-sparse-checkout/) works is that it can populate only certain directories, but the directory structure remains.

For example, If I have a repo with the above structure and I want to do sparse checkout if a `/dir2` I would end up with a directory structure like this:
```
- file0.txt
- /dir2/
   - file2.txt
``` 
But what we want to do here is to end up with `/file2.txt` and that is not what sparse checkout does.

I would actually consider completely removing this field from Devfile, It will create a lot of confusion and problems.

kadel:
from meeting:
" We are going to remove the sparseCheckoutDir from `project` and add it to `starterProject` and rename it to something like `subDir`. "

l0rd:
@kadel can we close this issue?

elsony:
Done

PR #129, opened 2020-09-15
rename sparseCheckoutDir from StarterProject

kadel:
### What does this PR do?

rename sparseCheckoutDir to subDir and use it only in StarterProject

https://github.com/devfile/api/issues/128


sleshchenko (approved):
LGTM
@davidfestal please validate as well

l0rd (approved):

davidfestal (approved):
OK for me.

Side question: I'm still wondering whether we should keep both the `git` and `github` project source types (that until now have the same structure).
Do we want to keep both for 2.0.0 or only keep `git` for now, and add the `github` union value in the future if really necessary ?

cc @l0rd

kadel:
> Do we want to keep both for 2.0.0 or only keep `git` for now, and add the `github` union value in the future if really necessary ?

This also crossed my mind. I would actually prefer removing `github` for now. We can re-introduce it later once there are some extra github feature that this can use.

sunix:
Sorry just been throught this one. In Che, sparseCheckoutDir is doing a git sparsecheckout. There is no confusion.
What it also do is keep the git directory so user can still `git commit` and so on.
In real life projects (not just testing samples), user would usually commit/push changes of a project.

kadel:
> Sorry just been throught this one. In Che, sparseCheckoutDir is doing a git sparsecheckout. There is no confusion.
> What it also do is keep the git directory so user can still `git commit` and so on.
> In real life projects (not just testing samples), user would usually commit/push changes of a project.

In that case, we could keep `sparseCheckoutDir` in `project` (and do proper sparse checkout) and in `StarterProject` use `subDir`.

In this case, it might make sense to make  `sparseCheckoutDir` a list to allow specify multiple dirs to checkout.

sunix:
> In that case, we could keep `sparseCheckoutDir` in `project` (and do proper sparse checkout) and in `StarterProject` use `subDir`.
> 
> In this case, it might make sense to make `sparseCheckoutDir` a list to allow specify multiple dirs to checkout.

+1

pkg/apis/workspaces/v1alpha2/projects.go
```diff
@@ -10,14 +10,26 @@ type Project struct {
...
 
+	// Populate the project sparsely with selected directories.
+	// +optional
+	SparseCheckoutDirs []string `json:"sparseCheckoutDirs,omitempty"`
+
 	ProjectSource `json:",inline"`
 }
+
 type StarterProject struct {
-	Project `json:",inline"`
```
kadel, line 16: This change also removes `ClonePath` from. I'm not sure how much sense it makes for `StarterProjects`. If we have use-case for this we can easily add it back (it will be backward compatible change) removing it in the future would be harder.

kadel:
updated,
@sunix @davidfestal  please let me know if this works for you.

amisevsk (commented):
Looks fine in general but I'm still not clear on the distinction between `sparseCheckoutDirs` and `subDir`. At first glance, it looks like we have similar functionality with two different names.

pkg/apis/workspaces/v1alpha2/projects.go
davidfestal: +1 for me

sunix:
hey @kadel,
We had that use case: several projects in the same git repo. For instance 1 maven project in one subpath and 1 nodejs project in another subpath. For che, that was what one of the proposal: https://github.com/eclipse/che/issues/15347
Let me know if this is the purpose of subdir. But if it is the case, it should be able to take multiple directories.

davidfestal:
@kadel didn't you need that before the ODO release (since it would be a breaking change) ? Is there a reason it hasn't been merged ?

kadel:
> @kadel didn't you need that before the ODO release (since it would be a breaking change) ? Is there a reason it hasn't been merged ?

We worked around this in odo by removing `sparceChecoutDir` from all devfiles used by odo.

PR #130, opened 2020-09-17
Locate all samples into one folder

sleshchenko:
### What does this PR do?
I found it a bit confusing to have devworkspace and devfile samples have in different places, probably there is a motivation behind it but I missed that.
So, this PR locates all samples into one folder. It's devfile-support/samples folder.

### Alternative
I don't really understand the purpose of devfile-support folder and it may even make sense to move all samples into `/sample` directory in the project root.

N/A

Made sure that validate-samples script works as expected

#### Docs PR
https://github.com/devfile/docs/pull/8


davidfestal:
Probably the purpose of having the devfile samples in a dedicated folder used to make more sense at the beginning of this repository, when the main purpose of this repository was to provide a Kubernetes API for the future Che DevWorkspace Controller, and also *derive* the  Devfile Json schema from it.

Now that the whole purpose of the repository is to define the main structure of the Devfile as well as associated K8S APIs, we might want to consider that the Devfile schema itself and DevWorkspace-related variants are just several ways to leverage the same core specification in different contexts.

What I'd recommend then, would be to have a single top-level folder `samples`, but still with one sub-folder for each type of yaml file:
- Devfile
- DevWorkspace
- DevWorkspaceTemplate

This would still allow the Devfile docs website to reference a folder with only devfiles, like it was until now: https://devfile.github.io/devfile/index.html#_devfile_v2_0_examples

davidfestal (changes requested):
see my comment [here](https://github.com/devfile/api/pull/130#issuecomment-694185713)

sleshchenko:
@davidfestal Could you please review it one more time, as well as docs related PR https://github.com/devfile/docs/pull/7 ?

amisevsk (approved):
Makes sense to me but I'd defer to David on if this is appropriate.

davidfestal (approved):
Both PRs approved (also the docs PR).
Thanks !

PR #133, opened 2020-09-23
Use controller-tools generators to generate GO files, CRDs, and Json schemas

davidfestal:
### What does this PR do?

This PR changes the way CRDs and Json schemas are generated from GO source code.
The new generation mechanism is based on the `controller-tools` (aka kube-builder) generator pattern and has several advantages:
- it provides an easy access to the AST of Go source code, and especially an easy way to read and use GO comment markers heavily used by kube-builder.
- it allows using a single generation framework and finally a single generation command-line to also generate a number of GO source files that until now were written manually (which was quite cumbersome very error-prone).
- Finally it also allows removing the dependency to the Operator-SDK CLI tools, and only relying on a given version (thelast one) of the `kube-builder` `sigs.k8s.io/controller-tools` GO module.

### How to review this PR

This PR touches a big amount of of files, and thus should be reviewed **commit-by-commit**.
It mainly consists of the following main series of changes:
- Some preliminary lightweight and non-impacting cleaning in the repo
- The addition of a new nested `generator` module to create the new generators. This is only new code, and can be reviewed as a whole.
- The changes in the existing repository to switch to the use of the new generators
- The removal of the `operator-sdk`, which involves to the removal or change of a big amount of files in the main `vendor` folder.

The PR doesn't have any related issue

Yes, it is tested.

You can [test it under che.openshift.io](https://che.openshift.io/dashboard/#/load-factory?url=https:%2F%2Fgithub.com%2Fdavidfestal%2Fapi%2Ftree%2Fstart-go-generator&override.projects.api.source.location=https://github.com/davidfestal/api&override.projects.api.source.branch=start-go-generator)


generator/overrides/gen.go
```diff
@@ -0,0 +1,435 @@
...
+	if err != nil {
+		root.AddError(err)
+		return
+	}
+	if n < len(outBytes) {
+		root.AddError(io.ErrShortWrite)
+	}
+}
+
+func updateComments(commentedNode ast.Node, commentGroup *ast.CommentGroup, keepRegexp string, skipRegexp string, additionalLines ...string) *ast.CommentGroup {
```
amisevsk, line 413: I think `dropRegex` might be clearer than `skipRegex` (does skip mean "skip including this line" or "skip processing this line"?)

generator/overrides/gen.go
```diff
@@ -0,0 +1,435 @@
...
+	if commentGroup == nil {
+		commentGroup = &ast.CommentGroup{}
+	}
+	commentsToKeep := []*ast.Comment{}
+	for _, comment := range commentGroup.List {
+		if matched, _ := regexp.MatchString(` *//`+keepRegexp+` *`, comment.Text); matched {
+			if matched, _ := regexp.MatchString(` *//`+skipRegexp+` *`, comment.Text); !matched {
```
amisevsk, line 420:
```suggestion
		if keep, _ := regexp.MatchString(` *//`+keepRegexp+` *`, comment.Text); keep {
			if drop, _ := regexp.MatchString(` *//`+skipRegexp+` *`, comment.Text); !drop {
```

generator/overrides/gen.go
amisevsk, line 413: I think (but could be wrong) that this function *prepends* `additionalLines` to the comment, rather than appends. If so this should be mentioned in a doc.

generator/overrides/gen.go
```diff
@@ -0,0 +1,435 @@
...
+					moreTypesToAdd = append(moreTypesToAdd, *fieldTypeToProcess)
+				}
+			}
+			return true
+		},
+		func(*astutil.Cursor) bool { return true },
+	).(*ast.GenDecl)
+
+	if len(errors) > 0 {
+		return nil, moreTypesToAdd, errors // Check there is only 1 typespec in it or trigger an error
```
amisevsk, line 384: Is this commit still needed? Looks like this is checked above.

generator/overrides/gen.go
```diff
@@ -0,0 +1,435 @@
...
+
+				if shouldSkip(overridesMarker) {
+					cursor.Delete()
+					return true
+				}
+
+				if overridesMarker.Description != "" {
+					astField.Doc = updateComments(
+						astField, astField.Doc,
+						` *\+[^ ][^ ]*.*`,
```
amisevsk, line 292:
```suggestion
						` *\+[^ ]+.*`,
```

generator/overrides/gen.go
```diff
@@ -0,0 +1,435 @@
...
+						for _, f := range typeToOverride.Fields {
+							pos := f.RawField.Pos()
+							fieldChange := fieldChanges[f.RawField.Pos()]
+							if pos != cursor.Node().Pos() &&
+								!shouldSkip(fieldChange.OverrideMarker) {
+								enumValues = append(enumValues, fieldChange.FieldInfo.Name)
+							}
+						}
+						kubebuilderAnnotation := "+kubebuilder:validation:Enum=" + strings.Join(enumValues, ";")
+						astField.Doc = updateComments(astField, astField.Doc, `.*`, ` *`+regexp.QuoteMeta("+kubebuilder:validation:Enum=")+`.*`, kubebuilderAnnotation)
```
amisevsk, line 354: Any idea why the `+kubebuilder:validation:Enum` tags end up at the top of comment blocks? I can't quite figure it out.

generator/schemas/gen.go
```diff
@@ -0,0 +1,308 @@
...
+			parser.NeedFlattenedSchemaFor(typeIdent)
+			currentJSONSchema, found := parser.FlattenedSchemata[typeIdent]
+			if !found {
+				root.AddError(fmt.Errorf("Json schema for type " + typeIdent.Package.Name + "/" + typeIdent.Name + " could not be generated"))
+				continue
+			}
+
+			fieldsToSkip := []string{}
+			if schemaGenerateMarker.OmitCustomUnionMembers {
+				fieldsToSkip = append(fieldsToSkip, "Custom")
```
amisevsk, line 154: This seems to drop a number of fields from the schemas present in the master branch, is this intentional?

generator/schemas/gen.go
```diff
@@ -0,0 +1,308 @@
...
+					schema.AdditionalProperties != nil {
+					return
+				}
+				schema.AdditionalProperties = &apiext.JSONSchemaPropsOrBool{
+					Allows: schema.XPreserveUnknownFields != nil && *schema.XPreserveUnknownFields,
+				}
+				return
+			})
+
+			// Fix descriptions to have them Markdown compatible
```
amisevsk, line 189: There's an identical comment above (line 159); why can't we do both at the same time?

generator/schemas/gen.go
```diff
@@ -0,0 +1,308 @@
...
+			}
+
+			(&currentJSONSchema).Title = schemaGenerateMarker.Title
+
+			jsonSchema, err := json.MarshalIndent(&currentJSONSchema, "", "  ")
+			if err != nil {
+				return err
+			}
+
+			toAddMardownDescriptions := orderedmap.New()
```
amisevsk, line 210:
```suggestion
			toAddMarkdownDescriptions := orderedmap.New()
```

amisevsk (commented):
Really good work

amisevsk:
@davidfestal I'm wondering (since we're getting deep into details on generation now) -- is it possible to use block/flow scalars for descriptions? E.g. [this](https://github.com/davidfestal/api/blob/a28f9f55586ff9fa17ef1e666be7121714e8a12c/crds/workspace.devfile.io_devworkspaces.yaml#L76-L84) could be rewritten (I think) to:
```yaml
description: |
    Command that consists in applying a given component definition,
    typically bound to a workspace event.
    
    For example, when an `apply` command is bound to a `preStart` event,
    and references a `container` component, it will start the container as a
    K8S initContainer in the workspace POD, unless the component has its
    `dedicatedPod` field set to `true`.
    
    When no `apply` command exist for a given component,
    it is assumed the component will be applied at workspace start
    by default.
```

diffJsons.sh
```diff
@@ -0,0 +1,44 @@
...
+
+onError() {
+  rm "$file1"
+  rm "$file2"
+}
+trap 'onError' ERR
+
+jq -S '.' "$1" > "$file1"
+jq -S '.' "$2" > "$file2"
+diff "$file1" "$file2"
```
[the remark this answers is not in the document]
davidfestal: You're right here. In fact the purpose of the script was more to fix the order to properties in Json objects. Which was the main problem to compare old and new json schemas.
But if you think that sort also arrays can be useful, let's do this change.
Waiting for your confirmation here :-)

generator/crds/gen.go
```diff
@@ -0,0 +1,170 @@
+package crds
```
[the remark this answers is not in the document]
davidfestal: In fact no file has a license in the whole repo as for now afaict.
Do you think we can keep this for a next PR ?

build.sh
```diff
@@ -16,96 +16,42 @@ RED='\033[0;31m'
...
+  cd "${CURRENT_DIR}"
 }
 trap 'onError' ERR
 
-sed -i -e '/"description":/s/ \\t/\\n/g' -e '/"description":/s/ \\n - /\\n- /g' -e '/"description":/s/ \\n \([^-]\)/\\n\\n\1/g' "${BASE_DIR}/schemas/devworkspace-template.json" "${BASE_DIR}/schemas/devworkspace.json"
+"${BASE_DIR}"/build-generator.sh
+
+cd "${BASE_DIR}"
+
+echo "Generating Plugin Overrides"
```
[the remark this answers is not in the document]
davidfestal: Fix in commit https://github.com/davidfestal/api/commit/79189725909aa0266d32a52299edb94bcae68541

devfile.api.code-workspace
```diff
@@ -0,0 +1,112 @@
...
+            "autoComplete": true,
+            "documentSymbols": true,
+            "findReferences": true,
+            "format": false,
+            "goToDefinition": true,
+            "goToTypeDefinition": true,
+            "hover": true,
+            "signatureHelp": true,
+            "rename": true,
+            "workspaceSymbols": true,
```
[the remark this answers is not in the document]
davidfestal: Fix in commit: https://github.com/devfile/api/pull/133/commits/1c92ff2ad4b0254d911061532be1aeebfc083bd4

pkg/apis/workspaces/v1alpha2/doc.go
```diff
@@ -2,4 +2,5 @@
 // +k8s:deepcopy-gen=package,register
 // +k8s:openapi-gen=true
 // +groupName=workspace.devfile.io
+// +devfile:jsonschema:version=2.0.0-alpha2
```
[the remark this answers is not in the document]
davidfestal: Yeah, hopefully this PR would help bootstrapping the version discussion.
However let me point out that this PR doesn't impose much constraint on versioning: It just assumes that K8S API versions would evolves more slowly than Json schema versions, which was already agreed on the last time we discussed about it, iirc. In other words, several Json Schema versions could correspond to a single K8S API.
However this lets us free to opt for a 1:1 mapping between KS API version and Json schema version, even though I personnally think it might be hard to manage over time.

schemas/latest/dev-workspace-template-spec.json
```diff
@@ -0,0 +1,3424 @@
+{
```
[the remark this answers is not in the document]
davidfestal: yeah. In fact I'm now deriving the name of the Json schema from the name of the annotated Go struct. For this I use a simple conversion from `CamleCase` to `kebab-case`. That's why `DevWorkspace`, now produces the `dev-workspace` json schema name.
If you think it should be changed I can change it, but at first I preferred keeping the code simple and the name mapping obvious.

generator/genutils/utils.go
```diff
@@ -0,0 +1,138 @@
...
+					discriminatorProp.Enum != nil &&
+					len(discriminatorProp.Enum) != 0 {
+					oneOf := []apiext.JSONSchemaProps{}
+					newEnumValues := []apiext.JSON{}
+				enumValueLoop:
+					for _, enumValue := range discriminatorProp.Enum {
+						fieldName := strings.Trim(string(enumValue.Raw), `"`)
+						unionMemberProperty := strcase.ToLowerCamel(fieldName)
+						if _, foundUnionMember := schema.Properties[unionMemberProperty]; !foundUnionMember {
+							continue discriminatorLoop
```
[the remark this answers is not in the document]
davidfestal: The discriminator enum contains values that do not correspond to any enum field. So we just skip the `oneOf` creation here since this union definition seems invalid.
TBF we should be more strict and implement real union validity checks, but I didn't do it for now.
I'll add a comment in the code to explain the logic here at least.

generator/genutils/utils.go
```diff
@@ -0,0 +1,138 @@
...
+		if schema.Type != "object" {
+			return
+		}
+
+		if len(schema.Properties) > 0 {
+		discriminatorLoop:
+			for _, discriminator := range unionDiscriminators {
+				discriminatorPropertyName := strcase.ToLowerCamel(discriminator.Name)
+				if discriminatorProp, found := schema.Properties[discriminatorPropertyName]; found &&
+					discriminatorProp.Enum != nil &&
```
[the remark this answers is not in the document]
davidfestal: Fix in commit https://github.com/devfile/api/pull/133/commits/3465390b28039c68fb26c1db19b1429bc55dc2c1

generator/genutils/utils.go
```diff
@@ -0,0 +1,138 @@
...
+					len(discriminatorProp.Enum) != 0 {
```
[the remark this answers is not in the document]
davidfestal: Fixed in commit https://github.com/devfile/api/pull/133/commits/3465390b28039c68fb26c1db19b1429bc55dc2c1

generator/genutils/utils.go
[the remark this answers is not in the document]
davidfestal: Fixed in commit https://github.com/devfile/api/pull/133/commits/3465390b28039c68fb26c1db19b1429bc55dc2c1

generator/genutils/utils.go
```diff
@@ -0,0 +1,138 @@
...
+func AddUnionOneOfConstraints(jsonSchema *apiext.JSONSchemaProps, unionDiscriminators []markers.FieldInfo, removeDiscriminators bool, fieldsToSkip ...string) {
+	mainVisitor := func(schema *apiext.JSONSchemaProps) (newVisitor Visitor, stop bool) {
+		if schema == nil {
+			return
+		}
+		if schema.Type != "object" {
+			return
+		}
+
+		if len(schema.Properties) > 0 {
```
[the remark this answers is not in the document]
davidfestal: Fixed in commit https://github.com/devfile/api/pull/133/commits/3465390b28039c68fb26c1db19b1429bc55dc2c1

generator/crds/gen.go
```diff
@@ -0,0 +1,170 @@
...
+			if err := ctx.WriteYAML(fileName, crd); err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+func (g Generator) writeFoFile(filename string, ctx *genall.GenerationContext, root *loader.Package, writeContents func(*bytes.Buffer)) error {
```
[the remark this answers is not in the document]
davidfestal: It is not used, you're right.

generator/crds/gen.go
```diff
@@ -0,0 +1,170 @@
...
+			}
+		default:
+			root.AddError(fmt.Errorf("The package should have a valid 'groupName' annotation"))
+			return nil
+		}
+
+		unionDiscriminators, found := unionDiscriminatorsByGV[groupVersion]
+		if !found {
+			unionDiscriminators = []markers.FieldInfo{}
+		}
```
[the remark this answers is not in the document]
davidfestal: Fixed in commit https://github.com/devfile/api/pull/133/commits/23fc981d8b177c2d4436bc891be554e990e16088

generator/crds/gen.go
```diff
@@ -0,0 +1,170 @@
...
+		versionedCRDs := make([]interface{}, len(crdVersions))
+		for i, ver := range crdVersions {
+			conv, err := crd.AsVersion(crdRaw, schema.GroupVersion{Group: apiext.SchemeGroupVersion.Group, Version: ver})
+			if err != nil {
+				return err
+			}
+			versionedCRDs[i] = conv
+		}
+
+		for i, crd := range versionedCRDs {
```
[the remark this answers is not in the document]
davidfestal: Fixed in commit https://github.com/devfile/api/pull/133/commits/23fc981d8b177c2d4436bc891be554e990e16088

generator/crds/gen.go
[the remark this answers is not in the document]
davidfestal: Fixed in commit https://github.com/devfile/api/pull/133/commits/23fc981d8b177c2d4436bc891be554e990e16088

generator/overrides/gen.go
```diff
@@ -0,0 +1,435 @@
...
+		}
+
+		config := printer.Config{
+			Tabwidth: 2,
+			Mode:     printer.UseSpaces,
+		}
+
+		buf := new(bytes.Buffer)
+		buf.WriteString("package " + root.Name)
+		buf.WriteString("\n\n")
```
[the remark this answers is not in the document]
davidfestal: Probably because it's the first one I worked on :-)

generator/overrides/gen.go
```diff
@@ -0,0 +1,435 @@
...
+
+//go:generate go run sigs.k8s.io/controller-tools/cmd/helpgen generate:headerFile=../header.go.txt,year=2020 paths=.
+
+// +controllertools:marker:generateHelp:category=Overrides
+
+// FieldOverridesInclude drives whether a field should be overriden in devfile parent or plugins
+type FieldOverridesInclude struct {
+	// Omit indicates that this field cannot be overriden in a devfile plugin.
+	Omit bool `marker:",optional"`
+	// OmmitInPlugin indicates that this field cannot be overriden in a devfile plugin.
```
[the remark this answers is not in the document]
davidfestal: True. that's an error.

generator/overrides/gen.go
```diff
@@ -0,0 +1,435 @@
...
+		ctx.Checker.Check(root, func(node ast.Node) bool {
+			// ignore interfaces
+			_, isIface := node.(*ast.InterfaceType)
+			return !isIface
+		})
+
+		root.NeedTypesInfo()
+
+		var rootStructToOverride *markers.TypeInfo
+		g.packageTypes = map[string]*markers.TypeInfo{}
```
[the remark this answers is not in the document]
davidfestal: Why ?

generator/overrides/gen.go
davidfestal: Because it is removed by the drop regexp (in case it exists but is wrong), and is added again as an additional line.
And as you guessed, `updateComments` prepends additional lines.

generator/schemas/gen.go
davidfestal: Yes, it omits all the `Custom` fields that are in unions. But it only does som when the `omitCustomUnionMarkers` field of the `devfile:jsonschema:generate` annotation is set to `true`, which is only the case for the `Devfile` struct.
This finally does the same as what the old script [`removeCustomElements.js`](https://github.com/devfile/api/pull/133/commits/a0161dc6fd553ff4b86c4c94e92fdfbc0b7e455e#diff-636447eafa5d31703c7bc90711bc912a) used to do.

generator/interfaces/gen.go
```diff
@@ -0,0 +1,203 @@
...
+}
+`)
+			}
+		})
+	}
+
+	return nil
+}
+
+func (g Generator) writeFoFile(filename string, ctx *genall.GenerationContext, root *loader.Package, writeContents func(*bytes.Buffer)) error {
```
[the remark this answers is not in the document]
davidfestal: You're right. I renamed and commented this function and extracted it in the `genutils` package.
In commit https://github.com/devfile/api/pull/133/commits/37c1e341bb7bb541566dbf0b63a536b64e1b6028

generator/overrides/gen.go
```diff
@@ -0,0 +1,435 @@
+package overrides
+
+import (
+	"bytes"
+	"fmt"
+	"go/ast"
+	"go/format"
+
+	//	"go/format"
```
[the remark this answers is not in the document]
davidfestal: Fixed in commit https://github.com/devfile/api/pull/133/commits/37c1e341bb7bb541566dbf0b63a536b64e1b6028

generator/overrides/gen.go
```diff
@@ -0,0 +1,435 @@
...
+	"github.com/elliotchance/orderedmap"
+)
+
+//go:generate go run sigs.k8s.io/controller-tools/cmd/helpgen generate:headerFile=../header.go.txt,year=2020 paths=.
+
+// +controllertools:marker:generateHelp:category=Overrides
+
+// FieldOverridesInclude drives whether a field should be overriden in devfile parent or plugins
+type FieldOverridesInclude struct {
+	// Omit indicates that this field cannot be overriden in a devfile plugin.
```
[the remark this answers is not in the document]
davidfestal: Fixed in commit https://github.com/devfile/api/pull/133/commits/37c1e341bb7bb541566dbf0b63a536b64e1b6028

generator/overrides/gen.go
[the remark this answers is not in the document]
davidfestal: I used the new function extracted to `genutils`, so that the approach is the same for all GO-source-related generators.

Fixed in commit https://github.com/devfile/api/pull/133/commits/37c1e341bb7bb541566dbf0b63a536b64e1b6028

generator/overrides/gen.go
```diff
@@ -0,0 +1,435 @@
...
+			root.AddError(err)
+			return err
+		}
+		g.writeOut(ctx, root, outContents)
+	}
+
+	return nil
+}
+
+type typeToProcess struct {
```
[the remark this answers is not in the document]
davidfestal: Fixed in commit https://github.com/devfile/api/pull/133/commits/37c1e341bb7bb541566dbf0b63a536b64e1b6028

generator/overrides/gen.go
davidfestal: Fixed in commit https://github.com/devfile/api/pull/133/commits/37c1e341bb7bb541566dbf0b63a536b64e1b6028

generator/overrides/gen.go
davidfestal: Fixed in commit https://github.com/devfile/api/pull/133/commits/37c1e341bb7bb541566dbf0b63a536b64e1b6028

generator/overrides/gen.go
davidfestal: Fixed in commit https://github.com/devfile/api/pull/133/commits/37c1e341bb7bb541566dbf0b63a536b64e1b6028

generator/overrides/gen.go
davidfestal: Fixed in commit https://github.com/devfile/api/pull/133/commits/37c1e341bb7bb541566dbf0b63a536b64e1b6028

generator/overrides/gen.go
davidfestal: Fixed in commit https://github.com/devfile/api/pull/133/commits/37c1e341bb7bb541566dbf0b63a536b64e1b6028

generator/schemas/gen.go
davidfestal: The comment is just wrong copy-paste. Let me fix it.

davidfestal: Fixed in commit https://github.com/devfile/api/pull/133/commits/5deb2bcccdeac386f390562308b7f6095847b199

davidfestal: Fixed in commit https://github.com/devfile/api/pull/133/commits/dc2f2e328fdcc52f55898f028ef900277015f387

generator/schemas/gen.go
davidfestal: Fixed in commit https://github.com/devfile/api/pull/133/commits/dc2f2e328fdcc52f55898f028ef900277015f387

davidfestal:
> @davidfestal I'm wondering (since we're getting deep into details on generation now) -- is it possible to use block/flow scalars for descriptions? E.g. [this](https://github.com/davidfestal/api/blob/a28f9f55586ff9fa17ef1e666be7121714e8a12c/crds/workspace.devfile.io_devworkspaces.yaml#L76-L84) could be rewritten (I think) to:
> 
> ```yaml
> description: |
>     Command that consists in applying a given component definition,
>     typically bound to a workspace event.
>     
>     For example, when an `apply` command is bound to a `preStart` event,
>     and references a `container` component, it will start the container as a
>     K8S initContainer in the workspace POD, unless the component has its
>     `dedicatedPod` field set to `true`.
>     
>     When no `apply` command exist for a given component,
>     it is assumed the component will be applied at workspace start
>     by default.
> ```

Unfortunately I don't know an easy way to do this, because the yaml marshsalling is not controller by the generators. I only change the content of related GO APIs.

generator/overrides/gen.go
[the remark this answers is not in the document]
davidfestal: Fixed in commit https://github.com/devfile/api/pull/133/commits/37c1e341bb7bb541566dbf0b63a536b64e1b6028

diffJsons.sh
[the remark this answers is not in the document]
amisevsk: It's up to you; it's more complicated and not foolproof so I don't think it's necessary to include in this PR (which is already quite large :wink: )

If we find it's a problem later we can include it then.

generator/crds/gen.go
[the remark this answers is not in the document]
amisevsk: :+1: I'll create an issue.

pkg/apis/workspaces/v1alpha2/doc.go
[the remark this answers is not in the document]
amisevsk: :+1: discussion can be elsewhere.

schemas/latest/dev-workspace-template-spec.json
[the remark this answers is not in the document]
amisevsk: Makes sense, I noticed the kebab-case conversion and didn't make the connection (as I got to that much later in the review)

generator/overrides/gen.go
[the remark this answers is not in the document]
amisevsk: Because struct fields represent internal state for the struct -- by updating it each iteration through `ctx.Roots`, we're just setting it to the last `ctx.Root`. It looks like it's being used as an instance variable. In other words, 
- If `g.packageTypes` is used elsewhere, we're setting it incorrectly since we use it to hold transient state
- If `g.packageTypes` is only used for this function, it could be just be `packageTypes := map[string]*markers.TypeInfo{}`

generator/overrides/gen.go
amisevsk: `s/commit/comment/`, of course -- woops :)

amisevsk:
> Unfortunately I don't know an easy way to do this, because the yaml marshsalling is not controller by the generators. I only change the content of related GO APIs.

Ah of course, that makes sense.

diffJsons.sh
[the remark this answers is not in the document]
davidfestal: OK, cool.

generator/overrides/gen.go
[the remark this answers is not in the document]
davidfestal: Yeah, I see what you mean.

amisevsk (commented):
Pretty much looks good to me, except for the generated schema changes (it seems to remove some types, e.g. customCommand).

schemas/latest/with-markdown-descriptions/dev-workspace-template-spec.json
```diff
@@ -1,19 +1,59 @@
 {
   "description": "Structure of the workspace. This is also the specification of a workspace template.",
+  "type": "object",
```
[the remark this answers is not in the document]
davidfestal: I checked the differences with the `diffJsons.sh`, and it turns out that the differences are expected are mostly related to one of the following reasons:
- Changes in descriptions in some fields are expected, due to automatic generation of overrides,
- Some fields are not required anymore, due to automatic generation of overrides with a consistent logic vs mandatory fields in overrides
- The complete removal of the ability to override a `custom` component, since it doesn't really make sense.

generator/overrides/gen.go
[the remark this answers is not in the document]
davidfestal: Fixed in commit https://github.com/devfile/api/pull/133/commits/d948ac3778ab00b7872dc858e8ca750775441d56 by passing `packageTypes` as argument to functions.

amisevsk (approved):
LGTM

~~One last issue: running `./build.sh` in the repo causes changes to go.mod (looks like generator dependencies are getting added there):~~ On second thought, this seems to be a VSCode issue. I'm OK to merge since it looks like the problem is on my end.

```diff
diff --git a/go.mod b/go.mod
index 62a6a43..2b6b103 100644
--- a/go.mod
+++ b/go.mod
@@ -3,15 +3,19 @@ module github.com/devfile/api
 go 1.13
 
 require (
+	github.com/elliotchance/orderedmap v1.3.0 // indirect
+	github.com/go-toolsmith/astcopy v1.0.0 // indirect
 	github.com/google/go-cmp v0.4.0 // indirect
 	github.com/hashicorp/go-multierror v1.1.0
+	github.com/iancoleman/strcase v0.1.2 // indirect
 	github.com/json-iterator/go v1.1.9 // indirect
 	github.com/mitchellh/reflectwalk v1.0.1
 	github.com/stretchr/testify v1.4.0
-	golang.org/x/net v0.0.0-20200226121028-0de0cce0169b // indirect
 	golang.org/x/sys v0.0.0-20200122134326-e047566fdf82 // indirect
+	gomodules.xyz/orderedmap v0.1.0 // indirect
 	k8s.io/api v0.18.2
 	k8s.io/apimachinery v0.18.2
 	sigs.k8s.io/controller-runtime v0.5.2
+	sigs.k8s.io/controller-tools v0.4.0 // indirect
 	sigs.k8s.io/yaml v1.2.0
 )
```

l0rd (approved):

sleshchenko:
not sure if it's related to this PR but che/add-link seems to be broken
https://github.com/devfile/api/pull/133/checks?check_run_id=1166741097

generator/main.go
```diff
@@ -0,0 +1,227 @@
...
+func main() {
+	helpLevel := 0
+	whichLevel := 0
+	showVersion := false
+
+	cmd := &cobra.Command{
+		Use:   "generate",
+		Short: "Generate Kubernetes API extension resources and code.",
+		Long:  "Generate Kubernetes API extension resources and code.",
+		Example: `	# Provide the right examples ...`,
```
sleshchenko, line 113: Any plan to provide the right example?

generator/schemas/gen.go
```diff
@@ -0,0 +1,308 @@
...
+}
+
+var (
+	jsonschemaVersionMarker  = markers.Must(markers.MakeDefinition("devfile:jsonschema:version", markers.DescribesPackage, ""))
+	jsonschemaGenerateMarker = markers.Must(markers.MakeDefinition("devfile:jsonschema:generate", markers.DescribesType, GenerateJSONSchema{}))
+)
+
+// +controllertools:marker:generateHelp
+
+// Generator generates Json schemas from the GO source code of the Kubernetes API
```
sleshchenko, line 51: So tiny comment: In go it's common practice to use upper case for abbreviation why not use it here?
```suggestion
// Generator generates JSON schemas from the GO source code of the Kubernetes API
```

generator/schemas/zz_generated.markerhelp.go
```diff
@@ -0,0 +1,42 @@
...
+
+// Code generated by helpgen. DO NOT EDIT.
+
+package schemas
+
+import (
+	"sigs.k8s.io/controller-tools/pkg/markers"
+)
+
+func (GenerateJSONSchema) Help() *markers.DefinitionHelp {
```
sleshchenko, line 13: I see that's generated `go run sigs.k8s.io/controller-tools/cmd/helpgen generate:headerFile=../header.go.txt,year=2020 paths=.` but it's not clear at least for me - when it's supposed to happen? Triggered manually after the main file is edited?
If it's not there yet - should we include it into build process to make sure it has actual state on each PR?

sleshchenko (commented):
just flushing in the meantime of review )

sleshchenko (commented):
My chrome gets crazy when try to display everything and I type even small comment.
It made me thinking that it would be cool to collapse autogenerated code by default and such feature already exists on Github :tada: - https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/customizing-how-changed-files-appear-on-github
please consider adding .gitattributes with generated files ignoring. Not sure if it's possible to ignore whole folder but not file by extension but I suppose yes since I've found

> A .gitattributes file uses the same syntax as a .gitignore file for specifying individual files or types of files.

davidfestal:
> not sure if it's related to this PR but che/add-link seems to be broken
> https://github.com/devfile/api/pull/133/checks?check_run_id=1166741097

I think it's related to a bug in the check, that doesn't support well the case of forks.

generator/genutils/utils.go
```diff
@@ -0,0 +1,185 @@
...
+		markers.SimpleHelp("Devfile", "indicates that a given Struct type is a K8S union, and its fields (apart from the discriminator) are mutually exclusive. K8S unions are described here: https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190325-unions.md#proposal"))
+	into.AddHelp(UnionDiscriminatorMarker,
+		markers.SimpleHelp("Devfile", "indicates that a given field of an union Struct type is the union discriminator. K8S unions are described here: https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190325-unions.md#proposal"))
+	return nil
+}
+
+// WriteFormattedSourceFile creates a Go source file in a given package, dumps to it the content provided by the `writeContents` function
+// and formats the result through go/fmt.
+// If formatting cannot be applied (due to some syntax error probably), it returns an error.
+func WriteFormattedSourceFile(filename string, ctx *genall.GenerationContext, root *loader.Package, writeContents func(*bytes.Buffer)) error {
```
sleshchenko, line 154: I wonder it it makes sense to split logic of utils in different files, like marker.go, json_schema.go, patch_strategy.go, file.go like the following https://github.com/devfile/api/commit/1878c78ead018a4ab8444a042237c67f9a35156e

devfile.yaml
```diff
@@ -1,13 +1,13 @@
 metadata:
-  name: kubernetes-api
+  name: devfile-api
 projects:
-  - name: kubernetes-api
+  - name: api
     source:
       location: 'https://github.com/devfile/api.git'
       type: git
     clonePath: src/github.com/devfile/api
```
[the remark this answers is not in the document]
davidfestal: I'd rather do it in a distinct PR if possible.

generator/interfaces/gen.go
```diff
@@ -0,0 +1,167 @@
...
+	"github.com/elliotchance/orderedmap"
+)
+
+//go:generate go run sigs.k8s.io/controller-tools/cmd/helpgen generate:headerFile=../header.go.txt,year=2020 paths=.
+
+var (
+	toplevelListMarker = markers.Must(markers.MakeDefinition("devfile:toplevellist", markers.DescribesField, struct{}{}))
+)
+
+type copyStructs struct {
```
sleshchenko, line 22: IDE tells it's used why then not remove it?

generator/interfaces/gen.go
```diff
@@ -0,0 +1,167 @@
...
+						}
+					}
+				}
+			}
+		}); err != nil {
+			root.AddError(err)
+			return nil
+		}
+
+		genutils.WriteFormattedSourceFile("keyed_definitions", ctx, root, func(buf *bytes.Buffer) {
```
sleshchenko, line 78: Unhandled error  is returned here and seems there is no usage of that func with handled error.

generator/main.go
davidfestal: I just missed it. Thanks for pointing this out.
This was fixed in commit https://github.com/devfile/api/pull/133/commits/66c0dd08da2c596e5c77995eefd59a9579975fc0

generator/schemas/zz_generated.markerhelp.go
davidfestal: I just missed it. Thanks for pointing this out.

This was fixed in commit https://github.com/devfile/api/pull/133/commits/66c0dd08da2c596e5c77995eefd59a9579975fc0

generator/schemas/gen.go
davidfestal: Fixed in commit https://github.com/devfile/api/pull/133/commits/38469ecd2f52343bef7a60071fc5bf4356baefac

generator/interfaces/gen.go
davidfestal: Good catch. It's an old remainder of the original code I used to bootstrap this file.

davidfestal: Fixed in commit https://github.com/devfile/api/pull/133/commits/54fc1dc3059653c84260c92fc82a09476742337c

davidfestal:
> .gitattributes

generator/genutils/utils.go
davidfestal: I can do it, if you think it would bring more clarity.

davidfestal: Done in commit https://github.com/devfile/api/pull/133/commits/7a6fc780cf9a34af3a95f1184aca4fd6b2d8fd43

generator/interfaces/gen.go
davidfestal: Fixed in commit https://github.com/devfile/api/pull/133/commits/7a6fc780cf9a34af3a95f1184aca4fd6b2d8fd43
I simply remove the returned value in the function, since it already registers errors inside the current root.

generator/schemas/gen.go
```diff
@@ -0,0 +1,308 @@
...
+
+			(&currentJSONSchema).Title = schemaGenerateMarker.Title
+
+			jsonSchema, err := json.MarshalIndent(&currentJSONSchema, "", "  ")
+			if err != nil {
+				return err
+			}
+
+			toAddMarkdownDescriptions := orderedmap.New()
+			json.Unmarshal(jsonSchema, toAddMarkdownDescriptions)
```
sleshchenko, line 211: unhandled error

sleshchenko (approved):
I didn't manage to review code-generation logic itself but approving it since I believe the PR is in a great state, high quality enough to move forward

Good job! :+1:

PR #134, opened 2020-09-23
Explicitly state PROJECTS_ROOT & PROJECT_SOURCE are reserved

### LINKED ISSUE #132: Should PROJECTS_ROOT and PROJECT_SOURCE be reserved environment variables?
There's a couple environment variables that the devfile spec defines:
- `$PROJECTS_ROOT`: A path where projects sources are mounted
- `$PROJECT_SOURCE`: A path to a project source, including the project root (`${PROJECTS_ROOT}/<project-name>`)

It's assumed that these environment variables will be set to the proper values and available in the containers defined in a devfile. 

In addition to the `sourceMapping` field, odo currently allows the projects root to be changed by overriding the `PROJECTS_ROOT` environment variable in the container spec, e.g.:
```
components:
- name: runtime
  container:
    image: quay.io/eclipse/che-java11-maven:nightly
    memoryLimit: 768Mi
    mountSources: true
    sourceMapping: /test
    command: ['tail']
    args: [ '-f', '/dev/null']
    env:
    - name: PROJECTS_ROOT
      value: /mycustom/path/
```

We had some discussion last week on `$PROJECT_SOURCE` and decided that it wasn't valid for odo to allow it to be overridden by users, as it relies on both the value of `$PROJECTS_ROOT` (see https://github.com/openshift/odo/issues/3781 for some discussion) and where the source code is located.

So, I guess I have a couple questions:
1) Should we update the devfile spec to formally not allow `$PROJECT_SOURCE` to be overridden in the container spec by users?
2) Should we do the same for `$PROJECTS_ROOT` too? Given that the project root can still be changed by configuring `sourceMapping`

maysunfaisal:

### What does this PR do?
This schema explicitly specifies that `PROJECTS_ROOT` & `PROJECT_SOURCE` env are reserved and cannot be overridden via component container's env

Fixes #132

### DIFF
--- a/crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
+++ b/crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
@@ -230,11 +230,8 @@ spec:
                             type: object
                           commandLine:
                             description: "The actual command-line string \n Special
-                              variables that can be used: \n  - `$PROJECTS_ROOT`:
-                              A path where projects sources are mounted \n  - `$PROJECT_SOURCE`:
-                              A path to a project source ($PROJECTS_ROOT/<project-name>).
-                              If there are multiple projects, this will point to the
-                              directory of the first one"
+                              variables that can be used: \n  - `$PROJECTS_ROOT` \n
+                              \ - `$PROJECT_SOURCE`"
                             type: string
                           component:
                             description: Describes component to which given action
@@ -285,11 +282,7 @@ spec:
                           workingDir:
                             description: "Working directory where the command should
                               be executed \n Special variables that can be used: \n
-                              \ - `$PROJECTS_ROOT`: A path where projects sources
-                              are mounted \n  - `$PROJECT_SOURCE`: A path to a project
-                              source ($PROJECTS_ROOT/<project-name>). If there are
-                              multiple projects, this will point to the directory
-                              of the first one"
+                              \ - `$PROJECTS_ROOT` \n  - `$PROJECT_SOURCE`"
                             type: string
                         required:
                         - commandLine
@@ -522,7 +515,12 @@ spec:
                           env:
                             description: "Environment variables used in this container
                               \n The following variables are reserved and cannot be
-                              overridden via env: \n  - `$PROJECTS_ROOT` \n  - `$PROJECT_SOURCE`"
+                              overridden via env: \n  - `$PROJECTS_ROOT`: A path where
+                              projects sources are mounted as defined by sourceMapping
+                              \n  - `$PROJECT_SOURCE`: A path to a project source
+                              ($PROJECTS_ROOT/<project-name>). If there are multiple
+                              projects, this will point to the directory of the first
+                              one"
                             items:
                               properties:
                                 name:
@@ -543,8 +541,8 @@ spec:
                           sourceMapping:
                             description: Optional specification of the path in the
                               container where project sources should be transferred/mounted
-                              when `mountSources` is `true`. When omitted, the value
-                              of the `PROJECTS_ROOT` environment variable is used.
+                              when `mountSources` is `true`. When omitted, the default
+                              value of /projects is used.
                             type: string
                           volumeMounts:
                             description: List of volumes mounts that should be mounted
@@ -925,11 +923,7 @@ spec:
                                     commandLine:
                                       description: "The actual command-line string
                                         \n Special variables that can be used: \n
-                                        \ - `$PROJECTS_ROOT`: A path where projects
-                                        sources are mounted \n  - `$PROJECT_SOURCE`:
-                                        A path to a project source ($PROJECTS_ROOT/<project-name>).
-                                        If there are multiple projects, this will
-                                        point to the directory of the first one"
+                                        \ - `$PROJECTS_ROOT` \n  - `$PROJECT_SOURCE`"
                                       type: string
                                     component:
                                       description: Describes component to which given
@@ -981,11 +975,8 @@ spec:
                                     workingDir:
                                       description: "Working directory where the command
                                         should be executed \n Special variables that
-                                        can be used: \n  - `$PROJECTS_ROOT`: A path
-                                        where projects sources are mounted \n  - `$PROJECT_SOURCE`:
-                                        A path to a project source ($PROJECTS_ROOT/<project-name>).
-                                        If there are multiple projects, this will
-                                        point to the directory of the first one"
+                                        can be used: \n  - `$PROJECTS_ROOT` \n  -
+                                        `$PROJECT_SOURCE`"
                                       type: string
                                   type: object
                                 id:
@@ -1224,7 +1215,12 @@ spec:
                                       description: "Environment variables used in
                                         this container \n The following variables
                                         are reserved and cannot be overridden via
-                                        env: \n  - `$PROJECTS_ROOT` \n  - `$PROJECT_SOURCE`"
+                                        env: \n  - `$PROJECTS_ROOT`: A path where
+                                        projects sources are mounted as defined by
+                                        sourceMapping \n  - `$PROJECT_SOURCE`: A path
+                                        to a project source ($PROJECTS_ROOT/<project-name>).
+                                        If there are multiple projects, this will
+                                        point to the directory of the first one"
                                       items:
                                         properties:
                                           name:
@@ -1245,8 +1241,8 @@ spec:
                                       description: Optional specification of the path
                                         in the container where project sources should
                                         be transferred/mounted when `mountSources`
-                                        is `true`. When omitted, the value of the
-                                        `PROJECTS_ROOT` environment variable is used.
+                                        is `true`. When omitted, the default value
+                                        of /projects is used.
                                       type: string
                                     volumeMounts:
                                       description: List of volumes mounts that should
@@ -1703,11 +1699,8 @@ spec:
                                 type: object
                               commandLine:
                                 description: "The actual command-line string \n Special
-                                  variables that can be used: \n  - `$PROJECTS_ROOT`:
-                                  A path where projects sources are mounted \n  -
-                                  `$PROJECT_SOURCE`: A path to a project source ($PROJECTS_ROOT/<project-name>).
-                                  If there are multiple projects, this will point
-                                  to the directory of the first one"
+                                  variables that can be used: \n  - `$PROJECTS_ROOT`
+                                  \n  - `$PROJECT_SOURCE`"
                                 type: string
                               component:
                                 description: Describes component to which given action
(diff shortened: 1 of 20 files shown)


elsony (changes requested):
The current approach is to repeat the same info on places that have access to those two variables.  Instead of repeating the same info all over the place, will it be cleaner to specify under the container env section and mentioned those two are reserved variable and cannot be overridden?  In reality, the user can only override those variables in the container's env setting so it is probably better to put that restriction right on that section.

GeekArthur:
I have generic comment, is it necessary to add the full definitions for `PROJECTS_ROOT` and `PROJECT_SOURCE` to all the places that may potentially use two env vars? It seems like a little bit redundant and it's not easy to maintain (we have to update all the places if it's necessary). Would it be better to define on one place then let all other places point to that place?

johnmcollier:
Yeah, I agree with @elsony, I think we just need a section in the container.env schema that outlines which variables are reserved and can't be set by the user.

maysunfaisal:
@elsony @GeekArthur @johnmcollier updated the PR! Makes sense.

GeekArthur:
@maysunfaisal It looks cleaner to me! I didn't find a PR review guide for this repo, probably we should let the code owner do the final review.

maysunfaisal:
@davidfestal @amisevsk @sleshchenko Can I pls get a review on this? Thx!

pkg/apis/workspaces/v1alpha2/commands.go
```diff
@@ -115,9 +115,9 @@ type ExecCommand struct {
 	//
 	// Special variables that can be used:
 	//
-	//  - `$PROJECTS_ROOT`: A path where projects sources are mounted
+	//  - `$PROJECTS_ROOT`
 	//
-	//  - `$PROJECT_SOURCE`: A path to a project source ($PROJECTS_ROOT/<project-name>). If there are multiple projects, this will point to the directory of the first one.
+	//  - `$PROJECT_SOURCE`
```
amisevsk, line 120: I'm not sure about removing the descriptions of the variables here is a good idea (at least, without some other, centralized location we can link to to explain it). Currently, the only way to figure out what `$PROJECTS_ROOT` and `$PROJECT_SOURCE` are to be used for is by checking the description for `env`, where the main purpose is to tell you that you can't override them. Basically, my thinking works out to:
- `ExecCommand` (the use-site): Describe what the variables can be used for
- `Container` (the define-site): Describe which variables can be overridden (or not)

An alternative if we want to keep the command descriptions terse is to e.g. link to a docs page (once a description of reserved env vars is added -- e.g. [here](https://devfile.github.io/devfile/assembly_making-a-workspace-portable-using-a-devfile.html#_adding_commands_to_a_devfile)).

Alternatively, a more generic doc would remove the "Special variables..." entirely; isn't it the case that you can use *any* environment variable in a command? The only difference for `$PROJECTS_ROOT` and `$PROJECT_SOURCE` is that they're supposed to automatically be set, but you should be able to reference e.g. `$M2_HOME`, etc. here as well.

```
// The actual command-line-string. Note environment variables from the container the command runs 
// in can be used in the command, e.g. 
//
// `cd ${PROJECT_SOURCE} && mvn clean install` 
```

I'm less certain this applies to the `workingDir`, where I think documenting the automatically-defined env vars is useful.

maysunfaisal: That makes sense. I've reverted the changes and just kept the section about reservation in env.

I've opened https://github.com/devfile/api/issues/138 to discuss and track on how we can define special variables in devfile spec context, as that seems to be out of this PR's scope.

sleshchenko (approved):

amisevsk (approved):

JPinkney (approved):

maysunfaisal:
I've rebased and updated the PR!

crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
```diff
@@ -285,11 +286,12 @@ spec:
...
-                              are mounted \n  - `${PROJECT_SOURCE}`: A path to a project
-                              source (${PROJECTS_ROOT}/<project-name>). If there are
-                              multiple projects, this will point to the directory
-                              of the first one."
+                              \ - `$PROJECTS_ROOT`: A path where projects sources
+                              are mounted as defined by container component's sourceMapping
+                              \n  - `$PROJECT_SOURCE`: A path to a project source
+                              ($PROJECTS_ROOT/<project-name>). If there are multiple
+                              projects, this will point to the directory of the first
+                              one"
```
elsony, line 294: Missing ending period

crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
```diff
@@ -231,10 +231,11 @@ spec:
                           commandLine:
                             description: "The actual command-line string \n Special
                               variables that can be used: \n  - `$PROJECTS_ROOT`:
-                              A path where projects sources are mounted \n  - `$PROJECT_SOURCE`:
+                              A path where projects sources are mounted as defined
+                              by container component's sourceMapping \n  - `$PROJECT_SOURCE`:
                               A path to a project source ($PROJECTS_ROOT/<project-name>).
                               If there are multiple projects, this will point to the
-                              directory of the first one."
+                              directory of the first one"
```
elsony, line 238: Missing ending period

crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
```diff
@@ -924,10 +928,11 @@ spec:
                                       description: "The actual command-line string
                                         \n Special variables that can be used: \n
                                         \ - `$PROJECTS_ROOT`: A path where projects
-                                        sources are mounted \n  - `$PROJECT_SOURCE`:
+                                        sources are mounted as defined by container
+                                        component's sourceMapping \n  - `$PROJECT_SOURCE`:
                                         A path to a project source ($PROJECTS_ROOT/<project-name>).
                                         If there are multiple projects, this will
-                                        point to the directory of the first one."
+                                        point to the directory of the first one"
```
elsony, line 935: Missing ending period

crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
```diff
@@ -979,11 +984,13 @@ spec:
...
-                                        A path to a project source (${PROJECTS_ROOT}/<project-name>).
-                                        If there are multiple projects, this will
-                                        point to the directory of the first one."
+                                        can be used: \n  - `$PROJECTS_ROOT`: A path
+                                        where projects sources are mounted as defined
+                                        by container component's sourceMapping \n
+                                        \ - `$PROJECT_SOURCE`: A path to a project
+                                        source ($PROJECTS_ROOT/<project-name>). If
+                                        there are multiple projects, this will point
+                                        to the directory of the first one"
```
elsony, line 993: Missing ending period

crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
```diff
@@ -1700,10 +1709,11 @@ spec:
...
                                 description: "The actual command-line string \n Special
                                   variables that can be used: \n  - `$PROJECTS_ROOT`:
-                                  A path where projects sources are mounted \n  -
-                                  `$PROJECT_SOURCE`: A path to a project source ($PROJECTS_ROOT/<project-name>).
+                                  A path where projects sources are mounted as defined
+                                  by container component's sourceMapping \n  - `$PROJECT_SOURCE`:
+                                  A path to a project source ($PROJECTS_ROOT/<project-name>).
                                   If there are multiple projects, this will point
-                                  to the directory of the first one."
+                                  to the directory of the first one"
```
elsony, line 1716: Missing ending period

crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
```diff
@@ -1755,11 +1765,12 @@ spec:
...
-                                  be used: \n  - `${PROJECTS_ROOT}`: A path where
-                                  projects sources are mounted \n  - `${PROJECT_SOURCE}`:
-                                  A path to a project source (${PROJECTS_ROOT}/<project-name>).
+                                  be used: \n  - `$PROJECTS_ROOT`: A path where projects
+                                  sources are mounted as defined by container component's
+                                  sourceMapping \n  - `$PROJECT_SOURCE`: A path to
+                                  a project source ($PROJECTS_ROOT/<project-name>).
                                   If there are multiple projects, this will point
-                                  to the directory of the first one."
+                                  to the directory of the first one"
```
elsony, line 1773: Missing ending period

crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
```diff
@@ -2434,12 +2448,13 @@ spec:
...
+                                            variables that can be used: \n  - `$PROJECTS_ROOT`:
                                             A path where projects sources are mounted
-                                            \n  - `${PROJECT_SOURCE}`: A path to a
-                                            project source (${PROJECTS_ROOT}/<project-name>).
+                                            as defined by container component's sourceMapping
+                                            \n  - `$PROJECT_SOURCE`: A path to a project
+                                            source ($PROJECTS_ROOT/<project-name>).
                                             If there are multiple projects, this will
-                                            point to the directory of the first one."
+                                            point to the directory of the first one"
```
elsony, line 2457: Missing ending period

crds/workspace.devfile.io_devworkspacetemplates.v1beta1.yaml
```diff
@@ -207,10 +207,11 @@ spec:
...
                           that can be used: \n  - `$PROJECTS_ROOT`: A path where projects
-                          sources are mounted \n  - `$PROJECT_SOURCE`: A path to a
-                          project source ($PROJECTS_ROOT/<project-name>). If there
-                          are multiple projects, this will point to the directory
-                          of the first one."
+                          sources are mounted as defined by container component's
+                          sourceMapping \n  - `$PROJECT_SOURCE`: A path to a project
+                          source ($PROJECTS_ROOT/<project-name>). If there are multiple
+                          projects, this will point to the directory of the first
+                          one"
```
elsony, line 214: Missing ending period. The same for the rest so just do a search and I am not marking the others.

crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
maysunfaisal: I've checked online and the general rule is; complete sentences should have periods and fragmented sentences should not have periods. I've updated the PR accordingly.

elsony (approved):
LGTM

davidfestal (approved):

Issue #136, opened 2020-09-25
Add license headers to files


amisevsk:
### Describe the bug
None of the Go files in this repo contain license headers. We should
1. Add license headers
2. Add a check to github actions to ensure license headers are set on all files for PRs

Jdubrick:
Closing issue as we have license headers implemented alongside a CI check for them.

PR #137, opened 2020-09-25
Add license headers to go files.

amisevsk:
### What does this PR do?
Adds a license header to all current files using [`addlicense`](https://github.com/google/addlicense). I re-used the license header we currently use for the [`devworkspace-operator`](https://github.com/devfile/devworkspace-operator/) since I'm not certain what the header should contain (likely we need to add other contributors?).

This PR also adds a license check to GitHub actions. 

https://github.com/devfile/api/issues/136

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -34,7 +34,7 @@ jobs:
     - name: Run GO tests
       run: go test -v ./...
 
-    - name: Check GO mod state      
+    - name: Check GO mod state
       run: |
         go mod tidy
         go mod vendor
@@ -45,7 +45,13 @@ jobs:
         go fmt -x ./...
         git diff --exit-code || { echo 'Go sources need to be formated. Execute "go fmt -x ./..." locally and commit changes to fix an issue'; exit 1; }
 
-    - name: Check Generator GO mod state      
+    - name: Check GO license headers
+      run: |
+        pushd $(mktemp -d); go get -u github.com/google/addlicense; popd
+        addlicense --check -v -f license_header.txt $(find . -name '*.go' -not -path '*/vendor/*') \
+          || { echo 'Go sources are missing license headers'. Use `addlicense --check -v -f license_header.txt $(find . -name '*.go' -not -path '*/vendor/*')` to ensure license headers are included; exit 1; }
+
+    - name: Check Generator GO mod state
       working-directory: generator
       run: |
         go mod tidy
--- a/generator/crds/gen.go
+++ b/generator/crds/gen.go
@@ -1,3 +1,15 @@
+//
+// Copyright (c) 2019-2020 Red Hat, Inc.
+// This program and the accompanying materials are made
+// available under the terms of the Eclipse Public License 2.0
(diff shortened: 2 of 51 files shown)


JPinkney (approved):

.github/workflows/ci.yaml
```diff
@@ -49,4 +49,10 @@ jobs:
...
       run: |
         go fmt -x ./...
-        git diff --exit-code || { echo 'Go sources need to be formated. Execute "go fmt -x ./..." locally and commit changes to fix an issue'; exit 1; }
+        git diff --exit-code || { echo 'Go sources need to be formatted. Execute "go fmt -x ./..." locally and commit changes to fix an issue'; exit 1; }
+
+    - name: Check GO license headers
+      run: |
+        pushd $(mktemp -d); go get -u github.com/google/addlicense; popd
+        addlicense --check -v -f license_header.txt $(find . -name '*.go' -not -path '*/vendor/*') \
+          || { echo 'Go sources are missing license headers'. Use addlicense to ensure license headers are included; exit 1; }
```
sleshchenko, line 58: `Use addlicense to ensure license headers are included` is not so specific to understand which command should be executed. What if we print the whole command here or even contribute Makefile with such command?

amisevsk: Update message, but I don't think this is the reason to add a makefile when there's already a `build.sh` in the repo.

sleshchenko: Updated message clearly explain what to do to solve the issue. So, resolving this comment.

Just an example where Makefile and scripts could live together https://github.com/openshift/odo/blob/master/Makefile )

sleshchenko (approved):

sleshchenko:
Seems GitHub action should be reworked somehow not to fail after License is dropped by code-generators https://github.com/devfile/api/pull/137/checks?check_run_id=1179505374

skabashnyuk:
I may suggest using a license header without years and change  `Red Hat, Inc. - initial API and implementation` to something company neutral like `Eclipse Che contributors. - initial API and implementation`. I think Eclipse IP team can assist with that.

amisevsk:
I'm fine making the license whatever the consensus is (though this PR is blocked by auto-generated code). 

Regarding

> I may suggest using a license header without years and change Red Hat, Inc. - initial API and implementation to something company neutral like Eclipse Che contributors. - initial API and implementation. I think Eclipse IP team can assist with that.

There's not a good option there; this isn't an Eclipse project AFAIK, and this repo moved from the che-incubator organization to devfile at one point in the past.

@davidfestal do you have any suggestions on what could be done to include license headers in autogenerated files? (or if there's a list of autogenerated files I can parse easily to exclude them from the check)

davidfestal:
> @davidfestal do you have any suggestions on what could be done to include license headers in autogenerated files? (or if there's a list of autogenerated files I can parse easily to exclude them from the check)

I assume that we should probably add an option to the generators, that would allow providing a file containing the license header, and then update the calls to the generation binary, to add this option when calling the generators.

amisevsk:
Closing PR as out-of-date/unnecessary.

Issue #75, opened 2020-06-08
mountsources default value for containers


gorkem:
The default value for for `mountsources` should be `true` for all component types except for the plugin type.

l0rd:
That's a good point because on the other end [we have discussed and agreed](https://github.com/devfile/kubernetes-api/issues/48) that the default value for `dedicatedPod` should be `true`. But `dedicatedPod: true` is compatible with `mountsources: true` only if the storage backend supports `RWX` OR if we use rsync to sync sources amongst pods.

gorkem:
I see the point. Perhaps we should just state that if not explicitly defined all component will mountsources except plugin types and those that have `dedicatedPod` set to be `true`.

johnmcollier:
Closed in https://github.com/devfile/api/pull/145

PR #145, opened 2020-10-02
Make mountSources true by default

johnmcollier:
### What does this PR do?
This PR updates the devfile schema to define the `mountSources` field as being true for all non-plugin components and components that don't set `dedicatedPod` to true.

https://github.com/devfile/api/issues/75

### DIFF
--- a/pkg/apis/workspaces/v1alpha2/containerComponent.go
+++ b/pkg/apis/workspaces/v1alpha2/containerComponent.go
@@ -42,6 +42,10 @@ type Container struct {
 	// +optional
 	Args []string `json:"args,omitempty" patchStrategy:"replace"`
 
+	// Toggles whether or not the project source code should
+	// be mounted in the component.
+	//
+	// Defaults to true for all component types except plugins and components that set `dedicatedPod` to true.
 	// +optional
 	MountSources bool `json:"mountSources,omitempty"`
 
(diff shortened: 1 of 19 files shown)


pkg/apis/workspaces/v1alpha2/containerComponent.go
```diff
@@ -42,6 +42,10 @@ type Container struct {
 	// +optional
 	Args []string `json:"args,omitempty" patchStrategy:"replace"`
 
+	// Toggles whether or not the project source code should
+	// be mounted in the component.
+	//
+	// Defaults to true for all component types except plugins and components that set `dedicatedPod` to true.
```
davidfestal, line 48: GO-wise, you should change the type to `*bool` instead of `bool` if you want to change the default value. But that's slightly less convenient to work with in the GO source code since you have to check whether the field is `nil` before taking its value.

davidfestal (changes requested):

pkg/apis/workspaces/v1alpha2/containerComponent.go
johnmcollier: Ah, that's a good point @davidfestal. I can update the type.

davidfestal (approved):
Seems OK to me, Che-wise. @l0rd is it OK for you as well ?

amisevsk (approved):

JPinkney (approved):

l0rd (approved):
LGTM

sleshchenko (approved):

sleshchenko (approved):

Issue #74, opened 2020-06-08
default values on JSON schema 


gorkem:
[devfile schema ](https://github.com/devfile/kubernetes-api/blob/master/schemas/devfile.json) should specify the default values for properties that have them.

davidfestal:
Initially I had specified default values as `kube-builder` annotations and they automatically went into the  CRD OpenAPI schema. 

But then there was an error, since it was a very recent feature and still not supported by the whole operator-sdk / OLM toolchain at this point, which is why I removed them.

But in the meantime we could quite easily accommodate this limitation in the script that generates CRDs and schemas from the Go source code.
I'll look into this after coming back from my PTO week (next week).

PR #148, opened 2020-10-05
Add `default` fields in Json schemas and `v1` CRDs  

davidfestal:
### What does this PR do?

This PR adds `default` fields in the Json schemas and the openAPI schemas of `v1` CRDs,
by adding related `kubebuilder:default` annotations in the GO source code.

It doesn't  add `default` fields to the openAPI schema of the `v1beta1` CRDs
since defaulting is not supported by the `v1beta1` apiextension API.

This should fix issue #74 

It is.

### DIFF
--- a/crds/workspace.devfile.io_devworkspaces.yaml
+++ b/crds/workspace.devfile.io_devworkspaces.yaml
@@ -474,6 +474,7 @@ spec:
                                       / \"ide\" / \"ide-dev\","
                                     type: object
                                   exposure:
+                                    default: public
                                     description: "Describes how the endpoint should
                                       be exposed on the network. \n - `public` means
                                       that the endpoint will be exposed on the public
@@ -498,6 +499,7 @@ spec:
                                     description: Path of the endpoint URL
                                     type: string
                                   protocol:
+                                    default: http
                                     description: "Describes the application and transport
                                       protocols of the traffic that will go through
                                       this endpoint. \n - `http`: Endpoint will have
@@ -556,6 +558,7 @@ spec:
                                 components that set `dedicatedPod` to true."
                               type: boolean
                             sourceMapping:
+                              default: /projects
                               description: Optional specification of the path in the
                                 container where project sources should be transferred/mounted
                                 when `mountSources` is `true`. When omitted, the default
@@ -632,6 +635,7 @@ spec:
                                       / \"ide\" / \"ide-dev\","
                                     type: object
                                   exposure:
+                                    default: public
                                     description: "Describes how the endpoint should
                                       be exposed on the network. \n - `public` means
                                       that the endpoint will be exposed on the public
@@ -656,6 +660,7 @@ spec:
                                     description: Path of the endpoint URL
                                     type: string
                                   protocol:
+                                    default: http
                                     description: "Describes the application and transport
                                       protocols of the traffic that will go through
                                       this endpoint. \n - `http`: Endpoint will have
@@ -730,6 +735,7 @@ spec:
                                       / \"ide\" / \"ide-dev\","
                                     type: object
                                   exposure:
+                                    default: public
                                     description: "Describes how the endpoint should
                                       be exposed on the network. \n - `public` means
                                       that the endpoint will be exposed on the public
@@ -754,6 +760,7 @@ spec:
                                     description: Path of the endpoint URL
                                     type: string
                                   protocol:
+                                    default: http
                                     description: "Describes the application and transport
                                       protocols of the traffic that will go through
                                       this endpoint. \n - `http`: Endpoint will have
@@ -1185,6 +1192,7 @@ spec:
                                                 / \"ide\" / \"ide-dev\","
                                               type: object
                                             exposure:
+                                              default: public
                                               description: "Describes how the endpoint
                                                 should be exposed on the network.
                                                 \n - `public` means that the endpoint
@@ -1212,6 +1220,7 @@ spec:
                                               description: Path of the endpoint URL
                                               type: string
                                             protocol:
+                                              default: http
                                               description: "Describes the application
                                                 and transport protocols of the traffic
                                                 that will go through this endpoint.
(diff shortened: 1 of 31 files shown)


amisevsk (approved):
LGTM.

JPinkney (approved):

l0rd (approved):

crds/workspace.devfile.io_devworkspaces.yaml
```diff
@@ -552,6 +554,7 @@ spec:
                             mountSources:
                               type: boolean
                             sourceMapping:
+                              default: /projects
```
sleshchenko, line 557: Do we really want here to have that default value? Then it will provide explicit value into devfile when user uses competition
Screenshot_20201006_103838

I wonder if it makes sense to omit by purpose default values for things we don't expect to configure very often but that's an assumption that may be totally wrong


P.S. I got default fields filled in only with local VSCode, Che devfile/api workspace seems to comple only container field name but not more, not sure what is the problem:
Screenshot_20201006_104445

sleshchenko (approved):

crds/workspace.devfile.io_devworkspaces.yaml
sleshchenko: P.S. I got default fields filled in only with local VSCode, Che devfile/api workspace seems to comple only container field name but not more, not sure what is the problem:
Screenshot_20201006_104445

Editor validation against schema does not seem to work as well
Screenshot_20201006_104732

davidfestal: Thanks @sleshchenko I'll have a look to those problems.

davidfestal: Oh, OK, I think I know the reason: it takes the default devfile 1.0 schema available in the schema store instead of the local one configured in the theia settings.json. I'll answer in the other issue.

davidfestal: > Do we really want here to have that default value? Then it will provide explicit value into devfile when user uses competition

Oh, I didn't realize that the yaml completion was automatically adding the defaults when doing object completion. That seems weird, but isn't it more a problem of the yaml language server ? 

> 
> I wonder if it makes sense to omit by purpose default values for things we don't expect to configure very often but that's an assumption that may be totally wrong

Well, I don't really know where to draw the line here, between "very often" and "not very often".

@l0rd @gorkem What do you think about that ? Do we still want to include default values in the Json schema itself instead of simply documenting them ?

sleshchenko: > That seems weird, but isn't it more a problem of the yaml language server ?

I'm not sure it's weird.
The default value is a value but not default behavior, they are a bit different things. I think such behavior works fine in a combination of the required field and defaultValue.
We might be interesting to propose users like 256Mi for memory limit to easier start from scratch and show the example for format.
Exposure type is another thing that we may want for user to be clear about.

But I agree about 
> between "very often" and "not very often".
that may be difficult to formulate rule when a field should be promoted with a default value and when not

davidfestal: > > That seems weird, but isn't it more a problem of the yaml language server ?
> 
> I'm not sure it's weird.
> The default value is a value but not default behavior, they are a bit different things. I think such behavior works fine in a combination of the required field and defaultValue.

I agree that if a field is mandatory and has a default value, it should be filled with the default value. But what if it is optional ? Why force the user to add it ?

JPinkney: IMO after reviewing hundreds of different JSON schemas that the YAML Language Server uses the default value should be auto inserted. It is the default value that will be processed if a user doesn't change it. According to the JSON schema spec `JSON processing tools may use this information to provide a default value for a missing key/value pair`. Also, it is the convention used by the JSON Language Server as well. If you didn't want this item to autocomplete automatically it should use the `examples` tag instead

davidfestal: @JPinkney I understand your point about the expected YAML language server behavior. However I'm not sure that, for the devfile, we have a strong agreement that we should force all the fields with a default value to be added automatically by completion *in all cases*.

So for now I just removed the `default` directives from the Json schemas that are IDE-targeted. But of course default values are still described in the docs.

generator/schemas/gen.go
```diff
@@ -207,21 +207,35 @@ func (g Generator) Generate(ctx *genall.GenerationContext) error {
...
+			genutils.EditJSONSchema(
+				&currentJSONSchema,
+				func(schema *apiext.JSONSchemaProps) (newVisitor genutils.Visitor, stop bool) {
+					if schema != nil {
+						schema.Default = nil
+					}
+					return
+				})
+
+			(&currentJSONSchema).Title = (&currentJSONSchema).Title + " - IDE-targeted variant"
```
sleshchenko, line 219: Do you list somewhere what is the difference between normal and IDE-targeted variant?
Markdown, missing default, ...

sleshchenko (approved):

generator/schemas/gen.go
davidfestal: Not now. Would simply adding a `Readme.md` file in the `ide-targeted` folder be sufficient ?
And possibly explain the difference in the main description inside the Json schema ?

crds/workspace.devfile.io_devworkspaces.yaml
davidfestal: Yes, but only with the `v1` apiextensions API: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#defaulting

generator/schemas/gen.go
davidfestal: > Does K8s/OS provision these defaults to CR with missing items?

Yes, but only with the `apiextensions/v1` CRD API: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#defaulting

davidfestal: > Not now. Would simply adding a `Readme.md` file in the `ide-targeted` folder be sufficient ?
> And possibly explain the difference in the main description inside the Json schema ?

Done in commit https://github.com/devfile/api/pull/148/commits/6e6ac081df479eff476f4ee2f1de5a9585031a0f

elsony (approved):
LGTM

Issue #152, opened 2020-10-05
Exec command should make component required


maysunfaisal:
Currently the spec does not define the component in an exec as required. I am wondering if that's the case, what is the point in have an exec command at all? If the there is an exec command in a devfile, it is to be executed in a container; which is to be represented by the component property.

https://github.com/devfile/api/blob/master/pkg/apis/workspaces/v1alpha2/commands.go#L125

davidfestal:
@l0rd wdyt about this one ? It seems we should make the component mandatory right ?

elsony:
I don't see much usage on a command without a component. Without the component info, the tools using the devfile won't know where are they going to execute the command.

l0rd:
Agreed. The component should be mandatory for exec commands. On the other hand apply commands should not require that.

mmulholla:
So this:

	// Describes component to which given action relates
	// +optional
	Component string `json:"component,omitempty"`

Becomes:

	// Describes component to which given action relates
	// 
	Component string `json:"component"`

?????

davidfestal:
Yes, exactly.

johnmcollier:
So, I think this is a good idea. But one thing I'm wondering is, will this cause issues for overriding parent commands in devfiles if the component isn't specified?

E.g.
```
schemaVersion: 2.0.0
metadata:
  name: nodejs
  version: 1.0.0
parent:
  uri: https://raw.githubusercontent.com/odo-devfiles/registry/master/devfiles/nodejs/devfile.yaml
  commands:
    - id: install
      exec:
        commandLine: npm install && echo hi
```

Is currently valid, and changes the `commandLine` for the `install` command. But, will it still be valid if we make `component` required?

mmulholla:
For example, https://github.com/devfile/api/blob/master/samples/devfiles/simple-devfile.yaml#L30 includes:

  - id: helloWorld
    exec:
      env:
        - name: "USER"
          value: "John Doe"
      commandLine: 'echo "Hello ${USER}"'

This example would now be invalid. Is this acceptable?

maysunfaisal:
@mmulholla yes, that would become invalid because the exec command has no component to execute itself.

mmulholla:
So given this and what John pointed out do we still want this change?

davidfestal:
> For example, https://github.com/devfile/api/blob/master/samples/devfiles/simple-devfile.yaml#L30 includes:
> 
> * id: helloWorld
>   exec:
>   env:
[...]

PR #161, opened 2020-10-08
Comp mandatory4 exec command

mmulholla:
### What does this PR do?
Make component required for exec and apply commands

https://github.com/devfile/api/issues/152

Ran schema_validate.sh

### DIFF
--- a/pkg/utils/overriding/test-fixtures/patches/add-command-and-component/patch.yaml
+++ b/pkg/utils/overriding/test-fixtures/patches/add-command-and-component/patch.yaml
@@ -4,7 +4,6 @@ commands:
     id: "commandWithTypeChanged"
   - exec:
       commandLine: "commandToReplaceCommandLineReplaced"
-      component: "mycomponent"
       env:
         - name:  "commandToReplaceEnvVarToReplace"
           value: "commandToReplaceEnvVarToReplaceNewValue"
@@ -13,7 +12,6 @@ commands:
     id: "commandToReplace"
   - exec:
       commandLine: "commandToAddCommandLine"
-      component: "mycomponent"
     id: "commandToAdd"
 components:
   - volume: {}
--- a/pkg/utils/overriding/test-fixtures/patches/add-command-and-component/patch.yaml
+++ b/pkg/utils/overriding/test-fixtures/patches/add-command-and-component/patch.yaml
@@ -4,6 +4,7 @@ commands:
     id: "commandWithTypeChanged"
   - exec:
       commandLine: "commandToReplaceCommandLineReplaced"
+      component: "mycomponent"
       env:
         - name:  "commandToReplaceEnvVarToReplace"
           value: "commandToReplaceEnvVarToReplaceNewValue"
@@ -12,6 +13,7 @@ commands:
     id: "commandToReplace"
   - exec:
       commandLine: "commandToAddCommandLine"
+      component: "mycomponent"
     id: "commandToAdd"
 components:
   - volume: {}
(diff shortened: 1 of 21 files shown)


pkg/utils/overriding/test-fixtures/patches/add-command-and-component/patch.yaml
```diff
@@ -4,6 +4,7 @@ commands:
     id: "commandWithTypeChanged"
   - exec:
       commandLine: "commandToReplaceCommandLineReplaced"
+      component: "mycomponent"
```
davidfestal, line 7: You should not have to add the component in the `patch.yaml`, precisely because it's expected to be an override, and overrides have a dedicated schema that makes most fields optional.

But of course the `original.yaml` and `result.yaml` will need it.

pkg/utils/overriding/test-fixtures/patches/command-and-env/patch.yaml
```diff
@@ -4,6 +4,7 @@ commands:
     id: "commandWithTypeChanged"
   - exec:
       commandLine: "commandToReplaceCommandLineReplaced"
+      component: "mycomponent"
```
davidfestal, line 7: You should not have to add the component in the `patch.yaml`, precisely because it's expected to be an override, and overrides have a dedicated schema that makes most fields optional.

But of course the `original.yaml` and `result.yaml` will need it.

davidfestal (changes requested):
I made one comment about the `patch.yaml` files.

In addition to that, don't we want to also make the `component` mandatory in the `Apply` command types ?

davidfestal (approved):

pkg/utils/overriding/test-fixtures/patches/command-and-env/patch.yaml
mmulholla: I can remove the components from the patch.yaml files before merging if you would like me to.

One side question, after installing the Red Hat Yaml checker to VSCode it shows an error if I leave the component out in these files. I added this to my settings:

    "yaml.validate": true,
    "yaml.schemas": {
        "/Users/mmulholl@us.ibm.com/devfiles/api/schemas/latest/devfile.json": ["/*.yaml"],
    }

Is there more I should add?

davidfestal: If you open the VSCode workspace from this workspace definition: https://github.com/devfile/api/blob/master/devfile.api.code-workspace, then you should already have the Yaml schemas proprely configured:

```json
        "yaml.schemas": {
            "./schemas/latest/ide-targeted/devfile.json": [ "samples/devfiles/*devfile.yaml", "samples/devfile-registry/*.yaml" ],
            "./schemas/latest/ide-targeted/dev-workspace.json": [ "samples/devworkspaces/*.yaml" ],
            "./schemas/latest/ide-targeted/dev-workspace-template.json": [ "samples/devworkspace-templates/*.yaml" ],
            "./schemas/latest/ide-targeted/dev-workspace-template-spec.json": [ "pkg/utils/overriding/test-fixtures/patches/**/original.yaml", "pkg/utils/overriding/test-fixtures/patches/**/result.yaml", "pkg/utils/overriding/test-fixtures/merges/*.yaml" ],
            "./schemas/latest/ide-targeted/parent-overrides.json": [ "pkg/utils/overriding/test-fixtures/**/patch.yaml" ]
        }
```

The difference with what you have in your settings, is that the patch yaml files are  validated with the overrides schema, and not with the main one.

pkg/utils/overriding/test-fixtures/patches/add-command-and-component/patch.yaml
mmulholla: Removed the component

pkg/utils/overriding/test-fixtures/patches/command-and-env/patch.yaml
mmulholla: Thank you. Have removed the component

PR #166, opened 2020-10-09
Extract the `DevfileHeader` from the K8S API.

davidfestal:
# What does this PR do?

This PR extract the Devfile header source code definition, from the K8S API package itself.
By devfile header, I mean the top-level fields that are not really part of the Devfile body main structure,
but are specific to the Devfile YAML syntax, and will *not* be used by the the K8S custom resources such a `DevWorkspace`, `DevWorkspaceTemplate`, etc ...
 
These fields are:
- `schemaVersion`
- `metadata` with the optional `name` and `version` 

They are not coupled with a given K8S API version package.

This PR doesn't fix any issue, but will make it easier to simply unmarshal any devfile 2+ go get header fields from raw Json or Yaml. This will in particular allow getting the schema version in order to proceed with a complete json schema validation after retrieving the json schema of the schema version.  

Yes.


pkg/devfile/devfile.go
```diff
@@ -0,0 +1,25 @@
+package devfile
```
amisevsk, line 1: Are we sure it makes sense to include packages from outside `api/...` here? This removes the devfile header from versioning. This makes sense since `devfile` isn't a CRD, so maybe the entire `devfile.go` file should be moved to `pkg/devfile` and import `DevWorkspaceTemplateSpec`.

davidfestal: > This makes sense since devfile isn't a CRD, so maybe the entire `devfile.go` file should be moved to `pkg/devfile` and import `DevWorkspaceTemplateSpec`

Well, since the `DevWorkspaceTemplateSpec` should in any case be part of the K8S API, importing it from outside the K8S API package, doing this would make the `common` package depend on a given K8S API version, which is precisely what we want to avoid.

The new `DevfileHeader` structure (which contains only `schemaVersion` and `DevfileMetadata`) is the only thing in the whole model that is not in any way coupled with (or used in) the versioned K8S API. On the contrary the main Devfile body structure (`DevWorkspaceTemplateSpec`) is also part of the K8S API.

mmulholla (commented):
LGTM

pkg/devfile/devfile.go
sleshchenko, line 1: maybe it's a bit better to rename file to header.go

pkg/devfile/devfile.go
```diff
@@ -0,0 +1,25 @@
...
+
+// DevfileHeader describes the structure of the devfile-specific top-level fields
+// that are not part of the K8S API structures
+type DevfileHeader struct {
+	// Devfile schema version
+	// +kubebuilder:validation:Pattern=^([2-9])\.([0-9]+)\.([0-9]+)(\-[0-9a-z-]+(\.[0-9a-z-]+)*)?(\+[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$
+	SchemaVersion string `json:"schemaVersion"`
```
sleshchenko, line 8: > will make it easier to simply unmarshal any devfile 2+ go get header fields from raw Json or Yaml.

Does it make sense to declare unmarshaling somewhere here? Or devfile/parser is a better place for such stuff and one day it will appear there?

sleshchenko (approved):

validate-samples.sh
```diff
@@ -46,7 +46,7 @@ rm -f validate-output.txt &> /dev/null
 
 for schema in "devfile" "dev-workspace" "dev-workspace-template"
 do
-  schemaPath="./schemas/latest/with-markdown-descriptions/${schema}.json"
+  schemaPath="./schemas/latest/ide-targeted/${schema}.json"
```
sleshchenko, line 49: Can we do anything to prevent the inconsistency we have currently in master? Seems validate samples github action does not fail when it's supposed to 
Screenshot_20201013_124237

amisevsk (approved):

JPinkney (approved):

pkg/devfile/devfile.go
davidfestal: Fixed

validate-samples.sh
davidfestal: Fixed in commit https://github.com/devfile/api/pull/166/commits/35a4a0177a8fd17b8c78851b73a113a2029cd369

Issue #188, opened 2020-10-21
Re-write the OCI registry metadata container in Go


johnmcollier:
The current metadata container for the OCI registry uses shell scripts to generate the index.json, push the devfiles to the registry, and start the Apache HTTP server. This approach gives us little flexibility in pushing up the devfile stacks, as we're not able to easily parse the index.json without pulling down additional dependencies like jq. It also means that we need to package the index-generator and ORAS CLIs (the latter of which is intended only for testing and not use in production) into the container

It would be wise to re-write the metadata container (and server) in Go to make use of the index-generator and oras Go libraries (rather than CLIs), as well as to host the index.json. This will give us additional flexibility in the future, as we look to define APIs on top of the metadata container's server.

GeekArthur:
The metadata container written by golang should follow the changes in this degisn proposal: https://github.com/devfile/api/pull/197

Issue #164, opened 2020-10-08
Build tools for generating registry based on registry/catalog info on a github repo


elsony:
## User Story
As a devfile admin, I want to generate a devfile registry based on the content in a registry catalog repo that defines all the available stacks to be added to the registry.

## Acceptance Criteria 
- [ ] It should read the content of a given registry catalog repo to gather all the info on the list of stacks to be include in the registry
- [ ] It should build a devfile registry packaged in the form as described in #163

## Links 
- Related issue: 
https://github.com/devfile/api/issues/163
This item is part of the epic #82

johnmcollier:
I've put a PR up (https://github.com/devfile/registry/pull/1) for review that implements the following:
1) Build scripts and docker file for producing a Devfile index docker container, as outlined in https://github.com/devfile/api/issues/163
2) GitHub workflow actions for running the build on PRs and publishing new images to Quay.io

PR #197, opened 2020-10-27
Add design proposal for devfile packaging

### LINKED ISSUE #163: Packaging of devfile registry for deployment
## User Story
As a devfile admin, I want to create a package of the devfile registry that can be hosted for other devfile users to access.

## Acceptance Criteria 
- [ ] It should define how the packaging of the devfile registry looks like
- [ ] It should define the mechanism to deploy in different environment, e.g. cluster vs standalone registry

## Links 
- Related issue: 
This item is part of the epic #82

johnmcollier:
### What does this PR do?
(For some context for those not on the devfile services team: we discussed internally last week on where to store design proposals related to devfiles, and decided upon creating a `docs/proposals` folder in this repo, similar to what [odo does](https://github.com/openshift/odo/tree/master/docs).)

This PR adds a design proposal outlining how devfiles (and their artifacts) that are stored in a GitHub repository should be packaged up for distribution in an OCI devfile registry. 

CC @elsony @GeekArthur 

Fixes #163 

N/A

### DIFF
--- a/docs/proposals/registry/devfile-packaging.md
+++ b/docs/proposals/registry/devfile-packaging.md
@@ -24,22 +24,25 @@ Storing the devfiles (and any associated artifacts) on a dedicated Github reposi
 But, with the move to OCI registries, we need a way to easily package the devfile stacks from GitHub and load them into the OCI registry
 
    - Needs to retrieve the devfiles without an internet connection
-     - Means we can’t have the metadata container git clone the devfiles before pushing them up
+     - Means we can’t have the bootstrap container git clone the devfiles before pushing them up
    - Needs to be easily built and distributable by registry admins.  
      - Once we support “deploying your own registry”, with custom devfiles this will become especially important
-   - Shouldn’t require building and deploying custom components of the registry to distribute (i.e. it shouldn’t be baked into the registry or metadata container images)
-     - Currently, the metadata container has the devfiles built into it, meaning users need to build their own metadata container image to customize devfiles
+   - Shouldn’t require building multiple components of the registry to distribute
 
 ## Proposed Changes
 
-**To address the requirements listed above, rather than bundling the devfiles as part of registry's bootstrap container, build the “devfile registry repository” into its own docker image**
+**To address the requirements listed above, have the “devfile registry repository” built into a custom bootstrap image, based on a base-image provided by us.**
 
-<img width="636" alt="Screen Shot 2020-10-26 at 3 39 33 PM" src="https://user-images.githubusercontent.com/6880023/97220180-767cdc80-17a1-11eb-85fd-cf1a5a623aeb.png">
+<img width="633" alt="Screen Shot 2020-10-29 at 12 25 53 PM" src="https://user-images.githubusercontent.com/6880023/97602641-eaa5c300-19e1-11eb-8c3b-1c3ee0cb6f11.png">
 
 
-This container image will contain just the devfiles (and their artifacts), as well as any other required files.
+The base-image will be the "bootstrap container" image that's currently in the [devfile/registry-support](https://github.com/devfile/registry-support) repo and will be hosted on the `devfile` org on quay.io. It contains the logic for loading the devfiles into the OCI registry as well as hosting the index.json.
 
-When deploying the OCI registry, the image will be specified as an init container, with an entrypoint command that copies the devfiles (and artifacts) to the bootstrap container’s volume. Then, as-is today, when the bootstrap container starts up, it generates the index.json and pushes the devfiles up into the OCI registry using `oras`.
+The Dockerfile will just contain lines to copy the devfiles and index.json into the container, nothing else will be required (unless the registry administrator chooses to add more).
+
+As part of the registry build in the registry repository (such as [devfile/registry](https://github.com/devfile/registry)), the index.json will be generated and any tests/validation specific to the repository will be run. After this, the Dockerfile will be built and an image will be produced and pushed up to a Docker registry (if required).
+
+When deploying the OCI registry, the registry's image will be specified as the bootstrap container. Then, when the bootstrap container starts, it pushes the devfiles up into the OCI registry using `oras` and serves the pre-generated index.json.
 
 An example of what the devfile registry repository may look like can be found at https://github.com/johnmcollier/registry/tree/registryDockerfile
 
@@ -48,4 +51,3 @@ An example of what the devfile registry repository may look like can be found at
 - Easy and straightforward method of way and distributing devfiles and their artifacts. 
     - If a user wants to deploy a registry with their own devfiles, this will be the only thing they have to change
 - Everything needed to preload the OCI registry is contained in this image
-- No need to build custom components of the registry (such as the bootstrap container)
--- a/docs/proposals/registry/devfile-packaging.md
+++ b/docs/proposals/registry/devfile-packaging.md
@@ -1,0 +1,51 @@
+# Devfile Packaging for OCI Registries
+
+This document will define how devfiles and their artifacts that are hosted on Github can be packaged up and deployed onto an OCI Devfile registry running on a Kubernetes cluster.
+
+I recommend having a read over the [OCI Registry Lifecycle design](https://docs.google.com/document/d/1rQHCp4SWslWWJv5KK3A_iXHvgbqjsuDkSKDD72ifJio/edit) first, before reading this doc. 
+
+
+## As-is Today
+Currently, devfiles are stored on Github in a “devfile registry repository” (such as [devfile/registry](https://github.com/devfile/registry), [odo-devfiles/registry](https://github.com/odo-devfiles/registry), and [eclipse/che-devfile-registry](https://github.com/eclipse/che-devfile-registry)). Generally, each registry repository will have a devfiles folder:
+
+<img width="742" alt="Screen Shot 2020-10-22 at 1 54 00 PM" src="https://user-images.githubusercontent.com/6880023/97219676-b2637200-17a0-11eb-8465-1063aa048768.png">
+
+
+Each folder under devfiles corresponds to a stack, with its own devfile and associated artifacts. There may be an index.json present for devfile consumers
+
+From this point, the Github repository may act as self-hosting devfile registry (as in the case of [odo-devfiles/registry](https://github.com/odo-devfiles/registry), or it may be built up into an Apache server container (as in the case of v1 registries like [eclipse/che-devfile-registry](https://github.com/eclipse/che-devfile-registry)).
+
+## Are any changes needed to accommodate OCI?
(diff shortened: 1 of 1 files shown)


docs/proposals/registry/devfile-packaging.md
```diff
@@ -0,0 +1,51 @@
...
+## Proposed Changes
+
+**To address the requirements listed above, rather than bundling the devfiles as part of registry's bootstrap container, build the “devfile registry repository” into its own docker image**
+
+<img width="636" alt="Screen Shot 2020-10-26 at 3 39 33 PM" src="https://user-images.githubusercontent.com/6880023/97220180-767cdc80-17a1-11eb-85fd-cf1a5a623aeb.png">
+
+
+This container image will contain just the devfiles (and their artifacts), as well as any other required files.
+
+When deploying the OCI registry, the image will be specified as an init container, with an entrypoint command that copies the devfiles (and artifacts) to the bootstrap container’s volume. Then, as-is today, when the bootstrap container starts up, it generates the index.json and pushes the devfiles up into the OCI registry using `oras`.
```
amisevsk, line 42: I don't know if this is underspecified or sufficient for now, but do we need to describe some sort of contract between the init container and the registry itself? It might help to have clarity in how the init command "glue" is going to be defined, e.g. with "env var `X` defines where devfiles are stored in the init container` or `env var `X` defines the path to devfiles in the registry`. 

Can I plop multiple init containers onto the registry and have them merge into one? This would be useful for e.g. adding a few of my own devfiles while keeping the default as-is. Having a defined description of where plugins should be and how they will be copied means potentially less work for init container authors.

amisevsk (approved):
This makes sense to me; ideally folks wanting to add a devfile to the registry can simply write a container `FROM <our-image>` and copy their devfiles in.

docs/proposals/registry/devfile-packaging.md
johnmcollier: That's a good question. I could probably add the following to the proposal to make things more clear:

> The following environment variables will be set in the init container image to define where the devfiles and plugins are to be copied from:
`DEVFILES_SOURCE=/devfiles`
`PLUGINS_SOURCE=/plugins`

and 

> The follow entrypoint command will be used to copy the devfiles and plugin over to the registry's volume:
`cp -rf $DEVFILES_SOURCE /registry && cp -rf $PLUGINS_SOURCE /registry`


Re multiple init containers: I hadn't thought of that, but I see no reason why we _couldn't_ allow it, and agreed that it could be useful as an alternative to modifying the default image. Maybe have something like the following in the registry custom resource?

```
devfileSource:
    - "quay.io/devfile/devfiles:latest"
    - "docker.io/johncollier/somedevfiles:latest"
    - "docker.io/someorg/moredevfiles:latest"
```

And when the registry is deployed, each of those entries map to an init container, copying the contents into the registry's volume?

elsony, line 42: I'll refer the actual build of the devfile registry as "registry build" from now on.  This proposal defers the registry build including the generation of the index.json to the time of the deployment phase of the registry instead of the during the build phase in the registry life cycle. The problem that I see with this approach is that it is too late in the development/deployment cycle to try to do the registry build during registry deployment. The registry build includes the creation of the index.json and should be part of the on boarding process of the registry.  As part of the registry build, we'll need do any the validation of the stack and some of those validation may even potentially running any of the stack validation test. This process may take long so we should do it as part of the registry build phase instead of doing it in the deployment phase. 

Taking a step back, I still do not see the benefit of the proposed devfile only package instead of the current bootstrap image approach. At the end of day, it is the registry admin that needs to deploy the registry. There is always a build step to build from the registry repo source to the metadata container or the proposed repo content only image.  As long as we provide enough support for registry admin to easily build that final image, e.g. we provide a parent image including the build tools for that knows how to build the index.json and do the validation, the Dockerfile file for the bootstrap image should should remain fairly simple. Using this approach, the benefits are:
1. we can push all the heavy lifting job on the bootstrap image build which is done as part of the build phase of the registry image to shorten the time for deployment. 
2. Any validation problem with the content of the stacks will be found out during the build phase instead of delaying to find out in the deployment phase. and thus, reducing the risk of deployment.

elsony (changes requested):

docs/proposals/registry/devfile-packaging.md
johnmcollier: > I'll refer the actual build of the devfile registry as "registry build" from now on. This proposal defers the registry build including the generation of the index.json to the time of the deployment phase of the registry instead of the during the build phase in the registry life cycle. The problem that I see with this approach is that it is too late in the development/deployment cycle to try to do the registry build during registry deployment. The registry build includes the creation of the index.json and should be part of the on boarding process of the registry. As part of the registry build, we'll need do any the validation of the stack and some of those validation may even potentially running any of the stack validation test. This process may take long so we should do it as part of the registry build phase instead of doing it in the deployment phase.

So, that's a good point. I think my design proposal was a little ambiguous in this regard. I wasn't planning on offloading the registry build to the deployment, but I didn't realize the index.json generation was part of the "registry build". We can certainly generate the index.json as part of the registry build (as well as doing any necessary validation / testing), and then include that index.json in the "devfile image". 

> Taking a step back, I still do not see the benefit of the proposed devfile only package instead of the current bootstrap image approach. At the end of day, it is the registry admin that needs to deploy the registry. There is always a build step to build from the registry repo source to the metadata container or the proposed repo content only image. As long as we provide enough support for registry admin to easily build that final image, e.g. we provide a parent image including the build tools for that knows how to build the index.json and do the validation, the Dockerfile file for the bootstrap image should should remain fairly simple. Using this approach, the benefits are:

So, I still prefer keeping the devfiles / stacks separate from the bootstrap image, as it means that registry admins don't have to maintain (and keep up-to-date) separate forks of the bootstrap image just to bundle different devfiles in a registry. Especially for compatibility with new versions of the registry operator, it means that admins don't have to rebuild / rebase their custom metadata image just to ensure it works with the new operator.

What if we adapt the current proposal so that it's clear the "registry build" does the index.json generation, and any validation / testing as you mentioned, but it also builds up this "devfile image" for consumption (with the generated index.json)? Then I think all 3 of your points should be addressed, and we're able to keep the devfiles separate from the bootstrap container.

johnmcollier: Discussed on today's meeting:
- Can't avoid the registry admin having to update their "devfile image" between registry updates anyways, in case things such as the index.json or devfile packaging format change between releases
- My approach would require a second persistent volume
- Elson's approach is more sound and not any less complicated for the registry admin.

johnmcollier: So Elson and I discussed things today and we decided to not go forward with the initContainer model, but rather package the devfiles into the bootstrap image in the registry

johnmcollier: @elsony Updated the design doc based on today's meeting, please re-review when you get the chance, thanks!

elsony (approved):
LGTM

docs/proposals/registry/devfile-packaging.md
```diff
@@ -0,0 +1,53 @@
...
+
+<img width="633" alt="Screen Shot 2020-10-29 at 12 25 53 PM" src="https://user-images.githubusercontent.com/6880023/97602641-eaa5c300-19e1-11eb-8c3b-1c3ee0cb6f11.png">
+
+
+The base-image will be the "bootstrap container" image that's currently in the [devfile/registry-support](https://github.com/devfile/registry-support) repo and will be hosted on the `devfile` org on quay.io. It contains the logic for loading the devfiles into the OCI registry as well as hosting the index.json.
+
+The Dockerfile will just contain lines to copy the devfiles and index.json into the container, nothing else will be required (unless the registry administrator chooses to add more).
+
+As part of the registry build in the registry repository (such as [devfile/registry](https://github.com/devfile/registry)), the index.json will be generated and any tests/validation specific to the repository will be run. After this, the Dockerfile will be built and an image will be produced and pushed up to a Docker registry.
+
```
GeekArthur, line 44: Just would like clarify couple of things, sounds like we need to update our existing "bootstrap container" to accommodate this design proposal, more specifically the changes would be:
1. Remove the code of pulling devfiles from "devfile registry repository" (GitHub devfile repository).
2. Remove the code of generating registry index, and seems like we decide to generate the initial registry index from "devfile registry repository", I think the change is only remove the initial index generating part but still leave the index generator in the "bootstrap container" as we may need to support the devfile update use case in the future.

docs/proposals/registry/devfile-packaging.md
```diff
@@ -0,0 +1,53 @@
...
+When deploying the OCI registry, the registry's image will be specified as the bootstrap container. Then, when the bootstrap container starts, it pushes the devfiles up into the OCI registry using `oras` and serves the pre-generated index.json.
```
GeekArthur, line 45: > After this, the Dockerfile will be built and an image will be produced and pushed up to a Docker registry.

Another question is about the line above, the `Dockerfile` refers to which Dockerfile? Is it the Dockerfile is the one admin builds and it is based on the "bootstrap image"? I would be better we have a terminology section in this design proposal to define the terminology (eg. "bootstrap container", "devfile registry repository", "registry Dockerfile", "registry support repository", etc) and we can use that in the following content.

docs/proposals/registry/devfile-packaging.md
GeekArthur: The reason to clarify this is because once we get the change list for "bootstrap container" then we can update that requirement in this issue https://github.com/devfile/api/issues/188 to ensure the new "bootstrap container" written in golang works properly.

johnmcollier: Yup, we need to update the bootstrap container to not pull the devfiles or create the index.json. Basically converting https://github.com/devfile/registry-support/tree/master/oci-devfile-registry-metadata to a base-image.

I think we can handle that as part of https://github.com/devfile/api/issues/164 since it's specifically referring to the build tools for the registry and implementing what's described here.

docs/proposals/registry/devfile-packaging.md
johnmcollier: Yup, it's the registry's Dockerfile. Sure, I can add a terminology section.

docs/proposals/registry/devfile-packaging.md
GeekArthur: Sounds good 👍

docs/proposals/registry/devfile-packaging.md
GeekArthur: Thanks John!

johnmcollier: @GeekArthur Updated to add a terminology section 👍

GeekArthur (approved):
LGTM, thanks for making the changes!

PR #199, opened 2020-10-29
Add new tests for v200 schema

mmulholla:


mmulholla:
Interestingly the new tests are run automatically as part of the pull request build.

test/devfiles/v200/commands/ExecMissingEnvOrder.yaml
```diff
@@ -0,0 +1,16 @@
...
+      commandLine: 'echo "Hello ${GREETING} ${USER}"'
+      component: api-tests
+      group: 
+        isDefault: true
+        kind: test
+      hotReloadCapable: false
+      label: "Command Exec test"
+      env:
+        - value : "USER"
+          name : USER
```
amisevsk, line 15: I'm not sure what this test is checking.

test/go/src/tests/api-v200_test.go
```diff
@@ -0,0 +1,220 @@
...
+	"github.com/ghodss/yaml"
+	"github.com/santhosh-tekuri/jsonschema"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+// Users struct which contains
```
amisevsk, line 15: What do we mean by "users struct"?

amisevsk (commented):
Please separate out testing code and readme changes from `yaml` files used for test cases; it's quite hard to review this PR as-is (it's a flat list of 141 files basically)

I'm really not clear on what we're actually trying to test here -- are we trying to validate that the schema we're generating matches what's expected? Are we actually just testing that the json generator does what it says on the box? If we have doubts about the correctness of the schema generator, we should be testing the schema generator.

The structure of the testing code is also very strange to me -- it seems vastly overcomplicated for what would be a fairly simple task. Have we considered using a [fuzzer](https://github.com/google/gofuzz) to generate randomly filled structs, serialize them to json, and check them against the schema?

test/devfiles/v200/commands/ExecMissingEnvOrder.yaml
mmulholla: May be a redendant one, just checking it does not matter which order you put the name and value in.

test/go/src/tests/api-v200_test.go
mmulholla: oops, cut and paste error, will fix it.

test/devfiles/v200/commands/ExecMissingEnvOrder.yaml
mmulholla: I fixed the issues you raised and also one I noticed. The easiest way to review is to go to the branch and then the test subdirectory which the PR creates: https://github.com/mmulholla/api/tree/AddAPITests/test

There you can see the README and easily browse the new files.

The tests may seem a bit pointless right now, but will be useful in the future when the schema is updated and we check if v200 devfiles still work with the new schema.

mmulholla:
I noticed an error in the test logic which was giving some false positives. Going to cancel this pull request until I have addressed the issues

mmulholla:
Fixed the issues and added a commit to the pull request so have re-opened it.

mmulholla:
Will close the pull request again and re-open later once the endpoint proticol issue has concluded and test modified accordingly.

test/devfiles/v200/commands/ExecMissingEnvOrder.yaml
mmulholla: Also not sure how these tests are over complicated. They perform two goals :
 
1. test every attriibute in the schema, make sure mandatory attributes are indeed mandatory and optional ones are indeed optional. Then where attributes have a restriction make sure those restrictions are enforced. 
2. Make it easy to add new tests for a new attribute or different schema.

Add to this that the validator only reports the first problem, so you need a test for every error case. There is obviously a bunch of different ways you can achieve these goals but, would be equally complicated.

Adding a fuzzer to generate values would be interesting but would need all the logic in the go code which then makes it more difficult to add and changes tests. Also not sure how it works when an attribute is restricted to a range of values.

Maybe i misunderstand how the fuzzer would be used.

test/devfiles/v200/commands/ExecBadDefault.yaml
```diff
@@ -0,0 +1,18 @@
+  - id: testExec1
```
sleshchenko, line 1: filename is too generic. At the first look, I thought it's about default value for some field, but it's about isDefault field. So, maybe ExecBadDefaultInGroup would be a bit more self-descriptive?

test/devfiles/v200/commands/ExecBadDefault.yaml
```diff
@@ -0,0 +1,18 @@
...
+        isDefault: maybe
+        kind: test
+      hotReloadCapable: false
+      label: "Command Exec test"
+      env:
+        - name: "USER"
+          value: "Test Tester"
+        - name : "GREETING"
+          value : "Hello" 
+      workingDir: $PROJECTS_ROOT
```
sleshchenko, line 18: Could you please add newlines in files?

test/devfiles/v200/commands/ExecBadDefault.yaml
mmulholla: OK, will rename them.

test/devfiles/v200/commands/ExecBadDefault.yaml
mmulholla: Sorry, not sure what you mean.

sleshchenko: I mean this 
Screenshot_20201111_171812

here you can find why it's better to add it https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline/729725

mmulholla: Thank you  all done:
- added newlines
- updated tests for recent changes - attributes updates and name and id restrictions
- added tests for. the new name and id restrictions 

Read for review again, hopefully all good this time :-)

mmulholla: Some tests failing on the build. failing to find files - need to dig into it, all was good locally on my mac.

test/go/src/tests/api-v200_test.go
mmulholla: all good now, differenence in how Mac and linux handle file names.

test/devfiles/v200/commands/ExecBadDefault.yaml
mmulholla: All good now, difference in how Mac and linux handle case in file names.

test/go/src/tests/api-v200_test.go
```diff
@@ -0,0 +1,258 @@
...
+							t.Errorf("  FAIL : %s : devfile was valid - Expected Error not found :  %s", testsToRunContent.Tests[i].FileName, testsToRunContent.Tests[i].ExpectOutcome)
+						} else {
+							passTests++
+							t.Logf("PASS : %s : %s", testsToRunContent.Tests[i].FileName, testJsonContent.SchemaFile)
+						}
+						f.Close()
+					}
+				}
+			}
+			t.Logf("%s : %d of %d tests passed", testJsonFile.Name(), passTests, totalTests)
```
sleshchenko, line 214: It's also useful to see number of skipped tests

test/go/src/tests/api-v200_test.go
```diff
@@ -0,0 +1,258 @@
...
+}
+
+// Users struct which contains
+// an array of users
+type Command struct {
+	id            string `yaml:"id"`
+	SchemaVersion string `yam:"SchemaVersion"`
+}
+
+func Test_WriteYaml(t *testing.T) {
```
sleshchenko, line 237: What does this test do?

sleshchenko (commented):
The test code seems to suppress a lot of errors and is not really readable.
I propose to move forward and improve it a bit like the following example https://github.com/mmulholla/api/compare/AddAPITests...sleshchenko:AddAPITestsPatch?expand=1
^ it's not finished yet but I hope shows the idea

mmulholla:
@sleshchenko - please let me know when you consider your branch complete

test/README.md
```diff
@@ -0,0 +1,41 @@
+# API Tests
+
+The API tests are intended to provide a comprehensive verification of the devfile schemas. This includes:
+- Ensuring every possible attribute is valid.
+- Ensuring all optional attributes are indeed optional.
+- Ensuring any possible specification errors are invalidated by the schema. For example:
+    - Missing mandatory attributes.
+    - Mutiple use of a one-of attribute.
+    - atttribute values of the wrong type.
```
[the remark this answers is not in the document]
amisevsk: Remove the extra 't' while we're at it :)

amisevsk (commented):
I won't block merging or anything, but in general, I'm not a fan of tests in this style -- we're basically testing that a few annotations are set correctly in `pkg/apis`. However, since the tests are written manually, there's a much larger chance that there is an error in the tests than the code itself. 

I don't see a case where I could make an error writing code that is caught by these tests -- they basically will fail whenever a mandatory/optional/enum field is changed, but the solution there is to just update the tests to accommodate the change. It feels like it'll end up being rote maintenance for tests for the sake of having them. 

Also, these tests will have to be kept in sync with documentation, which I suppose is the ultimate source of truth. This seems like it will be a large maintenance burden for little benefit.

elsony:
@amisevsk The main goal of these tests is to make sure that we don't change the API and breaking the existing devfiles. In case there is an error occurs, it is either some changes have been introduced that breaks existing devfiles. It is true that the document should match but there is no way for us to confirm that the doc matches the code to enforce compatibility. These tests are meant for catching that. For intentional changes on the schema, I'd expect the test change will go with the code changes to reflect those.

test/README.md
```diff
@@ -0,0 +1,41 @@
+# API Tests
+
+The API tests are intended to provide a comprehensive verification of the devfile schemas. This includes:
+- Ensuring every possible attribute is valid.
+- Ensuring all optional attributes are indeed optional.
+- Ensuring any possible specification errors are invalidated by the schema. For example:
+    - Missing mandatory attributes.
+    - Mutiple use of a one-of attribute.
```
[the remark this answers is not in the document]
mmulholla: fixed

test/README.md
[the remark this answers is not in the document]
mmulholla: fixed

test/README.md
```diff
@@ -0,0 +1,41 @@
...
+1. run the test
+
+### add a test for a schema changes
+
+1. Modify an an existing yaml snippet or create a new one.
+2. If approprate create a new snippet for any possible error cases, for example to omit a required attribute.
+3. If a new yaml snippet was created add a test which uses the snippet to the appropriate `json/xxxxxx-tests.json` file. Be careful to ensure the FileName set for the test is unique for all tests - this is the name used for the yaml file which is generated for the test. For failure scenarios you may need to run the test first to set the outcome correctly. 
+4. If a new  `json/xxxxxx-tests.json` file is created, any existing `test-xxxxxxx.json` files must be updated to use the new file.
+
+
```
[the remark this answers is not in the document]
mmulholla: Fixed.

test/README.md
```diff
@@ -0,0 +1,41 @@
...
+1. create a new `test/go/src/tests/json/v200/test-<schema name>.json` file for the schema. In the json file  specify the location of the schema to test (relative to the root directory of the repository), and the list of the existing tests to use. If the generated yaml files require a schemaVersion attribute include its value in the json file. see - *link to sample schema to be added*
+1. run the test
+
+### add a test for a schema changes
+
+1. Modify an an existing yaml snippet or create a new one.
+2. If approprate create a new snippet for any possible error cases, for example to omit a required attribute.
+3. If a new yaml snippet was created add a test which uses the snippet to the appropriate `json/xxxxxx-tests.json` file. Be careful to ensure the FileName set for the test is unique for all tests - this is the name used for the yaml file which is generated for the test. For failure scenarios you may need to run the test first to set the outcome correctly. 
+4. If a new  `json/xxxxxx-tests.json` file is created, any existing `test-xxxxxxx.json` files must be updated to use the new file.
+
```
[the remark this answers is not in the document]
mmulholla: Fixed

test/README.md
```diff
@@ -0,0 +1,41 @@
...
+from the test/go/src/test directory run 
+- `go test -v`
+
+The test will read each of the test-xxxxxx.json files and run the tests defined within. The generated .yaml files used for the tests are created in a `tmp/v200/test-xxxxxx/` directory. These files are not deleted when the test finishes so they can be used to assess any errors, however they will be deleted by a subsequent run of the test. Running the test with the -v option ensures you see a full list of passes and failures. 
+
+## Adding Tests
+
+### add a test for a new schema file
+
+1. create a new `test/go/src/tests/json/v200/test-<schema name>.json` file for the schema. In the json file  specify the location of the schema to test (relative to the root directory of the repository), and the list of the existing tests to use. If the generated yaml files require a schemaVersion attribute include its value in the json file. see - *link to sample schema to be added*
```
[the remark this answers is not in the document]
mmulholla: Typos fixed. I am planning to add a link once this is merged so that it will link into the devfile/api repo.

test/README.md
```diff
@@ -0,0 +1,41 @@
...
+1. run the test
+
+### add a test for a schema changes
+
+1. Modify an an existing yaml snippet or create a new one.
+2. If approprate create a new snippet for any possible error cases, for example to omit a required attribute.
```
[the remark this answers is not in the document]
mmulholla: Fixed.

test/go/src/tests/api-v200_test.go
```diff
@@ -0,0 +1,258 @@
...
+				t.Errorf("FAIL : failed to read : %s : %v", testJsonFile.Name(), err)
+			}
+
+			var testJsonContent TestJson
+
+			// Unmarshall the contents of the json file which defines the tests to run for each test
+			err = json.Unmarshal(byteValue, &testJsonContent)
+			if err != nil {
+				t.Fatalf("FAIL : failed to unmarshal : %s : %v", testJsonFile.Name(), err)
+				continue
```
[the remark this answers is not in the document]
mmulholla: Good spot, updated Fatalf to Errorf.

test/go/src/tests/api-v200_test.go
mmulholla: Removed - just a harmless old function I used to try stuff out.

mmulholla:
Further to Angels comments about complexity I simplified the directory structure and renamed the test to make thing more obvious. Also I set it up so when we want to create a new set of tests for a new version of the schema it is easy to re-use this test directory structure as a starting point - I added instruction for this in the README.

test/go/src/tests/api-v200_test.go
mmulholla: Sure. will also address this is a subsequent pull request.

test/go/src/tests/api-v200_test.go
```diff
@@ -0,0 +1,258 @@
...
+	Files         []string `json:"Files"`
+}
+
+const testDir = "../../../"
+const schemaDir = "../../../../"
+const jsonDir = "./json/v200/"
+const tempRootDir = "./tmp/"
+const tempDir = "./tmp/v200/"
+
+func Test_API_200(t *testing.T) {
```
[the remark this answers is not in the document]
mmulholla: Agree it is a bit cumbersome. I saw your branch and plan to handle this in a later pull request. Not decided yet but thinking I may create, for example,  a new json-utils.go file to provide an api for access to the json files. 

Opened this issue: https://github.com/devfile/api/issues/298

elsony (approved):
/lgtm (with some comments going to be covered by #298)

PR #209, opened 2020-11-06
Add devworkspaces shortname: dw

l0rd:
### What does this PR do?

Add a shortname for the `DevWorkspace` Custom Resource so that we can:

```bash 
$ kubectl get dw 
$ kubectl get devworkspaces 
```

Make it easier to manage devworkspaces with kubectl.

```
$ kubectl apply -f crds/workspace.devfile.io_devworkspaces.v1beta1.yaml 
customresourcedefinition.apiextensions.k8s.io/devworkspaces.workspace.devfile.io created 
$ kubectl api-resources 
(...) 
devworkspaces                     dw          workspace.devfile.io           true         DevWorkspace
```

### DIFF
--- a/crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
+++ b/crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
@@ -26,6 +26,7 @@ spec:
     plural: devworkspaces
     shortNames:
     - dev
+    - dw
     singular: devworkspace
   preserveUnknownFields: false
   scope: Namespaced
--- a/crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
+++ b/crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
@@ -25,7 +25,6 @@ spec:
     listKind: DevWorkspaceList
     plural: devworkspaces
     shortNames:
-    - dev
     - dw
     singular: devworkspace
   preserveUnknownFields: false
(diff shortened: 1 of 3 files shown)


amisevsk (commented):
I'm not sure about the `dev` shortname -- it feels like it could be likely to conflict with something else eventually. However I don't feel strongly enough one way or another to block this PR. 

I'd also toss `dw` as an option here (maybe in addition to `dev`) -- this is similar to e.g. `ReplicaSet = rs`, `DeploymentConfig = dc`, etc.

l0rd:
That's exactly what I was thinking: both dev and dw. I haven't found a way
to add both using an annotation though.

>

johnmcollier:
@l0rd 

You should be able to do the following to add multiple shortnames:
```
 // +kubebuilder:resource:path=devworkspaces,scope=Namespaced,shortName=dev;dw
```

l0rd:
Thanks @johnmcollier, that worked. I have added `dw` too.

Shortname `dw` is accurate, `dev` is cool. I will let this open a few days to see what others prefer.

sleshchenko (approved):
+1 for dw
+0 for dev. Agree that it's easy to remember but I see it too general. It could be alias for devworkspace, devfile, anything starting with dev. Alternative cool and easy to remember (but a bit longer) could be devws. But I don't have a strong opinion.

elsony:
> +1 for dw
> +0 for dev. Agree that it's easy to remember but I see it too general. It could be alias for devworkspace, devfile, anything starting with dev. Alternative cool and easy to remember (but a bit longer) could be devws. But I don't have a strong opinion.

I agree. dw and devws are better.  dev is too general.

amisevsk (approved):
LGTM; while we're at it, do we want to add something like `dwt` for DevWorkspaceTemplates?

l0rd:
There is consensus on `dw`. I have updated the PR so that it is the only shortname.

Issue #190, opened 2020-10-22
Deployment of devfile registry in cluster


elsony:
In order for a devfile registry to make available for tools to consume or use, the devfile registry needs to run inside a cluster. This item is to investigate to see how we can deploy the registry to a given cluster.

This item is part of Epic #82 

- [x] Design doc for how the registry should be deployed on clusters
- [x] Consumption of the devfile packaging as described in https://github.com/devfile/api/issues/163
- [ ] Implementation of operator based on the design doc
   - [x] Potential proxy server for safe guarding the devfile registry

Link:
- This is the implementation for the design in https://github.com/devfile/api/issues/162

johnmcollier:
Moving to in progress...

I've updated this issue's description to include the required tasks for this item.

johnmcollier:
I've put up the design proposal for the Devfile Registry operator and custom resource in https://github.com/devfile/api/pull/210

I've also created a proof of concept devfile registry operator in https://github.com/johnmcollier/registry-operator

johnmcollier:
Design proposal was merged last week. It can now be found here: https://github.com/devfile/api/blob/master/docs/proposals/registry/registry-operator.md

I've put up a PR in the new `registry-operator` repo to merge in the Devfile Registry operator functionality: https://github.com/devfile/registry-operator/pull/2

johnmcollier:
Proxy server PR in https://github.com/devfile/registry-support/pull/11

PR #210, opened 2020-11-06
Create design proposal for a Devfile Registry operator

johnmcollier:
Signed-off-by: John Collier <jcollier@redhat.com>

### What does this PR do?
This PR adds a design proposal for a "Devfile Registry Operator", to be used to deploy and manage devfile registries on a Kubernetes or OpenShift cluster

https://github.com/devfile/api/issues/190

N/A

#### Docs PR
N/A


johnmcollier:
CC @elsony @GeekArthur

docs/proposals/registry/registry-operator.md
```diff
@@ -0,0 +1,97 @@
...
+
+#### Registry 
+`devfileIndexImage` (string) Image used to bootstrap the OCI registry with devfiles and host the index.json.
+
+`ociRegistryImage` (string) Image used for the OCI registry instance. Defaults to `registry:2.7.1`
+
+#### Storage
+`enabled` (bool) Enables persistent storage for the registry. Defaults to true.
+
+`registryVolumeSize` (string). Size of the volume to use. Defaults to 3Gi.
```
amisevsk, line 47: This seems like a very large default for what amounts to a few thousand lines worth of text files an their intermediate representations.

docs/proposals/registry/registry-operator.md
```diff
@@ -0,0 +1,97 @@
...
+    ingressDomain: ''
+```
+
+## Operator Functionality
+
+### `DevfileRegistry` Creation
+When an instance of a `DevfileRegistry` resource is created in the namespace where the Devfile Registry operator is deployed, the following occurs in the operator's `reconcile` loop:
+
+1) A service exposing the devfile index server and OCI registry is created
+2) If persistence was enabled, The service that exposes the registry is created
```
amisevsk, line 82: Should this make reference to a PVC instead?

docs/proposals/registry/registry-operator.md
```diff
@@ -0,0 +1,97 @@
...
+3) The deployment containing the OCI registry and bootstrap container is deployed
+4) The OpenShift route or Kubernetes ingress resource is created
+    a) (OpenShift only) If TLS was enabled, configure the OpenShift route to re-use the cluster's certificate
+    b) (Kubernetes only) If TLS was enabled and `tls.secretName` was set, configure the ingress resource to use it
```
amisevsk, line 86: It's worth looking into cert-manager support here -- I believe it's somewhat of a k8s standard.

amisevsk (commented):
Generally makes sense to me. A few suggestions below.

docs/proposals/registry/registry-operator.md
johnmcollier: D'oh. I did some reformatting of the checklist and messed things up haha. I'll go fix that.

docs/proposals/registry/registry-operator.md
johnmcollier: So, that's a good point and we probably won't need 3Gi of storage for the OCI registry.

1Gi seems like a pretty common minimum and should still give us breathing room if any of the devfile stacks require additional blobs to be stored (e.g. vsix plugins).

docs/proposals/registry/registry-operator.md
johnmcollier: So, I had thought about integrating with cert-manager, but since it's an optional install I decided against it.

What about something like this on K8s?:
- If `tls.enabled` is true, and no tls secret is set in `tls.secretName`, then cert-manager must be installed and we use it to create a certificate.
- If `tls.enabled` is true, and a tls secret is set in `tls.secretName`, then we use it instead of cert-manager

amisevsk: That makes sense and is somewhat in-line with our approach in the devworkspace controller/che. It's good to support both so that cert-manager isn't a *hard* requirement but I would lean towards cert manager being the default with a workaround of manually creating your cert.

docs/proposals/registry/registry-operator.md
```diff
@@ -0,0 +1,97 @@
...
+    ingressDomain: ''
+```
+
+## Operator Functionality
+
+### `DevfileRegistry` Creation
+When an instance of a `DevfileRegistry` resource is created in the namespace where the Devfile Registry operator is deployed, the following occurs in the operator's `reconcile` loop:
+
+1) A service exposing the devfile index server and OCI registry is created
+2) If persistence was enabled, a persistent volume claim (PVC) is provisioned.
```
elsony, line 82: Do we foresee the persistence to be disabled in our regular use case and are we planning to support that? Disabling the persistence means we need to initialize the entire registry in case the registry goes down for some reasons. Given that we foresee the registry may contain more than just devfiles, e.g. other resources that is associated with a stack, we may want to run with persistence always enabled.

docs/proposals/registry/registry-operator.md
```diff
@@ -0,0 +1,97 @@
...
+3) The deployment containing the OCI registry and bootstrap container is deployed
+4) The OpenShift route or Kubernetes ingress resource is created
+    a) (OpenShift only) If TLS was enabled, configure the OpenShift route to re-use the cluster's certificate
```
elsony, line 85: Fix the formatting

docs/proposals/registry/registry-operator.md
elsony: The other option is use cert-manager when it is already installed

elsony (changes requested):

docs/proposals/registry/registry-operator.md
johnmcollier: Yup, agreed. Sounds good.

docs/proposals/registry/registry-operator.md
johnmcollier: So, it certainly won't be a regular use case in production (for the reasons that you mentioned), but I included it for use in development/testing, especially in environments where storage or even dynamic provisioning may not be set up or available. In that case, we'll use an emptyDir volume to provide the storage, which will last for the entire lifecycle of the pod (including restarts).

I'm open to removing it however since it's just for dev/test, what are your thoughts?

elsony: I am fine with keeping it for dev/testing. We may want to clarify it in the proposal on that.

docs/proposals/registry/registry-operator.md
```diff
@@ -0,0 +1,101 @@
...
+`ingressDomain` (string) Ingress domain to be used when deploying on Kubernetes.
+
+### DevfileRegistry Custom Resource Formatting
+```
+apiVersion: registry.devfile.io/v1alpha1
+kind: DevfileRegistry
+metadata:
+  name: registry
+spec:
+  bootstrapImage: ''
```
GeekArthur, line 64: According to the design above, I think `bootstrapImage` should be updated to `devfileIndexImage`

docs/proposals/registry/registry-operator.md
```diff
@@ -0,0 +1,101 @@
...
+
+### Resource Kind
+`DevfileRegistry`
+
+### Scope
+`Namespaced`
+
+### Fields
+
+#### Registry
```
GeekArthur, line 39: I assume the `Registry` here is an additional field that includes the two images? IIRC that is one of our old fields design and currently we move the two images one level up without `Registry` field

GeekArthur (changes requested):

docs/proposals/registry/registry-operator.md
johnmcollier: `Registry` here doesn't have any binding on the format of the custom resource spec, it's just a header to group fields together.

docs/proposals/registry/registry-operator.md
johnmcollier: D'oh good catch. Thought I got all references, but must've missed some.

docs/proposals/registry/registry-operator.md
johnmcollier: But actually, since the `registry` header is kind of ambiguous and confusing, I'll remove it and move the two fields up one level to match the spec.

GeekArthur: Okay, from the header level perspective, I saw `Registry`, `Storage` and `TLS` are under header level `####` and the header level `Fields` is `###` so I interpret it to the format below:
```yaml
spec:
  registry:
    bootstrapImage: ''
    ociRegistryImage: ''
  tls:
    enabled: true
    secretName: ''
  storage:
    enabled: true
    registryVolumeSize: '3Gi'
```

But that's fine if we just use it for group the definitions

GeekArthur (approved):
LGTM, thanks for making the changes!

elsony (approved):
LGTM

amisevsk (approved):
LGTM

JPinkney (approved):
LGTM

docs/proposals/registry/registry-operator.md
```diff
@@ -0,0 +1,100 @@
...
+
+`ociRegistryImage` (string) Image used for the OCI registry instance. Defaults to `registry:2.7.1`
+
+#### TLS
+`enabled` (bool) TLS mode for the devfile registry. Defaults to true. If on OpenShift, the cluster's routing certificate will be re-used. If on Kubernetes, cert-manager will be used to provision a TLS secret.
+
+`tlsSecretName` (string) Optional pre-existing Kubernetes secret for TLS termination on the ingress resource. If set, cert-manager will **not** be used to provision a TLS secret.
+
+#### Storage
+`enabled` (bool) Enables persistent storage for the registry. Defaults to true. Can be disabled for development or testing only.
```
sleshchenko, line 49: Sorry for being late for party )

From https://docs.google.com/document/d/1rQHCp4SWslWWJv5KK3A_iXHvgbqjsuDkSKDD72ifJio/edit#heading=h.cexfsm1d9e6n
> Update
The registry admin updates the devfile registry resource to pull down any new/updated stacks
The registry operator stands up a new registry deployment, that uses the updated container images with the updated devfile stacks
Once the new registry deployment is ready, the operator updates the registry ingress resource to point to the new registry, and tears down the old registry.

Does it mean that we require ReadWriteMany access mode?

docs/proposals/registry/registry-operator.md
```diff
@@ -0,0 +1,100 @@
...
+### Fields
+
+`devfileIndexImage` (string) Image used to bootstrap the OCI registry with devfiles and host the index.json.
+
+`ociRegistryImage` (string) Image used for the OCI registry instance. Defaults to `registry:2.7.1`
+
+#### TLS
+`enabled` (bool) TLS mode for the devfile registry. Defaults to true. If on OpenShift, the cluster's routing certificate will be re-used. If on Kubernetes, cert-manager will be used to provision a TLS secret.
+
+`tlsSecretName` (string) Optional pre-existing Kubernetes secret for TLS termination on the ingress resource. If set, cert-manager will **not** be used to provision a TLS secret.
```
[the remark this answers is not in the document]
johnmcollier: Huh, good to know. I'll put this down when we go look at TLS

docs/proposals/registry/registry-operator.md
johnmcollier: Nope, it'll be a new volume as well that's brought up as well, so still only ReadWriteOnce that's used.

PR #211, opened 2020-11-09
Fix events description

sleshchenko:
### What does this PR do?
Some time ago commands had names but then it was changed to ID, so this PR fixes events description which still references command's name which does not exist.

N/A

Haven't tested since it's just description changes.

#### Docs PR
N/A


amisevsk (approved):
Makes sense to me

davidfestal (approved):
Good catch ! Thanks @sleshchenko  !

Issue #169, opened 2020-10-13
Add `attributes` free-form map field for tool-specific fields in a number of places 


davidfestal:
**Is your feature request related to a problem? Please describe.**

How to support tool-specific fields in a number of places in Devfile 2.0 ?

**Describe the solution you'd like**

During the Devfile 2.0 early discussions, we had agreed that there should be an `attribute` free-form string map at a number of places in the top-level elements (commands, components, projects), to allow tools to provide tool-specific information.
This would be done the same as what already exists for `Endpoint`s
This would be good to have that before the 2.0.0 release.

elsony:
I'd say yes for StarterProject as well. I don't see any reason for excluding StarterProject.

elsony:
Based on the discussion, we are moving this to 2.1 to discuss together with how we associate commands to projects in a single devfile.

elsony:
Can this be closed now given that PR #214 has been merged?

davidfestal:
Yes, it can be closed

elsony:
Closing this as per the last comment

PR #214, opened 2020-11-10
Add `attributes` free-form map field for tool-specific fields in a number of places

davidfestal:
### What does this PR do?

This PR adds the `attributes` map, to enable adding tool-specific fields, at the root of the following objects:
components, projects, starterProjects and commands.

The `attributes` map is not only a map of strings, but also allows adding any json-schema values in map values (YAML values in yaml documents or Json values in JSON documents). here is an example:

```yaml
components:
  - name: a-container
    attributes:
      tool1-string-attribute: a-string
      tool2-object-attribute:
        free-form-string-field: "value" 
        free-form-boolean-field: true 
      tool3-array-attribute:
        - value1 
        - value2 
```

This PR also updates the `attributes` already present in the container `endpoints` definition to allow objects in map values. This is **not a breaking change** though since the current proposal also covers the "map of strings" case.

In go the attribute values would be retrieved through one of the `Attributes` methods, some of which are demonstrated in the following snippet:

```go
component : ...
attributes := component.Attributes

var err error

withoutErrorChecking := attributes.GetString("tool1-string-attribute")

withErrorChecking := attributes.GetString("tool1-string-attribute", &err)

myStruct := struct {
  StringField string `json:"free-form-string-field"`
  BooleanField string `json:"free-form-boolean-field"`
}

err = attributes.GetInto("tool2-object-attribute", &myStruct)

var arrayValues []interface{} = attributes.Get("tool3-array-attribute")
```
 
This fixes issue https://github.com/devfile/api/issues/169

Yes, it is.


pkg/apis/workspaces/v1alpha2/attributes.go
```diff
@@ -0,0 +1,9 @@
+package v1alpha2
+
+import (
+	apiext "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+)
+
+// Attributes provides a way to add a map of arbitrary YAML/JSON
+// objects.
+type Attributes map[string]apiext.JSON
```
sleshchenko, line 9: from the model point of view, it's so useful to have any JSON in the field, but from client(odo, devworkspace) perspective of view, it means that schema does not validate attributes format and every field access requires checking that attributes have the right format, if I did not miss something. So, we need some library code that allows easily to get attribute value as string, int, interface or error if the field has the wrong format.
Am I missing something and it's not going to be an issue for clients to read the attributes?

davidfestal: You're right, I'm looking into it.

amisevsk: To add here, I don't know how elegant or useful embedding full json as an attribute will ultimately be. Even with `map[string]string` you _could_ ostensibly use json, but the main problem is that it's hard to write. I think I'd rather see something like
```yaml
attributes:
  controller.devfile.io/plugin/name: "vscode-java"
  controller.devfile.io/plugin/vscode-extensions/java: "https://my-url.com/java.vsix"
  controller.devfile.io/plugin/vscode-extensions/java-debug: "https://my-url.com/java-debug.vsix"
```
than have the above collapsed into a json blob
```yaml
attributes:
  controller.devfile.io/plugin: '{"name": "vscode-java", "plugins": ["java.vsix", "java-debug.vsix"]}'
```
The latter could get quite complicated as there are longer entries. It also forces attributes to be complex, as we couldn't use something like `my-attribute: my-value`.

amisevsk (commented):
I have no suggestions apart from the point brought up by Serhii

pkg/apis/workspaces/v1alpha2/attributes.go
davidfestal: > than have the above collapsed into a json blob

@amisevsk we just don't need to have json collasped into a JSON blob.
This is a json-like arbitrary value, directly included inside the JSON or YAML document.
So in a YAML document it will a yaml fragment, and in a JSON document it will be a JSON document.

Your example would be:

```yaml
    attributes:
      name: vscode-java
      plugins:
        - java.vsix
        - java-debug.vsix
```

That's the whole point of the change.

davidfestal: And with the helpers I'm currently adding to the PR, when getting the GO structs, you would extract it like that:

```go
var Attributes attributes = ...
myName, isAString := attributes.Get("name").Interface().(string)

// And for plugins you might use want to use the `DecodeInto` method, to decode the arbitrary JSON-like structure into a given, well-known GO type. 

plugins := &[]string{}
err := attributes.Get("plugins").DecodeInto(plugins)

// Of course you also have a method to know if a given key is available in the Attributes:

isThere := attributes.Exists("aMissingKey")
```

amisevsk: This makes sense to me and is a lot cleaner than my suggestion :)

davidfestal:
I simplified the `attributes` api helpers and added many tests.

@amisevsk @sleshchenko on this personnal branch https://github.com/amisevsk/devworkspace-operator/compare/use-v1alpha2...davidfestal:use-v1alpha2-and-new-attributes you will find examples of use in the changes I made on @amisevsk `use-v1alpha2` to accomodate the new attributes in the `endpoints` case.

Still need to document public methods in the new `attributes` package though.

pkg/apis/workspaces/v1alpha2/commands.go
```diff
@@ -58,7 +58,11 @@ type Command struct {
 	// Mandatory identifier that allows referencing
 	// this command in composite commands, from
 	// a parent, or in events.
-	Id           string `json:"id"`
+	Id string `json:"id"`
+	// Map of implementation-dependant free-form YAML attributes.
+	// +optional
+	// +devfile:overrides:include:omit=true
+	Attributes   attributes.Attributes `json:"attributes,omitempty"`
```
amisevsk, line 65: Is this formatted? I don't know if field type should align with struct tags below

pkg/attributes/attributes.go
```diff
@@ -0,0 +1,353 @@
...
+// objects.
+type Attributes map[string]apiext.JSON
+
+// GetDecodedInto allows decoding the attribute with the given key
+// into a given interface. The provided interface should be a pointer
+// to a struct, to an array, or to any simple type.
+//
+// An error is returned if the provided interface type is not compatible
+// with the attribute content
+func (attributes Attributes) GetDecodedInto(key string, into interface{}) error {
```
amisevsk, line 20: I'd suggest `DecodeKeyInto` as a function name here, or something similar. This name might be too similar to `DecodeInto` below.

Also I think `DecodeInto` and `GetDecodedInto` should be adjacent in this file.

pkg/attributes/attributes.go
```diff
@@ -0,0 +1,353 @@
...
+// a map of interfaces or an array of interfaces.
+//
+// An optional error holder can be passed as an argument
+// to receive any error that might have occured during the attribute
+// decoding
+func (attributes Attributes) Get(key string, errorHolder ...*error) interface{} {
+	if attribute, exists := attributes[key]; exists {
+		container := &[]interface{}{}
+		err := json.Unmarshal([]byte("[ "+string(attribute.Raw)+" ]"), container)
+		if err != nil && len(errorHolder) > 0 && errorHolder != nil {
```
amisevsk, line 43: I think we should return the error instead of passing in a struct optionally. This approach makes it very easy to hide ignoring a potential error.

pkg/apis/workspaces/v1alpha2/attributes.go
maysunfaisal: I am wondering if it makes sense to have these helper functions in devfile/api or devfile/library

amisevsk (approved):
Generally LGTM

pkg/apis/workspaces/v1alpha2/attributes.go
davidfestal: @maysunfaisal In fact it is there, because my plan was to also use it in the `DevfileMetadata` struct, which is also in the `devfile/api` repo, but in the non-K8S-versioned `devfile` package.

pkg/devfile/header.go
```diff
@@ -22,4 +26,6 @@ type DevfileMetadata struct {
 	// +optional
 	// +kubebuilder:validation:Pattern=^([0-9]+)\.([0-9]+)\.([0-9]+)(\-[0-9a-z-]+(\.[0-9a-z-]+)*)?(\+[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$
 	Version string `json:"version,omitempty"`
+
+	attributes.Attributes `json:",inline"`
```
maysunfaisal, line 30: Should it be `json:"attributes,omitempty"`

davidfestal: I was trying to have the free-form attributes at the same level as the main metadata, but this is obviously not really possible.
I finally changed it according to your suggestion in my last commit:  https://github.com/devfile/api/pull/214/commits/1a8f497e4106b5c5b23b61d020b0d9df4799e455

pkg/apis/workspaces/v1alpha2/commands.go
davidfestal: It is yes. That's probably a bug of the formatter itself.

pkg/attributes/attributes.go
davidfestal: The whole point of this approach is to avoid requiring a multiple value return, in order to be able to chain calls when we know there is no reason for an error.

davidfestal: But here I can change it. I'd be more reluctant to change it for the `GetString()`... methods, since it's quite useful to be able to chain them or directly use the result in an expression.

amisevsk (approved):
LGTM after updates.

pkg/attributes/attributes.go
```diff
@@ -0,0 +1,368 @@
...
+// Booleans allows returning only the attributes whose content
+// is a JSON boolean.
+//
+// An optional error holder can be passed as an argument
+// to receive any error that might have be raised during the attribute
+// decoding
+func (attributes Attributes) Booleans(errorHolder ...*error) map[string]bool {
+	result := map[string]bool{}
+	for key := range attributes {
+		if value, isRightType := attributes.Get(key, errorHolder...).(bool); isRightType {
```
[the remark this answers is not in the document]
davidfestal: I thought about gathering errors, but didn't want to add a dependency to a non-standard error package to tackle error gathering. Didn't put much thought in it though.
So for now only the last error is returned.

pkg/attributes/attributes.go
```diff
@@ -0,0 +1,368 @@
...
+// to receive any error that might have be raised during the attribute
+// decoding
+func (attributes Attributes) GetBoolean(key string, errorHolder ...*error) bool {
+	if attribute, exists := attributes[key]; exists {
+		result := new(bool)
+		err := json.Unmarshal(attribute.Raw, result)
+		if err == nil {
+			return *result
+		}
+		if len(errorHolder) > 0 && errorHolder != nil {
```
sleshchenko, line 94: IDE suggests that `Condition is always 'true' because 'errorHolder' is always not 'nil'` here and in other places in this file.

pkg/attributes/attributes.go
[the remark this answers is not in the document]
sleshchenko: OK with that, but maybe it worths a few words in the documentation.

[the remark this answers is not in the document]
davidfestal: You're right. Let me do it.

pkg/attributes/attributes.go
sleshchenko: > when we know there is no reason for an error

Can we identify when we know when there is a reason for the error and when there is no?
Like in the shared example https://github.com/amisevsk/devworkspace-operator/compare/use-v1alpha2...davidfestal:use-v1alpha2-and-new-attributes#diff-cee3fd6d4114264b1f235c57bd27c71a9c40cd809efa214a395f5e4cfee8a92fR80
is it safe to propagate an empty string in case the user misconfigured protocol or path as an object?
It shows that such an aproach forces client not to check the erorr. As an alternative, we could have two separate methods(for each type) one would return the error, another will ignore the error but would be named in the way client understand error is suppressed. Like GetStringOrDefault(), default depends on type: string - "", int - 0, interface - {}.

pkg/attributes/attributes.go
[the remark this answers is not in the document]
davidfestal: Add comments whenever relevant in commit https://github.com/devfile/api/pull/214/commits/cf8f396b3d6352db2280c80030c5d3e6a368eeb3

pkg/attributes/attributes.go
davidfestal: Good catch ! Thanks !

davidfestal: Fixed in commit https://github.com/devfile/api/pull/214/commits/ed4b1865c5d7ee354d764eb777f116dba96af9d4

pkg/attributes/attributes.go
sleshchenko: More relevant case, where some boolean value is mistyped by user, like public: "true", instead of public: true. Then user will get unexpected behavior instead of error:
```go
func TestBooleanAsString(t *testing.T) {
	stringMap := Attributes{}.FromStringMap(map[string]string{
		"public": "true",
	})
	publicAttr := stringMap.GetBoolean("public")
	assert.Equal(t, true, publicAttr)
}
```

davidfestal:
@maysunfaisal @elsony @l0rd @yangcao77 

Do you want to review this PR after the very last changes ?

sleshchenko (approved):
Good job.

I don't have the ideal solution to propose with error handling, so approving.

About string/boolean... Maybe to avoid unexpected behavior, we need to implement autoconversion between primitive types, because if the format allows and user defines like `true` instead of `"true"`, or `"123"` instead of `123`, there is no a real reason why we should fail it with a validation error or start with default values.
Or we just need to declare that clients need to validate attributes they expect before using, for devworkspace controller it may be done with webhooks and then it should be pretty safe to ignore error on reconciling.

davidfestal:
> About string/boolean... Maybe to avoid unexpected behavior, we need to implement autoconversion between primitive types, because if the format allows and user defines like `true` instead of `"true"`, or `"123"` instead of `123`, there is no a real reason why we should fail it with a validation error.

I've implemented that in commit https://github.com/devfile/api/pull/214/commits/5640a2403c18ceaf91749b0bcb47922f948718e9
and added relevant tests.

@amisevsk @sleshchenko Do you think you could add a last review step after this commit ?

maysunfaisal:
@davidfestal looks good 👍

maysunfaisal (approved):

yangcao77 (approved):
thanks David for this PR, good changes! lgtm

pkg/attributes/attributes.go
```diff
@@ -0,0 +1,448 @@
...
+// Exists returns `true` if the attribute with the given key
+// exists in the attributes map.
+func (attributes Attributes) Exists(key string) bool {
+	_, exists := attributes[key]
+	return exists
+}
+
+type convertPrimitiveFunc func(attributes Attributes, key string, attributeType string) (interface{}, error)
+
+func (attributes Attributes) getPrimitive(key string, zeroValue interface{}, resultType string, convert convertPrimitiveFunc, errorHolder ...*error) interface{} {
```
amisevsk, line 38: Couldn't we derive `resultType` from `zeroValue`? Would it be possible to implement this function as
```go
func (attributes Attributes) getPrimitive(key string, into interface{}, convert convertPrimitiveFunc, errorHolder ...*error) {
//....
}
```
so that it could e.g. be called
```go
var value string
attributes.getPrimitive(myKey, &value, ...)
```
switching based on a raw string describing the type feels very similar to just switching on a `reflect.Type`

pkg/attributes/attributes.go
```diff
@@ -0,0 +1,448 @@
...
+			if retryError == nil && convertedValue != nil {
+				return convertedValue
+			}
+		}
+	} else {
+		err = errors.New("Attribute with key '" + key + "' does not exist")
+	}
+	if len(errorHolder) > 0 && errorHolder[0] != nil {
+		*errorHolder[0] = err
+	}
```
amisevsk, line 72: I'm not clear on what we're doing here -- why are we passing in a variadic error if we only ever set the first element?

amisevsk (approved):
A couple of questions but otherwise LGTM.

pkg/attributes/attributes.go
sleshchenko: As far as I understand how optional parameters could be achieved in go. So, we could call specify `GetString()` instead of `GetString(nil)`. But it allows us to call `GetString(err, err, err, err)` and the only first is used.

I'm OK with both cases with their pros:
- `GetString()`: which is kind of syntax sugar;
- `GetString(nil)`: is a bit more explicit and suggests that error is ignored - in case you know parameter type;

pkg/attributes/attributes.go
```diff
@@ -32,70 +33,137 @@ func (attributes Attributes) Exists(key string) bool {
...
-// not a JSON boolean, then the result will be the `false` zero value
-// and an error is raised.
+// not a JSON boolean  (or a JSON string that can be converted into a JSON boolean),
+// then the result will be the `false` zero value and an error is raised.
+//
+// String values can be converted to boolean values according to the following rules:
+//
+// - strings "1", "t", "T", "TRUE", "true", and "True" will be converted to a `true` boolean
+//
+// - strings "0, "f", "F", "FALSE", "false", "False" will be converted to a `true` boolean
```
sleshchenko, line 141:
```suggestion
// - strings "0, "f", "F", "FALSE", "false", "False" will be converted to a `false` boolean
```

pkg/attributes/attributes.go
davidfestal: well, I somehow tried, but since I didn't want to use reflection, and wanted to keep it simple (it's just an internal private helper used in 3 functions), I chose this design.

pkg/attributes/attributes.go
davidfestal: You're right @sleshchenko. I just used variadic parameter to provide an optional parameter.
However in this form this API also allows, in the future, passing an array of errors and possibly getting as many errors as the provide variadic list allows, instead of only returning the last error encountered.

sleshchenko (approved):

pkg/attributes/attributes.go
sleshchenko: >  passing an array of errors and possibly getting as many errors as the provide variadic list allows, instead of only returning the last error encountered.

Has client initialize errors instances? I mean the client is not able to consume the items provided in the array, is it?
If not, then how you see it in the future? Like client initializes 100 errors and some of them are modified?

davidfestal: @sleshchenko I'm OK to change it to `GetString(key string, errorHolder *error)`,
and enforce the client to explicitly pass a `nil` error pointer in the second argument to clearly state it ignores the error.

@maysunfaisal @yangcao77 @amisevsk any preference between the use of an optional parameter through the use of variadics or a single error pointer ?

pkg/attributes/attributes.go
amisevsk: Fair enough.

pkg/attributes/attributes.go
amisevsk: I think an optional parameter would be nice to have, but also allows us to implicitly skip error checking with no way to easily notice it in code. Without knowing the API we're using in detail, it might be hard to notice e.g. 
```go
  myString := attributes.GetString("myKey")
```
is discarding an error silently.

We're kind of departing general Go paradigms for error handling (which I'm not a fan of anyways), so it's hard to reason about what makes sense here. I'm generally fine with whatever we choose, but it might be confusing for (eventual) newcomers.

davidfestal: So maybe @sleshchenko proposal of using `GetString("aString", nil)` is the best here ?This would still allow using `GetString` in expressions (by avoiding multiple value returns) while avoiding silent error discarding ?

davidfestal:
As deiscussed today, we're waiting for you GO to merge this @elsony, based of the devfile parser library readiness.

elsony:
@davidfestal I have checked with the team. It turns out that the parser code is currently pulling in a specific tag of the schema so merging this change is not going to break the parser code. I also checked with @mmulholl and he'll adjust his test PR to accommodate that. Given that his test code has not been merged yet, so it is not going to be a problem. So feel free to merge this PR whenever you think is ready.

pkg/attributes/attributes.go
davidfestal: Done in commit https://github.com/devfile/api/pull/214/commits/2c5bf7950b093c7edbc8c77b3a6d2463e178ddc5

l0rd (approved):

Issue #217, opened 2020-11-16
Add an optional artifact field in devfile metadata


yangcao77:
Currently the devfile lacks of ability to allow consumers to provide more information that might be useful, but is not defined in devfile schema. 
For example,
the additional information on dockerfile and deployment manifest in odo devfile registry:
https://github.com/odo-devfiles/registry/blob/6b942e596a6bc1f4bec3123a60f89210fa95666c/devfiles/nodejs/devfile.yaml#L5-L6

and

the additional information on dockefile location that will be used in dev-console

Proposing to add an optional artifact field in devfile metadata to allow other unknown information

```
type MetadataArtifact struct {
	Name  string `json:"name"`
	Value string `json:"value"`
}

type DevfileMetadata struct {

	// Optional devfile name
	Name string `json:"name,omitempty"

	// Optional semver-compatible version
	Version string `json:"version,omitempty"

        // Optional devfile metadata artifact
	Artifact []MetadataArtifact  `json:"artifact,omitempty"
}
```

yangcao77:
closing this issue. Dup of #169

PR #218, opened 2020-11-16
introduce metadata artifact

yangcao77:

### What does this PR do?
This PR introduces an optional artifact field in metadata, which allows devfile consumers to provide additional information that is not defined in schema

#217


davidfestal (commented):
Just a heads-up that, in the context of issue https://github.com/devfile/api/issues/169 (adding free-form attributes for implementation-specific needs), I provided a mechanism that can be used commonly in both the body of the devfile *and* the devfile header.

[Here](https://github.com/devfile/api/pull/214/files#diff-43ae4c58f57a79e53cda9964a2073908de425be4e50b1932f6d513033c95690dR30) you can find how it is very simply added into the `DevfileHeader`.

The `DevfileHeader` metadata schema is already produced to allow adding non validated fields. The only missing thing was a way to store the and retrieve them, which should available now with the `Attributes` field.

Please note that the `Attributes` solution should be more general that the one provided in the PR, since it allows structured fields to be added, and not only string values.

yangcao77:
closing this PR. Dup of #214

PR #219, opened 2020-11-17
Initial proposal for Console devfile feature w/o Build Guidance Spec

maysunfaisal:

### What does this PR do?
This PR outlines a proposal for the OpenShift Console to consume build guidance within the boundaries of devfile spec 2.0.0 because the Build Guidance spec PR #127 is still an open discussion.

This proposal is for the POC PR https://github.com/openshift/console/pull/6321

### DIFF
--- a/docs/proposals/console/devfile-import.md
+++ b/docs/proposals/console/devfile-import.md
@@ -55,6 +55,10 @@ components:
         value: "bar"
 ```
 
+Without the build guidance devfile spec, the proposed change here is to mention the dockerfile location in the metadata attribute. The attributes is a free form key-value pair and in the above example, a `alpha.build-dockerfile` key holds the dockerfile location; which would have otherwise been specified by the Dockerfile Build Guidance spec.
+
+Asides the devfile change, the POC PR would need to be updated to use the above devfile's image and endpoint. 
+
 Console's devfile import POC is using a `BuildConfig` to build the Dockerfile from the dockerfile path location, currently found from the context dir. The POC PR `BuildConfig` outputs it to an `ImageStreamTag`. However, the POC PR assumes the `ImageStream` and `ImageStreamTag` are all the same name as the application name. This is tightly coupled and dependant on the information entered in the Console Devfile Import form and thus does not allow us to mention a custom image name in the devfile.
 
 If we want to decouple `ImageStream` and `ImageStreamTag` from the application name, so that a custom image name can be specified in the devfile's component container; then we would need to update the following files: 
@@ -67,7 +71,7 @@ If we want to decouple `ImageStream` and `ImageStreamTag` from the application n
   
 This allows us to use the image name from devfile.yaml rather than always using the application name.
 
-The above proposed devfile container component can also mentions an endpoint instead of hardcoding a default 8080. This can be updated in `createOrUpdateResources()` in `frontend/packages/dev-console/src/components/import/import-submit-utils.ts`
+The proposed devfile also mentions an endpoint instead of hardcoding a default 8080. This can be updated in the POC PR's function `createOrUpdateResources()` in `frontend/packages/dev-console/src/components/import/import-submit-utils.ts`
 
 These devfile information would be parsed and returned by the library and thus ensuring a consistent UX.
 
--- a/docs/proposals/console/devfile-import.md
+++ b/docs/proposals/console/devfile-import.md
@@ -2,37 +2,38 @@
 
 This document outlines how the console is going to use a devfile 2.0.0 spec for it's import feature targeted for the upcoming release.
 
-## As-Is Today
+## Dev Preview POC
 
-Currently the Devfile import feature is mocked with a POC [openshift/console PR](https://github.com/openshift/console/pull/6321). The POC PR, requires the build guidance devfile spec to be implemented. However, the build guidance spec is still an open discussion in the [devfile/api PR](https://github.com/devfile/api/pull/127). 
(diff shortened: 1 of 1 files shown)


docs/proposals/console/devfile-import.md
```diff
@@ -0,0 +1,65 @@
...
+        value: "bar"
+```
+
+Console's devfile import POC is using a `BuildConfig` to build the Dockerfile from the dockerfile path location, currently found from the context dir. The POC PR `BuildConfig` outputs it to an `ImageStreamTag`. However, the POC PR assumes the `ImageStream` and `ImageStreamTag` are all the same name as the application name. This is tightly coupled and dependant on the information entered in the Console Devfile Import form and thus does not allow us to mention a custom image name in the devfile.
+
+If we want to decouple `ImageStream` and `ImageStreamTag` from the application name, so that a custom image name can be specified in the devfile's component container; then we would need to update 1. image stream name 2. build config output image stream tag 3. deployment's container image in `pkg/server/devfile-handler.go` and also the annotation mapping for the `ImageStreamTag` in `getTriggerAnnotation()` in `frontend/packages/dev-console/src/components/import/import-submit-utils.ts`. This allows us to use the image name from devfile.yaml rather than always using the application name.
```
elsony, line 58: Format the 3 points.

docs/proposals/console/devfile-import.md
```diff
@@ -0,0 +1,65 @@
...
+The console devfile import page has the Git Repo Url, Git Repo Ref & the Build Context Path
+
+<img src="https://user-images.githubusercontent.com/31771087/99319303-4ae89180-2837-11eb-8933-eaaf41160bcd.png">
+
+The Dockerfile is assumed to be present in the context directory. The Container Port is assumed to be 8080 and the Image Name is assumed to be the same as the application name.
+
+### Note:
+The context path is present in both the Console import form and the open devfile spec(Refer to the OpenShift Console Devfile Import screenshot above and the open Build Guidance PR screenshot below). Do we require two context dir information? Is the context dir in the Console form for devfile.yaml and the context dir in the build guidance spec for Dockerfile relative to devfile.yaml? Or do we always assume both the devfile.yaml and Dockerfile are always at the same dir level.
+
+<img src="https://user-images.githubusercontent.com/31771087/99319306-4c19be80-2837-11eb-9639-a5c130deb4ba.png">
```
elsony, line 27: We should clarify that the first phase of the implementation will only support DockerFile, SourceToImage will not be implemented.

docs/proposals/console/devfile-import.md
```diff
@@ -0,0 +1,65 @@
...
+
+The above proposed devfile container component can also mentions an endpoint instead of hardcoding a default 8080. This can be updated in `createOrUpdateResources()` in `frontend/packages/dev-console/src/components/import/import-submit-utils.ts`
+
+These devfile information would be parsed and returned by the library and thus ensuring a consistent UX.
+
+### Note:
+The above devfile proposal and POC PR assumes the `BuildConfig` outputs the build to an `ImageStreamTag` which is used by the OpenShift internal registry. To push the image to a non-OpenShift Image Registry, the `BuiidConfig` output can pushed to a private registry. OpenShift [doc](https://docs.openshift.com/container-platform/4.6/builds/managing-build-output.html) outlines how this can be achieved via a `BuildConfig` and pushes the build to a private or dockerhub registry. Pushing to a private registry requires secret configuration from the Docker config, and this OpenShift [doc](https://docs.openshift.com/container-platform/3.11/dev_guide/builds/build_inputs.html#using-docker-credentials-for-private-registries) explains how to achieve it.
```
elsony, line 65: To scope the work on the first phase, we should probably clarify that we are going to implement the ImageStream path only. We are not going to implement the private or dockerhub registry on the first phase of the implementation.

elsony (changes requested):

docs/proposals/console/devfile-import.md
maysunfaisal: Updated

docs/proposals/console/devfile-import.md
maysunfaisal: Updated

docs/proposals/console/devfile-import.md
maysunfaisal: Updated

docs/proposals/console/devfile-import.md
```diff
@@ -0,0 +1,77 @@
...
+
+ For phase 1, only the Dockerfile Build Guidance is to be implemented. SourceToImage(S2I) will be implemented in the future sprints.
+
+<img src="https://user-images.githubusercontent.com/31771087/99319306-4c19be80-2837-11eb-9639-a5c130deb4ba.png">
+
+## Proposed Changes
+
+With the target deadline approaching soon, we may need to provide an alternative path for the OpenShift Console to consume the devfile. With the Build Guidance spec [devfile/api PR](https://github.com/devfile/api/pull/127) still an open discussion, this document proposes a solution within the boundaries of the devfile 2.0.0 spec to achieve this.
+
+Here is a proposed 2.0.0 spec devfile that can support build guidance for OpenShift Console:
```
johnmcollier, line 35: Can you make it clear what exactly your proposal will add/change? It's a little hard to tell what's changing, apart from the `alpha.build-dockerfile` attribute.

johnmcollier (changes requested):

docs/proposals/console/devfile-import.md
maysunfaisal: That is perhaps the only change without the build guidance spec in the current devfile spec. I'll be more clear!

johnmcollier: Okay, that makes sense.

maysunfaisal: Updated

elsony (approved):
/lgtm

johnmcollier (approved):

Issue #220, opened 2020-11-18
Add "message" field to DevWorkspace status and column for kubectl get


amisevsk:
## Is your feature request related to a problem? Please describe.
There are many valid reasons for a DevWorkspace startup to fail and we represent status this via the `Failed` value for `.status.phase`. However, the only way to get insight into *why* a particular DevWorkspace failed to start is by inspecting the full yaml of the DevWorkspace and checking which Conditions are set.

## Describe the solution you'd like
We should add a field, e.g. `Message` or `StatusInfo` to `DevWorkspaceStatus` and include it in the output for `kubectl get`. This might result in output such as
```
$ kc get devworkspaces
NAME    WORKSPACE ID                PHASE    URL   INFO
theia   workspacecd070100795e4809   Failed         Plugin X not found in registry
```

cc: @davidfestal

sleshchenko:
Do we have any example when INFO may be not empty for not failed state?

JPinkney:
Personally, I like this idea. I dislike having to dig through the logs to find out why my devworkspace failed. I'd prefer it if we had some easy way to find out why it failed for users (and even ourselves)

davidfestal:
@amisevsk Do we want to use the same field names are those used in [`PodStatus`](https://v1-18.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#podstatus-v1-core) for consistency:

- *message*: A human readable message indicating details about why the pod is in this condition.
- *reason*: A brief CamelCase message indicating details about why the pod is in this state. e.g. `Evicted`

amisevsk:
> Do we have any example when INFO may be not empty for not failed state?

In general I'd use the field to communicate info that requires user intervention, since it cannot be hidden and a normal running state is captured by `phase = Running`. Ideally we'd hide the field when empty in the `get devworkspaces output`

> Do we want to use the same field names are those used in PodStatus for consistency:

That makes sense, I'm generally bad at naming things :D

johnmcollier:
Just chiming in here since I'm trying to do a similar thing for the registry operator - what about exposing the detailed status message as part of the `kubectl describe devworkspace <resource>` output on each state change?

Seems like that can be done with the manager's event recorder.

amisevsk:
> Just chiming in here since I'm trying to do a similar thing for the registry operator - what about exposing the detailed status message as part of the kubectl describe devworkspace <resource> output on each state change?

> Seems like that can be done with the manager's event recorder.

[...]

PR #221, opened 2020-11-18
Add additional field to devworkspace status for showing info to users

amisevsk:
### What does this PR do?
Adds a field to the status subresource of DevWorkspaces which can be used to provide additional information about e.g. workspace failures and is printed in the short-form output of commands like `kuberctl get devworkspaces`

Resolves https://github.com/devfile/api/issues/220, but should not be merged without discussion in the issue. I just wrote the PR since it's no effort to implement.

When applied to a cluster, output of `kubectl get devworkspaces` is something like 
```
kc get devworkspace
NAME    WORKSPACE ID                PHASE    INFO   URL
theia   workspaceb8b430aa340343da   Failed
```
(propagating failure reason to the status is not yet implemented on the controller side)

### DIFF
--- a/crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
+++ b/crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
@@ -6,19 +6,6 @@ metadata:
   creationTimestamp: null
   name: devworkspaces.workspace.devfile.io
 spec:
-  additionalPrinterColumns:
-  - JSONPath: .status.workspaceId
-    description: The workspace's unique id
-    name: Workspace ID
-    type: string
-  - JSONPath: .status.phase
-    description: The current workspace startup phase
-    name: Phase
-    type: string
-  - JSONPath: .status.ideUrl
-    description: Url endpoint for accessing workspace
-    name: URL
-    type: string
   group: workspace.devfile.io
   names:
     kind: DevWorkspace
@@ -33,7 +20,20 @@ spec:
     status: {}
   version: v1alpha1
   versions:
-  - name: v1alpha1
+  - additionalPrinterColumns:
+    - JSONPath: .status.workspaceId
+      description: The workspace's unique id
+      name: Workspace ID
+      type: string
+    - JSONPath: .status.phase
+      description: The current workspace startup phase
+      name: Phase
+      type: string
+    - JSONPath: .status.ideUrl
+      description: Url endpoint for accessing workspace
+      name: URL
+      type: string
+    name: v1alpha1
     schema:
       openAPIV3Schema:
         description: DevWorkspace is the Schema for the devworkspaces API
@@ -4063,6 +4063,10 @@ spec:
               ideUrl:
                 description: URL at which the Worksace Editor can be joined
                 type: string
+              message:
+                description: Message is a short user-readable message giving additional
+                  information about an object's state
+                type: string
               phase:
                 type: string
               workspaceId:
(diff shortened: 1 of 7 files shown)


davidfestal (commented):
see my comment on the issue: https://github.com/devfile/api/issues/220#issuecomment-729705822 about trying to be consistent with the corresponding fields in `PodStatus` for example ?

amisevsk:
@davidfestal Finally got around to updating this PR :)

amisevsk:
The conversion tests, as if to tell me they're actually useful, found a bug in conversion :). It is fixed now.

pkg/apis/workspaces/v1alpha2/devworkspace_types.go
```diff
@@ -67,6 +70,7 @@ const (
 // +kubebuilder:resource:path=devworkspaces,scope=Namespaced,shortName=dw
 // +kubebuilder:printcolumn:name="Workspace ID",type="string",JSONPath=".status.workspaceId",description="The workspace's unique id"
 // +kubebuilder:printcolumn:name="Phase",type="string",JSONPath=".status.phase",description="The current workspace startup phase"
+// +kubebuilder:printcolumn:name="Info",type="string",JSONPath=".status.statusInfo",description="Additional information about workspace state"
```
davidfestal, line 73:
```suggestion
// +kubebuilder:printcolumn:name="Info",type="string",JSONPath=".status.message",description="Additional information about workspace state"
```

davidfestal (changes requested):

pkg/apis/workspaces/v1alpha2/devworkspace_types.go
amisevsk: Woops, sorry.

crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
```diff
@@ -4074,7 +4078,24 @@ spec:
...
+  - additionalPrinterColumns:
+    - JSONPath: .status.workspaceId
+      description: The workspace's unique id
+      name: Workspace ID
+      type: string
+    - JSONPath: .status.phase
+      description: The current workspace startup phase
+      name: Phase
+      type: string
+    - JSONPath: .status.statusInfo
```
sleshchenko: I was confused why we have statusInfo and message at the same time, but PR check explained it in clear manner https://github.com/devfile/api/pull/221/checks?check_run_id=1567151866
CRD must be updated. Now it's fixed.

crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
```diff
@@ -33,7 +20,20 @@ spec:
     status: {}
   version: v1alpha1
   versions:
-  - name: v1alpha1
+  - additionalPrinterColumns:
```
sleshchenko, line 23: Is 
```
    - JSONPath: .status.message
      description: Additional information about workspace state
      name: Info
      type: string
```
is not propagated to v1alpha1.additionalPrinterColumns by purpose? I think we can do it safely since we already have added a message to the status.

sleshchenko (approved):
LGTM, just one question left

crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
amisevsk: I think it should be fine to add it, I was mostly just focussed on v1alpha2.

amisevsk:
The output for the current state of this PR is
```
NAME    WORKSPACE ID                PHASE      INFO   URL
theia   workspace210c0f0d88854521   Starting          http://workspace210c0f0d88854521-theia-3100.192.168.49.2.nip.io
```
Ideally we would be able to hide empty fields; I'm also considering moving INFO to be last in the list, but that may be confusing.

Failed workspaces still do not put anything there, as the changes aren't implemented in the controller.

amisevsk:
cc: @davidfestal waiting on your re-review

davidfestal (approved):

Issue #233, opened 2020-11-24
Enable attributes overriding


yangcao77:
The current `attributes` filed in components/commands/projects are not able to be overriden. 
As we discussed in the meeting, we want to enable attributes overriding to provide a way for user to inject attributes to existing devfile in registry/git repo

elsony:
Adding this to the 2.0 plan to be included as part of the 2.0 release as discussed.

elsony:
@yangcao77 Is there any remaining work on this item? Can we close this?

yangcao77:
This one can be closed.

PR #234, opened 2020-11-24
enable attributes overriding

yangcao77:

### What does this PR do?
This PR enables attributes overriding in commands,components and projects. 
Adds attributes fields in override objects.

Also change back to allow additional properties in metadata 

#233 
#236


maysunfaisal (commented):
lgtm

yangcao77:
@davidfestal @amisevsk May I get a 'lgtm' for this PR? Maysun's approval did not pass the review requirement.

maysunfaisal (approved):
Re-approving with lgtm.

Readded `PreserveUnknownFields` to the Metadata because odo devfile registries have devfiles with additional property in metadata and odo has already released with those registries. So this is needed to be backward compatible

pkg/devfile/header.go
```diff
@@ -11,6 +11,7 @@ type DevfileHeader struct {
 	// +kubebuilder:validation:Pattern=^([2-9])\.([0-9]+)\.([0-9]+)(\-[0-9a-z-]+(\.[0-9a-z-]+)*)?(\+[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$
 	SchemaVersion string `json:"schemaVersion"`
 
+	// +kubebuilder:pruning:PreserveUnknownFields
```
davidfestal, line 14: @yangcao77 Though this PR is already merged, I just have a question about the reason for adding this line. It would allow adding arbitrary fields directly inside the `metadata` object without any way to retrieve them from the API. Isn't it precisely the situation we fixed by adding `attributes` one level down ?

yangcao77: Without this line,  the schema validation is going to fail when there are additional properties in metadata. 

The devfile registries for Odo 2.0 have devfiles ( [example](https://github.com/odo-devfiles/registry/blob/0b0e73401e462324a8db32b60c625f56ad639391/devfiles/nodejs/devfile.yaml#L5-L6) ) contain additional properties in metadata.  Without this line, using latest Odo (integrated with latest parser) with a created project would cause failure.  Since Odo 2.0 has already released with those registries,  devfile 2.0 changes need to be backward compatible.

davidfestal: OK, make perfectly sense. Thanks for the explanation !

PR #235, opened 2020-11-24
Add pattern validation and maxlength validation to name and id

yangcao77:
### What does this PR do?
This PR adds pattern validation and maxlength validation to devfile name and id, to validation the following rule during schema validation

```
^([a-z0-9]([-a-z0-9]*[a-z0-9]))?$

limit to lowercase chars i.e.; no uppercase allowed
limit within 63 chars
no special chars allowed, only dash(-)
start with an alphanumeric character
end with an alphanumeric character

```

#201 

Yes.
Tested with name: `runtime@runtimeruntimeruntimeruntimeruntimeruntimeruntimeruntimeruntimeruntime`
got error:
```
invalid devfile schema. errors :
- components.0.name: String length must be less than or equal to 63
- components.0.name: Does not match pattern '^([a-z0-9]([-a-z0-9]*[a-z0-9]))?$'
```

### DIFF
--- a/crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
+++ b/crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
@@ -4335,6 +4335,8 @@ spec:
                           description: Mandatory identifier that allows referencing
                             this command in composite commands, from a parent, or
                             in events.
+                          maxLength: 63
+                          pattern: ^[a-z]([-a-z0-9]*[a-z0-9])?$
                           type: string
                         vscodeLaunch:
                           description: Command providing the definition of a VsCode
@@ -4516,6 +4518,8 @@ spec:
                                     - none
                                     type: string
                                   name:
+                                    maxLength: 63
+                                    pattern: ^[a-z]([-a-z0-9]*[a-z0-9])?$
                                     type: string
                                   path:
                                     description: Path of the endpoint URL
@@ -4602,6 +4606,8 @@ spec:
                                       containers mount the same volume name then they
                                       will reuse the same volume and will be able
                                       to access to the same files.
+                                    maxLength: 63
+                                    pattern: ^[a-z]([-a-z0-9]*[a-z0-9])?$
                                     type: string
                                   path:
                                     description: The path in the component container
@@ -4678,6 +4684,8 @@ spec:
                                     - none
                                     type: string
                                   name:
+                                    maxLength: 63
+                                    pattern: ^[a-z]([-a-z0-9]*[a-z0-9])?$
                                     type: string
                                   path:
                                     description: Path of the endpoint URL
@@ -4737,6 +4745,8 @@ spec:
                             component from other elements (such as commands) or from
                             an external devfile that may reference this component
                             through a parent or a plugin.
+                          maxLength: 63
+                          pattern: ^[a-z]([-a-z0-9]*[a-z0-9])?$
                           type: string
                         openshift:
                           description: Allows importing into the workspace the OpenShift
@@ -4780,6 +4790,8 @@ spec:
                                     - none
                                     type: string
                                   name:
+                                    maxLength: 63
+                                    pattern: ^[a-z]([-a-z0-9]*[a-z0-9])?$
                                     type: string
                                   path:
                                     description: Path of the endpoint URL
@@ -5037,6 +5049,8 @@ spec:
                                     description: Mandatory identifier that allows
                                       referencing this command in composite commands,
                                       from a parent, or in events.
+                                    maxLength: 63
+                                    pattern: ^[a-z]([-a-z0-9]*[a-z0-9])?$
                                     type: string
                                   vscodeLaunch:
                                     description: Command providing the definition
@@ -5222,6 +5236,8 @@ spec:
                                               - none
                                               type: string
                                             name:
+                                              maxLength: 63
+                                              pattern: ^[a-z]([-a-z0-9]*[a-z0-9])?$
                                               type: string
                                             path:
                                               description: Path of the endpoint URL
@@ -5314,6 +5330,8 @@ spec:
                                                 volume name then they will reuse the
                                                 same volume and will be able to access
                                                 to the same files.
+                                              maxLength: 63
+                                              pattern: ^[a-z]([-a-z0-9]*[a-z0-9])?$
                                               type: string
                                             path:
                                               description: The path in the component
@@ -5373,6 +5391,8 @@ spec:
                                               - none
                                               type: string
                                             name:
+                                              maxLength: 63
+                                              pattern: ^[a-z]([-a-z0-9]*[a-z0-9])?$
                                               type: string
                                             path:
                                               description: Path of the endpoint URL
@@ -5437,6 +5457,8 @@ spec:
                                       the component from other elements (such as commands)
                                       or from an external devfile that may reference
                                       this component through a parent or a plugin.
+                                    maxLength: 63
+                                    pattern: ^[a-z]([-a-z0-9]*[a-z0-9])?$
                                     type: string
                                   openshift:
                                     description: Allows importing into the workspace
@@ -5485,6 +5507,8 @@ spec:
                                               - none
                                               type: string
                                             name:
+                                              maxLength: 63
+                                              pattern: ^[a-z]([-a-z0-9]*[a-z0-9])?$
                                               type: string
                                             path:
                                               description: Path of the endpoint URL
@@ -5825,6 +5849,8 @@ spec:
                               description: Mandatory identifier that allows referencing
                                 this command in composite commands, from a parent,
                                 or in events.
+                              maxLength: 63
+                              pattern: ^[a-z]([-a-z0-9]*[a-z0-9])?$
                               type: string
                             vscodeLaunch:
                               description: Command providing the definition of a VsCode
@@ -6005,6 +6031,8 @@ spec:
                                         - none
                                         type: string
                                       name:
+                                        maxLength: 63
+                                        pattern: ^[a-z]([-a-z0-9]*[a-z0-9])?$
                                         type: string
                                       path:
                                         description: Path of the endpoint URL
@@ -6093,6 +6121,8 @@ spec:
                                           If several containers mount the same volume
                                           name then they will reuse the same volume
                                           and will be able to access to the same files.
+                                        maxLength: 63
+                                        pattern: ^[a-z]([-a-z0-9]*[a-z0-9])?$
                                         type: string
                                       path:
                                         description: The path in the component container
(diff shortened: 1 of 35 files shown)


pkg/apis/workspaces/v1alpha2/components.go
```diff
@@ -29,6 +29,8 @@ type Component struct {
 	// Mandatory name that allows referencing the component
 	// from other elements (such as commands) or from an external
 	// devfile that may reference this component through a parent or a plugin.
+	// +kubebuilder:validation:Pattern=^([a-z0-9]([-a-z0-9]*[a-z0-9]))?$
+	// +kubebuilder:validation:MaxLength=63
```
maysunfaisal, line 33: MinLength should be 1? It should error out if there is an empty string for some reason

maysunfaisal: Probably doesnt matter if name/id is mandatory

pkg/apis/workspaces/v1alpha2/commands.go
```diff
@@ -58,6 +58,8 @@ type Command struct {
 	// Mandatory identifier that allows referencing
 	// this command in composite commands, from
 	// a parent, or in events.
+	// +kubebuilder:validation:Pattern=^([a-z0-9]([-a-z0-9]*[a-z0-9]))?$
```
amisevsk, line 61: I think this pattern should be 
```suggestion
	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
```
the only difference is that it the empty string invalid. This also would match the regex used in k8s for [DNS labels](https://github.com/kubernetes/kubernetes/blob/53a9d106c4aabcd550cc32ae4e8004f32fb0ae7b/pkg/util/validation.go#L36)

pkg/apis/workspaces/v1alpha2/endpoint.go
```diff
@@ -47,6 +47,8 @@ const (
 )
 
 type Endpoint struct {
+	// +kubebuilder:validation:Pattern=^([a-z0-9]([-a-z0-9]*[a-z0-9]))?$
+	// +kubebuilder:validation:MaxLength=63
```
amisevsk, line 51: Should we have a shorted maxlength for endpoints? I believe k8s treats some routing objects (e.g. port names) differently (DNS subdomain, e.g. [here](https://github.com/kubernetes/kubernetes/blob/53a9d106c4aabcd550cc32ae4e8004f32fb0ae7b/pkg/util/validation.go#L54-L64))

amisevsk (commented):
Generally looks good, a few questions about length/regex below.

pkg/apis/workspaces/v1alpha2/endpoint.go
yangcao77: The code you pointed to was for DNS952. Do we need to follow the rules for DNS 952?

amisevsk: I'm not sure, and it's probably fine to use 63 chars for now. We just may want to keep in mind in the future that on the Kubernetes end, some things are more restricted.

pkg/apis/workspaces/v1alpha2/commands.go
yangcao77: `^[a-z0-9]([-a-z0-9]*[a-z0-9])?$` allows to create resource name with all numbers. 
I manually tried on openshift 4.1 on aws, creating an ingress/service with a numeric name would fail.
So I restrict the rule to be `^[a-z]([-a-z0-9]*[a-z0-9])?$`

pkg/apis/workspaces/v1alpha2/endpoint.go
yangcao77: In kube doc, it states the maxlength for name/id is 63 (https://kubernetes.io/docs/concepts/overview/working-with-objects/names/), which follows DNS1123. 
But for the regex, it is different, even the doc and regex states the name can start and end with an alphanumeric character, but when I manually tested, a numeric name is not valid.

elsony: Our main goal is to have the fields to be able to easily translate into K8 resources with minimal impact on other environments. A char limit of 24 looks fairly limiting and users can easily go beyond that. I'd suggest using the DNS 952 regex and allow 63 chars limited.

pkg/apis/workspaces/v1alpha2/commands.go
amisevsk: Makes sense -- good catch!

pkg/apis/workspaces/v1alpha2/endpoint.go
amisevsk: Yeah I've found kube documentation to be a little lacking here. I think the current system is good for now and we can potentially refine it if need be later.

crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
```diff
@@ -24,8 +24,6 @@ spec:
     kind: DevWorkspace
     listKind: DevWorkspaceList
     plural: devworkspaces
-    shortNames:
-    - dw
```
amisevsk, line 28: I think CRDs might need to be rebuilt after the rebase on master

amisevsk (approved):
One minor comment below. On the specific regex used, I found the more current validation code [here](https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go). It looks like kube uses label and subdomain from DNS1123 (`^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`), and label from DNS1035 (`"^[a-z]([-a-z0-9]*[a-z0-9])?$"`) but it's not easy to see which definition applies where, and which resources have additional restrictions. I think for now using the DNS1123 regex and a limit of 63 chars should be fine.

From the devworkspace operator perspective, we generally use generated names for actual kube resources and store the original id/name in a label, so specifics shouldn't be a huge issue there, but certain resources (e.g. container ports) can be tricky.

yangcao77:
> One minor comment below. On the specific regex used, I found the more current validation code here. It looks like kube uses label and subdomain from DNS1123 (^[a-z0-9]([-a-z0-9]*[a-z0-9])?$), and label from DNS1035 ("^[a-z]([-a-z0-9]*[a-z0-9])?$") but it's not easy to see which definition applies where, and which resources have additional restrictions. I think for now using the DNS1123 regex and a limit of 63 chars should be fine.

Do you mean we use DNS1035 regex and DNS1123  limit of 63 chars? As I previously posted in https://github.com/devfile/api/pull/235#discussion_r531097575 the service and ingress name follow DNS1035 rule.

amisevsk:
> Do you mean we use DNS1035 regex and DNS1123 limit of 63 chars? As I previously posted in #235 (comment) the service and ingress name follow DNS1035 rule.

What I'm trying to say is that unless we have a good reason for it, we may as well stick with the most permissive, basic one (DNS1123, 63 chars). From the devworkspace-operator perspective, in all cases I can think of names for services, ingresses, etc. will be autogenerated by the controller. I won't speak for other consumers (e.g. `odo`) so there may be additional reasons for restriction I'm unaware of.

yangcao77:
> > Do you mean we use DNS1035 regex and DNS1123 limit of 63 chars? As I previously posted in #235 (comment) the service and ingress name follow DNS1035 rule.
> 
> What I'm trying to say is that unless we have a good reason for it, we may as well stick with the most permissive, basic one (DNS1123, 63 chars). From the devworkspace-operator perspective, in all cases I can think of names for services, ingresses, etc. will be autogenerated by the controller. I won't speak for other consumers (e.g. `odo`) so there may be additional reasons for restriction I'm unaware of.

I tested with Odo, with DNS1123 rule in schema, it won't block any numeric name. Odo would fail on pod & ingress creation if an endpoint is with a numeric name. So I suggest to use DNS1035 regex for the pattern validation, and keep 63 limit for maxlength. also see Elson's comment: https://github.com/devfile/api/pull/235#discussion_r531246203
```
Failed to start component with name nodejs. Error: Failed to create the component: unable to create or update component: unable to create Deployment nodejs: Deployment.apps "nodejs" is invalid: spec.template.spec.containers[0].ports[0].name: Invalid value: "3000": must contain at least one letter or number (a-z, 0-9)
```

amisevsk:
> I tested with Odo, with DNS1123 rule in schema, it won't block any numeric name. Odo would fail on pod & ingress creation if an endpoint is with a numeric name. So I suggest to use DNS1035 regex for the pattern validation, and keep 63 limit for maxlength. also see Elson's comment: #235 (comment)

But it's still not that simple: 

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: test-pod
spec:
  containers:
    - command:
        - sleep
        - "3600"
      image: busybox
      name: pods-simple-container
      ports:
      - name: "3000a"
        containerPort: 8080
```
is valid but does not match the regex. Also, 
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: test-pod
spec:
  containers:
    - command:
        - sleep
        - "3600"
      image: busybox
      name: pods-simple-container
      ports:
      - name: "test1test2test34"
        containerPort: 8080
```
does match the DNS1035 but is invalid:
```
❯ kc apply -f test.yml
The Pod "test-pod" is invalid: spec.containers[0].ports[0].name: Invalid value: "test1test2test34": must be no more than 15 characters
```

amisevsk:
The specific issue in using `3000` is very suspicious to me. I can reproduce it with plain `kubectl`, but it seems like a serialization/deserialization issue. The message ("must contain a letter **or** a number") makes me think it's not properly deserializing at some point -- yaml is pretty bad about how numbers are handled (strings don't need to be quoted unless they *could* be interpreted as numbers).

yangcao77:
@amisevsk I see. How about we limit maxlength for Endpoint name to be 15, and 63 for the other name & id? For regex pattern, we still use DNS1123 rule for all name & id `(^[a-z0-9]([-a-z0-9]*[a-z0-9])?$)`, and do extra validation in our validator for the all number case?

amisevsk:
> How about we limit maxlength for Endpoint name to be 15, and 63 for the other name & id?

Makes sense to me :+1:

amisevsk (approved):
Thanks!

davidfestal:
@yangcao77 Sorry for coming late here, but I just realized that the test files in https://github.com/devfile/api/tree/master/pkg/utils/overriding/test-fixtures have not been updated to according to the new naming constraints implemented in this PR.

If you open those test files in vscode (opened from the vscode workspace file), you should see schema-related errors in them.

yangcao77:
@davidfestal Missed those files. I will open a PR to update those files EOD.

davidfestal:
> @davidfestal Missed those files. I will open a PR to update those files EOD.

Thanks !

Issue #172, opened 2020-10-13
Discuss regarding merging meta.yaml and devfile.yaml


GeekArthur:
**Is your feature request related to a problem? Please describe.**
Currently the registry index generator generates the `index.json` based on the fields in `meta.yaml` of each devfile component, we may need to use fields in `devfile.yaml` of each devfile component as well. Therefore, merging `meta.yaml` and `devfile.yaml` can help us centralize all the fields in one file then we can parse and generate `index.json` from one place.

Two things we need to discuss here:
- Does `meta.yaml` has any special usage for consumers (eg. Che, odo)?
- Is there anything that blocks us from merging `meta.yaml` and `devfile.yaml`?

**Describe alternatives you've considered**
Alternative 1: 
We still use `meta.yaml` and `devfile.yaml` separately + **DON'T ADD** fields from `meta.yaml` to `devfile.yaml` (Today's version) -> we need to handle the case where there is a conflict exists between `meta.yaml` and `devfile.yaml`

Alternative 2:
We still use `meta.yaml` and `devfile.yaml` separately + **ADD** fields from `meta.yaml` to `devfile.yaml` -> only parse `devfile.yaml` then generate `index.json`

@elsony @davidfestal Please feel free to add questions and comments and we can discuss from there, thanks!

GeekArthur:
I already checked the `meta.yaml` schema in che repo, to summary this, we need to add the following fields from `meta.yaml` to `devfile.yaml`
```
displayName: String
description: String
tags: String[]
icon: String
globalMemoryLimit: String
```

PR #248, opened 2020-12-02
Merge meta.yaml and devfile.yaml

GeekArthur:
Signed-off-by: jingfu wang <jingfu.j.wang@ibm.com>

### What does this PR do?
This PR adds new metadata fields to the existing devfile schema so that we can merge meta.yaml and devfile.yaml

Fixes https://github.com/devfile/api/issues/172

### DIFF
--- a/pkg/devfile/header.go
+++ b/pkg/devfile/header.go
@@ -30,4 +30,24 @@ type DevfileMetadata struct {
 	// Map of implementation-dependant free-form YAML attributes.
 	// +optional
 	Attributes attributes.Attributes `json:"attributes,omitempty"`
+
+	// Optional devfile display name
+	// +optional
+	DisplayName string `json:"displayName,omitempty"`
+
+	// Optional devfile description
+	// +optional
+	Description string `json:"description,omitempty"`
+
+	// Optional devfile tags
+	// +optional
+	Tags []string `json:"tags,omitempty"`
+
+	// Optional devfile icon
+	// +optional
+	Icon string `json:"icon,omitempty"`
+
+	// Optional devfile global memory limit
+	// +optional
+	GlobalMemoryLimit string `json:"globalMemoryLimit,omitempty"`
 }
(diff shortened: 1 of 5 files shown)


johnmcollier (approved):
Schema changes look fine to me.

But... I think we also need some folks from the devworkspace team to take a look at this and make sure it's not gonna affect anything on their end

GeekArthur:
> Schema changes look fine to me.
> 
> But... I think we also need some folks from the devworkspace team to take a look at this and make sure it's not gonna affect anything on their end

Agree, let me ask some folks to review it.
@davidfestal @amisevsk @sleshchenko Can you please help review this PR as I find you are the owner of this repo? Thanks.

pkg/devfile/header.go
```diff
@@ -30,4 +30,24 @@ type DevfileMetadata struct {
...
+	// +optional
+	Tags []string `json:"tags,omitempty"`
+
+	// Optional devfile icon
+	// +optional
+	Icon string `json:"icon,omitempty"`
+
+	// Optional devfile global memory limit
+	// +optional
+	GlobalMemoryLimit string `json:"globalMemoryLimit,omitempty"`
```
amisevsk, line 52: I'm not sure if `GlobalMemoryLimit` is needed (but I'm also not opposed to adding it). The story behind this field is that we needed a way to get the total memory required by a workspace, and this was hard to do from a very simple registry -- plugin memory requirements are stored in their own meta.yamls and can change arbitrarily while the devfile registry is running, and further the devfile itself can override the values set in the plugin. This field is *manually* set by the devfile author to smooth a use-case in Che, but the value here can drift apart from reality (if e.g. one of your plugins updates and starts using more memory). 

I wonder if the new devfile registry could offer some metrics/calculation to get this number directly without user input?

amisevsk (approved):
LGTM generally.

sleshchenko (approved):
Nothing to add apart from Angel's comment

pkg/devfile/header.go
GeekArthur: Thanks for bringing that up, AFAIK currently the new devfile registry/OCI devfile registry doesn't provide a mechanism to directly get the number of global memory limit, so for the timing being, let's still add this field. Given this PR is only for purely migrating fields from meta.yaml, I think we can merge this PR for now, then open separate issue to discuss the existence of the specific field if it's needed as I think that may or may not bring more discussion and design.

Issue #263, opened 2020-12-04
`Attributes` overriding fails when overriding for non primitive attributes (arrays or objects)


davidfestal:
**Describe the bug**

When trying to override non-primitive (arrays or objects) elements of an `attributes` field, this fails with the following error: 
`merging an object in json but data type is not struct, instead is: map`

**To Reproduce**
1. Create a new folder `pkg/utils/overriding/test-fixtures/patches/command-with-attributes` in the repository
2. Cteate the following 3 files inside this new repository:
  - `original.yaml`:
```yaml
commands:
  - attributes:
      objectAttributeToReplaceFully:
        attributeField: 9.9
    apply:
      component: "mycomponent"
    id: command-with-attributes-changed
```
  - `patch.yaml`:
```yaml
commands:
  - attributes:
      objectAttributeToReplaceFully:
        newAttributeField: newAttributeFieldValue
    apply: {}
    
    id: command-with-attributes-changed
```
  - `result.yaml`:
```yaml
commands:
  - attributes:
      objectAttributeToReplaceFully:
        newAttributeField: newAttributeFieldValue
    apply:
      component: "mycomponent"
    id: command-with-attributes-changed
```
3. In the file `pkg/utils/overriding/overriding_test.go`, replace the following line (311):
```go
t.Error("Received error but did not expect one")
```
by
```go
t.Error("Received error but did not expect one" + actual)
```
4. Run the overriding test at the root of the project:
```bash
go test ./pkg/utils/overriding
```
5. See the error in the Go test result:
```
--- FAIL: TestOverridingPatches (0.00s)
    --- FAIL: TestOverridingPatches/command-with-attributes (0.00s)
        overriding_test.go:311: Received error but did not expect onemerging an object in json but data type is not struct, instead is: map
FAIL
FAIL    github.com/devfile/api/pkg/utils/overriding     0.008s
FAIL
```

**Expected behavior**

The result of the attribute override should be the one described in the  `result.yaml` file above.

PR #264, opened 2020-12-04
Fixing `Attributes` overriding failure when overriding for non primitive attributes

davidfestal:
### What does this PR do?

This PR fixes a bug that prevents overriding non-primitive (arrays or objects) elements of an `attributes` field.

https://github.com/devfile/api/issues/263

Yes, tests are added in the PR itself.

### DIFF
--- a/pkg/utils/overriding/overriding.go
+++ b/pkg/utils/overriding/overriding.go
@@ -138,16 +138,15 @@ var _ strategicpatch.LookupPatchMeta = mapEnabledPatchMetaFromStruct{
 	strategicpatch.PatchMetaFromStruct{},
 }
 
-
 func (s *mapEnabledPatchMetaFromStruct) replaceMapWithSingleKeyStruct(key string, elementIsSlice bool) {
-	t := s.T	
+	t := s.T
 	if t.Kind() == reflect.Ptr {
 		t = t.Elem()
 	}
 
 	if t.Kind() == reflect.Map {
 		typeName := t.Name()
-		tag := `json:"` + key + `"` 
+		tag := `json:"` + key + `"`
 		elemType := t.Elem()
 		if typeName == "Attributes" {
 			if elementIsSlice {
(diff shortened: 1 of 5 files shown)


maysunfaisal:
@davidfestal can we also update the attributes.Get() function? For example, when we dont have the key ie it does not exist, can we update the error holder with that info? I was using that in devfile/library and it silently fails when that happens 

https://github.com/devfile/api/blob/master/pkg/attributes/attributes.go#L179

or I can open a new PR with the changes if you'd like.

elsony:
LGTM. I haven't marked this as approved to wait for responses to Mayun's comment.

pkg/utils/overriding/overriding.go
```diff
@@ -128,3 +129,56 @@ func ensureOnlyExistingElementsAreOverridden(spec *workspaces.DevWorkspaceTempla
...
 		spec, overrides)
 }
+
+type mapEnabledPatchMetaFromStruct struct {
+	strategicpatch.PatchMetaFromStruct
+}
+
+var _ strategicpatch.LookupPatchMeta = mapEnabledPatchMetaFromStruct{
+	strategicpatch.PatchMetaFromStruct{},
+}
```
amisevsk, line 139: If this is just asserting that `mapEnabledPatchMetaFromStruct` satisfies the `strategicpatch.LookupPatchMeta` interface, I think the canonical way to do this is
```
var _ strategicpatch.LookupPatchMeta = (*mapEnabledPatchMetaFromStruct)(nil)
```

amisevsk (approved):
LGTM.

sleshchenko (approved):

davidfestal:
> @davidfestal can we also update the attributes.Get() function? For example, when we dont have the key ie it does not exist, we should update the error holder with that info.. I was using the Get() in devfile/library and it silently fails when that happens
> 
> https://github.com/devfile/api/blob/master/pkg/attributes/attributes.go#L179
> 
> or I can open a new PR with the changes if you'd like.

@maysunfaisal I'd rather discuss this in a distinct issue if you don't mind. Considering the key absence as an error is something that probably can be discussed. There is an `Attributes.Exists(key string)` method to explicitly check if a given key exists, so we could consider simply returning the corresponding zero value from the `Get` methods without seeing this case an error.  

However I see in the code that currently there isn't a consistent behavior between the `Get<PrimitiveType>` functions and the `Get` function in this regard. So we should at least decide on a consistent way, and implement the change.

Do you think you could open an issue for it ? I'm +1 if you can also propose the associated PR :-)

pkg/utils/overriding/overriding.go
davidfestal: Thanks for the remark @amisevsk !
Fixed in commit https://github.com/devfile/api/pull/264/commits/e33a83451dbb9fcda54b712decc74fb234e50d78

Issue #227, opened 2020-11-19
Endpoint protocol and secure updates


mmulholla:
1. Currently the endpoint protocol is limited to "http", "ws", "tcp", "ws". This should be expanded to include "https" and "wss"

2. The description for the secure flag should be updated to make it clear that when it is set to true the protocol should be set to either https or wss.

Note also that parser documentation must make it clear that when the protocl is set, for example, to http, and the secure attribute is set to true it is the responsibilty of the consumer of the parser to convert http to https.  

This was discussed in the dev file cabal on Nov 19th: https://docs.google.com/document/d/1Zn6D1gNu1gDidtillni0hNAeU5pscfqwhx4mv3MmmwY

mmulholla:
I created a branch with some changes. You can see them here:

https://github.com/devfile/api/compare/master...mmulholla:protocolUpdates

Would appreciate feedback before I create a pull request. Also I am not sure if other changes are needed.

mmulholla:
This is now complete

PR #268, opened 2020-12-07
Protocol updates

mmulholla:
See #227 

Note, first commit is the change, second commit is the generated files.

I pulled the latest master before generating.

### DIFF
--- a/pkg/apis/workspaces/v1alpha2/endpoint.go
+++ b/pkg/apis/workspaces/v1alpha2/endpoint.go
@@ -92,7 +92,7 @@ type Endpoint struct {
 	Protocol EndpointProtocol `json:"protocol,omitempty"`
 
 	// Describes whether the endpoint should be secured and protected by some
-	// authentication process. This requires a protocol of "https" or "wss".
+	// authentication process. This requires a protocol of `https` or `wss`.
 	// +optional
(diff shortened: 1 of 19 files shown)


davidfestal (approved):

pkg/apis/workspaces/v1alpha2/endpoint.go
```diff
@@ -92,7 +92,7 @@ type Endpoint struct {
 	Protocol EndpointProtocol `json:"protocol,omitempty"`
 
 	// Describes whether the endpoint should be secured and protected by some
-	// authentication process
+	// authentication process. This requires a protocol of "https" or "wss".
```
davidfestal, line 95:
```suggestion
	// authentication process. This requires a protocol of `https` or `wss`.
```

We have implemented minimal markdown support for the Json schema descriptions generated from the GO comments. That's also used in the generated API reference.
So using backticks here will allow a nicer markdown rendering, like it is already done [here](https://github.com/devfile/api/pull/268/files?file-filters%5B%5D=.go&file-filters%5B%5D=.json#diff-075e68dfa603f8a68957e3f7e1d69ab13eeab7363188e5c6e9f7a13a6d6070fcR87) for example.

PR #270, opened 2020-12-07
Check for invalid key in Get()

### LINKED ISSUE #269: Attributes.Get does not err out on an absent key
**Describe the bug**
[Attributes.Get](https://github.com/devfile/api/blob/master/pkg/attributes/attributes.go#L178) does not handle error & cases when the key is not found, which is inconsistent with the other Get functions.

The Get() in this case fails silently when no attribute key is found.

**To Reproduce**
Call the Get() function on a key which does not exist. No error is generated in the error holder.

maysunfaisal:

### What does this PR do?
Handles err when key is invalid for Get()

Fixes #269 

Yes, added new unit tests

#### Docs PR

N/A

### DIFF
--- a/pkg/attributes/attributes.go
+++ b/pkg/attributes/attributes.go
@@ -185,6 +185,8 @@ func (attributes Attributes) Get(key string, errorHolder *error) interface{} {
 		if len(*container) > 0 {
 			return (*container)[0]
 		}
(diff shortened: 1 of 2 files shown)


davidfestal (approved):
Thanks @maysunfaisal !

PR #278, opened 2020-12-11
Define devfile versioning and release process

johnmcollier:
### What does this PR do?
This adds a proposal on how to handle versioning in the devfile spec, as well as outlining a process for releasing the devfile schema and Go modules.

### What issues does this PR fix or reference?
https://github.com/devfile/api/issues/150

### Is your PR tested? Consider putting some instruction how to test your changes
N/A

#### Docs PR
N/A

Issue #150, opened 2020-10-05
Define version management of K8S API versions and schema versions 


davidfestal:
**Is your feature request related to a problem? Please describe.**

We should define:
- how we manage K8S API versions and JSON schema versions. How they are related together.
- how we update them,
- how that relates to release eng..., GH tagging (for GO module / package imports)

elsony:
Designs on the K8s API versions and JSON schema versions can be found in https://docs.google.com/presentation/d/1ohM1HzPB59a3ajvB7rVWJkKONLBAuT0mb5P20_A6vLs/edit#slide=id.g9cf6e5096d_0_8

johnmcollier:
Proposal up for review at https://github.com/devfile/api/pull/278

I've also created an example `make-release.sh` script and GitHub actions to perform a release on my repository:
- https://raw.githubusercontent.com/johnmcollier/api/master/make-release.sh
- https://github.com/johnmcollier/api/blob/master/.github/workflows/release-schema.yaml

PR #280, opened 2020-12-16
Update Devfile API Go module to add version to module path

johnmcollier:
### What does this PR do?
This PR updates the Go module path of the Devfile API to add the major version to its path, allowing v2 (and higher) versions of the module to be published.

As discussed on yesterday's Devfile cabal, to be able to publish the Devfile API Go module with a version of v2 or higher (and to ensure the module's version matches the Devfile schema version), we decided to update the module path of the Devfile API module to include the major version (`v2` in this case). Some discussion on why this is needed can be found [here](https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher).

Once this PR is merged, consumers of the Devfile API Go module will need to update the import paths of the module's packages (`github.com/devfile/api/pkg/...` -> `github.com/devfile/api/v2/pkg/...`) when they update the version of the API module that they pull in that has these changes.

N/A, but related to https://github.com/devfile/api/issues/150 and https://github.com/devfile/api/pull/278 (I can open a dedicated issue if needed).

N/A

#### Docs PR
N/A

CC  @amisevsk @davidfestal @elsony @kadel @l0rd

### DIFF
--- a/generator/overrides/gen.go
+++ b/generator/overrides/gen.go
@@ -127,7 +127,7 @@ func (g Generator) Generate(ctx *genall.GenerationContext) error {
 		genutils.WriteFormattedSourceFile(fileNamePart, ctx, root, func(buf *bytes.Buffer) {
 			buf.WriteString(`
 import (
-	attributes "github.com/devfile/api/pkg/attributes"
+	attributes "github.com/devfile/api/v2/pkg/attributes"
 )
 
 `)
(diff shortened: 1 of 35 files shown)


amisevsk (approved):
LGTM as discussed.

sleshchenko (approved):

davidfestal (approved):

PR #282, opened 2020-12-16
Index server REST APIs docs

GeekArthur:
Signed-off-by: jingfu wang <jingfu.j.wang@ibm.com>

### What does this PR do?
This PR adds the docs for index server REST APIs so that consumers (eg. Che) can refer to them.

### DIFF
--- a/docs/proposals/registry/index-server-REST-API.md
+++ b/docs/proposals/registry/index-server-REST-API.md
@@ -4,7 +4,7 @@ This documentation explains how to use the index server REST APIs
 ## Gets registry index file  
 ### HTTP request
 ```
-GET http://<registry host>/index
+GET http://{registry host}/index
 ```
 ### Request parameters
 | Parameter | Description |
--- a/docs/proposals/registry/index-server-REST-API.md
+++ b/docs/proposals/registry/index-server-REST-API.md
@@ -167,6 +167,7 @@ curl http://devfile-registry.192.168.1.1.nip.io/index
 
 ## Gets registry stack devfile
 Gets the specific registry stack devfile content from HTTP response
+
 Note: this REST API only returns the content of `devfile.yaml`, it won't return other resources in the stack
 ### HTTP request
 ```
--- a/docs/proposals/registry/index-server-REST-API.md
+++ b/docs/proposals/registry/index-server-REST-API.md
@@ -35,7 +35,7 @@ curl http://devfile-registry.192.168.1.1.nip.io/index
     },
     "resources": [
       "devfile.yaml",
-      "meta.yaml"
+      "archive.tar"
     ]
   },
   {
@@ -49,7 +49,7 @@ curl http://devfile-registry.192.168.1.1.nip.io/index
     },
     "resources": [
       "devfile.yaml",
-      "meta.yaml"
+      "archive.tar"
     ]
   },
   {
@@ -67,7 +67,7 @@ curl http://devfile-registry.192.168.1.1.nip.io/index
     },
     "resources": [
       "devfile.yaml",
-      "meta.yaml"
+      "archive.tar"
     ]
   },
   {
@@ -85,7 +85,7 @@ curl http://devfile-registry.192.168.1.1.nip.io/index
     },
     "resources": [
       "devfile.yaml",
-      "meta.yaml"
+      "archive.tar"
     ]
   },
   {
@@ -103,7 +103,7 @@ curl http://devfile-registry.192.168.1.1.nip.io/index
     },
     "resources": [
       "devfile.yaml",
-      "meta.yaml"
+      "archive.tar"
     ]
   },
   {
@@ -122,7 +122,7 @@ curl http://devfile-registry.192.168.1.1.nip.io/index
     },
     "resources": [
       "devfile.yaml",
-      "meta.yaml"
+      "archive.tar"
     ]
   },
   {
@@ -140,7 +140,7 @@ curl http://devfile-registry.192.168.1.1.nip.io/index
     },
     "resources": [
       "devfile.yaml",
-      "meta.yaml"
+      "archive.tar"
     ]
   },
   {
@@ -159,7 +159,7 @@ curl http://devfile-registry.192.168.1.1.nip.io/index
     },
     "resources": [
       "devfile.yaml",
-      "meta.yaml"
+      "archive.tar"
     ]
   }
 ]
@@ -190,8 +190,6 @@ schemaVersion: 2.0.0
 metadata:
   name: nodejs
   version: 1.0.0
-  alpha.build-dockerfile: "https://raw.githubusercontent.com/odo-devfiles/registry/master/devfiles/nodejs/build/Dockerfile"
-  alpha.deployment-manifest: "https://raw.githubusercontent.com/odo-devfiles/registry/master/devfiles/nodejs/deploy/deployment-manifest.yaml"
 starterProjects:
   - name: nodejs-starter
     git:
--- a/docs/proposals/registry/index-server-REST-API.md
+++ b/docs/proposals/registry/index-server-REST-API.md
@@ -1,0 +1,239 @@
+# Index server REST APIs
+This documentation explains how to use the index server REST APIs
+
+## Gets registry index file  
+### HTTP request
+```
+GET http://<registry host>/index
+```
+### Request parameters
+| Parameter | Description |
+| --------- | ----------- |
+| registry host | the URL/ingress that exposes registry service |
+### Request body
+The request body must be empty.
+### Request example
+```
+curl http://devfile-registry.192.168.1.1.nip.io/index
+```
+### Response example
+```
+[
+  {
+    "name": "java-maven",
+    "displayName": "Maven Java",
+    "description": "Upstream Maven and OpenJDK 11",
+    "tags": [
+      "Java",
+      "Maven"
+    ],
+    "projectType": "maven",
+    "language": "java",
+    "links": {
+      "self": "devfile-catalog/java-maven:latest"
+    },
+    "resources": [
+      "devfile.yaml",
+      "meta.yaml"
+    ]
+  },
+  {
+    "name": "java-openliberty",
+    "displayName": "Open Liberty",
+    "description": "Open Liberty microservice in Java",
+    "projectType": "docker",
+    "language": "java",
+    "links": {
+      "self": "devfile-catalog/java-openliberty:latest"
+    },
+    "resources": [
+      "devfile.yaml",
+      "meta.yaml"
+    ]
+  },
+  {
+    "name": "java-quarkus",
+    "displayName": "Quarkus Java",
+    "description": "Upstream Quarkus with Java+GraalVM",
+    "tags": [
+      "Java",
+      "Quarkus"
+    ],
+    "projectType": "quarkus",
+    "language": "java",
+    "links": {
+      "self": "devfile-catalog/java-quarkus:latest"
+    },
+    "resources": [
+      "devfile.yaml",
+      "meta.yaml"
+    ]
+  },
+  {
+    "name": "java-springboot",
+    "displayName": "Spring Boot®",
+    "description": "Spring Boot® using Java",
+    "tags": [
+      "Java",
+      "Spring"
+    ],
+    "projectType": "spring",
+    "language": "java",
+    "links": {
+      "self": "devfile-catalog/java-springboot:latest"
+    },
+    "resources": [
+      "devfile.yaml",
+      "meta.yaml"
+    ]
+  },
+  {
+    "name": "java-vertx",
+    "displayName": "Vert.x Java",
+    "description": "Upstream Vert.x using Java",
+    "tags": [
+      "Java",
+      "Vert.x"
+    ],
+    "projectType": "vertx",
+    "language": "java",
+    "links": {
+      "self": "devfile-catalog/java-vertx:latest"
+    },
+    "resources": [
+      "devfile.yaml",
+      "meta.yaml"
+    ]
+  },
+  {
+    "name": "nodejs",
+    "displayName": "NodeJS Runtime",
+    "description": "Stack with NodeJS 12",
+    "tags": [
+      "NodeJS",
+      "Express",
+      "ubi8"
+    ],
+    "projectType": "nodejs",
+    "language": "nodejs",
+    "links": {
+      "self": "devfile-catalog/nodejs:latest"
+    },
+    "resources": [
+      "devfile.yaml",
+      "meta.yaml"
+    ]
+  },
+  {
+    "name": "python",
+    "displayName": "Python",
+    "description": "Python Stack with Python 3.7",
+    "tags": [
+      "Python",
+      "pip"
+    ],
+    "projectType": "python",
+    "language": "python",
+    "links": {
+      "self": "devfile-catalog/python:latest"
+    },
+    "resources": [
+      "devfile.yaml",
+      "meta.yaml"
+    ]
+  },
+  {
+    "name": "python-django",
+    "displayName": "Django",
+    "description": "Python3.7 with Django",
+    "tags": [
+      "Python",
+      "pip",
+      "Django"
+    ],
+    "projectType": "django",
+    "language": "python",
+    "links": {
+      "self": "devfile-catalog/python-django:latest"
+    },
+    "resources": [
+      "devfile.yaml",
+      "meta.yaml"
+    ]
+  }
+]
+```
+
+## Gets registry stack devfile
+### HTTP request
+```
+GET http://{registry host}/devfiles/{stack}
+```
+### Request parameters
+| Parameter | Description |
+| ----------| ----------- |
+| registry host | the URL/ingress that exposes registry service |
+| stack | registry stack name |
+### Request body
+The request body must be empty.
+### Request example
+```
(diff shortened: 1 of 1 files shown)


docs/proposals/registry/index-server-REST-API.md
```diff
@@ -0,0 +1,239 @@
...
+    "displayName": "Maven Java",
+    "description": "Upstream Maven and OpenJDK 11",
+    "tags": [
+      "Java",
+      "Maven"
+    ],
+    "projectType": "maven",
+    "language": "java",
+    "links": {
+      "self": "devfile-catalog/java-maven:latest"
```
amisevsk, line 33: Do we have any specification for the index response, or is that part of a separate spec? I believe these self-links should match up with the example below, at least (e.g. `devfiles/nodejs`)

docs/proposals/registry/index-server-REST-API.md
```diff
@@ -0,0 +1,239 @@
...
+| --------- | ----------- |
+| registry host | the URL/ingress that exposes registry service |
+### Request body
+The request body must be empty.
+### Request example
+```
+curl http://devfile-registry.192.168.1.1.nip.io/index
+```
+### Response example
+```
```
amisevsk, line 20: We could add syntax highlighting:
```suggestion
```json
```

docs/proposals/registry/index-server-REST-API.md
```diff
@@ -0,0 +1,239 @@
...
+| registry host | the URL/ingress that exposes registry service |
+| stack | registry stack name |
+### Request body
+The request body must be empty.
+### Request example
+```
+curl http://devfile-registry.192.168.1.1.nip.io/devfiles/nodejs
+```
+### Response example
+```
```
amisevsk, line 184: As above
```suggestion
```yaml
```

docs/proposals/registry/index-server-REST-API.md
```diff
@@ -0,0 +1,239 @@
+# Index server REST APIs
+This documentation explains how to use the index server REST APIs
+
+## Gets registry index file
```
elsony, line 4: Can we add a description for each entry to briefly explain what each API is used for?

elsony (changes requested):

docs/proposals/registry/index-server-REST-API.md
GeekArthur: Yes, these self-links are from the index that is generated from current default registry https://github.com/devfile/registry, and the example below uses the current default registry as well. Regarding the generic specification of index/index response, it is in index generator module: https://github.com/devfile/registry-support/blob/master/index/generator/schema/schema.go

docs/proposals/registry/index-server-REST-API.md
GeekArthur: Sure, will update in the next commit.

docs/proposals/registry/index-server-REST-API.md
GeekArthur: Sure, will update in the next commit.

docs/proposals/registry/index-server-REST-API.md
GeekArthur: The description is similar to the heading, let me try to add more details to it.

docs/proposals/registry/index-server-REST-API.md
```diff
@@ -0,0 +1,241 @@
...
+| ----------| ----------- |
+| registry host | the URL/ingress that exposes registry service |
+| stack | registry stack name |
+### Request body
+The request body must be empty.
+### Request example
+```
+curl http://devfile-registry.192.168.1.1.nip.io/devfiles/nodejs
+```
+### Response example
```
johnmcollier, line 185: I think we need to be careful here.

As per your [PR](https://github.com/devfile/registry-support/pull/17), the response from the `devfiles/<name>` endpoint serves a byte array based on contents of the given stack in the devfile registry. What's returned won't necessarily be _just_ the devfile.yaml. It could also contain other stack resources. 

I'm also still skeptical that we need this endpoint, as this is something that should be done via the devfile registry library (by way of oras) and not over a rest API. Have we confirmed if this endpoint is something that Che needs or if it could be done exclusively through the library? CC @elsony @l0rd

johnmcollier (changes requested):

docs/proposals/registry/index-server-REST-API.md
GeekArthur: I don't think the byte array can contain other stack resources and it should *only* contain the devfile.yaml because we get the byte array from index server `pullStackFromRegistry` function and inside that function we only allow devfile media type https://github.com/devfile/registry-support/blob/master/index/server/main.go#L180, which means we only pull the devfile.yaml I think.

However regarding the necessity of this endpoint, let's double confirm with Che team to see if it's needed, we can remove it if it's not used by any consumer.

johnmcollier: Ah, I missed the part about only allowing the devfile media type. Good point.

In that case, I think we still need to update the doc here to be clear that only the devfile.yaml is returned, and nothing else from the stack is retrieved.

GeekArthur: Yep, I can emphasize that to avoid confusion.

elsony (approved):
LGTM. Approve based on the condition that we have confirmed the list of APIs needed by the Che team (especially around resource download APIs).

docs/proposals/registry/index-server-REST-API.md
```diff
@@ -0,0 +1,243 @@
...
+```yaml
+schemaVersion: 2.0.0
+metadata:
+  name: nodejs
+  version: 1.0.0
+  alpha.build-dockerfile: "https://raw.githubusercontent.com/odo-devfiles/registry/master/devfiles/nodejs/build/Dockerfile"
+  alpha.deployment-manifest: "https://raw.githubusercontent.com/odo-devfiles/registry/master/devfiles/nodejs/deploy/deployment-manifest.yaml"
```
l0rd, line 194: This looks like something that should be specified as a component of type kubernetes

docs/proposals/registry/index-server-REST-API.md
```diff
@@ -0,0 +1,243 @@
...
+```
+curl http://devfile-registry.192.168.1.1.nip.io/devfiles/nodejs
+```
+### Response example
+```yaml
+schemaVersion: 2.0.0
+metadata:
+  name: nodejs
+  version: 1.0.0
+  alpha.build-dockerfile: "https://raw.githubusercontent.com/odo-devfiles/registry/master/devfiles/nodejs/build/Dockerfile"
```
l0rd, line 193: This should be probably part of the spec instead of the metadata.

docs/proposals/registry/index-server-REST-API.md
l0rd: @elsony I am not sure I understand the question about the "download the corresponding resources of the stacks". What are those resources?

docs/proposals/registry/index-server-REST-API.md
```diff
@@ -0,0 +1,243 @@
...
+      "Maven"
+    ],
+    "projectType": "maven",
+    "language": "java",
+    "links": {
+      "self": "devfile-catalog/java-maven:latest"
+    },
+    "resources": [
+      "devfile.yaml",
+      "meta.yaml"
```
sleshchenko, line 38: Is it a really good example? I thought devfile and meta yamls are already merged into within https://github.com/devfile/api/pull/248
What are other possible resources apart from devfile.yaml?

docs/proposals/registry/index-server-REST-API.md
```diff
@@ -0,0 +1,243 @@
...
+The request body must be empty.
+### Request example
+```
+curl http://devfile-registry.192.168.1.1.nip.io/index
+```
+### Response example
+```json
+[
+  {
+    "name": "java-maven",
```
sleshchenko, line 24: Would we like to have an icon as a reference or base64 content?

docs/proposals/registry/index-server-REST-API.md
sleshchenko: So, is it a valid self-link? Or it should be 
```suggestion
      "self": "/devfiles/java-maven"
```
?

docs/proposals/registry/index-server-REST-API.md
sleshchenko: it does not seem to be allowed in the devfile metadata  all https://github.com/devfile/api/blob/071eecafa93c1f6455f5b6a8c34e57ebbdb6d27a/pkg/devfile/header.go#L20
So, it may be part of devfile or component attributes, but I don't have any idea about this kind of attributes

GeekArthur: These fields are directly copies from the registry repo: https://github.com/devfile/registry/blob/master/stacks/nodejs/devfile.yaml#L5-L6, seems like the [devfile API reference](https://docs.devfile.io/devfile/user-guide/api-reference.html) doesn't have those fields under metadata field, I am not sure why those fields exist in the registry repo. @johnmcollier @elsony I think we can remove those fields from registry repo if they are not necessary under metadata.

docs/proposals/registry/index-server-REST-API.md
GeekArthur: These fields are directly copies from the registry repo: https://github.com/devfile/registry/blob/master/stacks/nodejs/devfile.yaml#L5-L6, seems like the [devfile API reference](https://docs.devfile.io/devfile/user-guide/api-reference.html) doesn't have those fields under metadata field, I am not sure why those fields exist in the registry repo. @johnmcollier @elsony I think we can remove those fields from registry repo if they are not necessary under metadata.

docs/proposals/registry/index-server-REST-API.md
l0rd: @GeekArthur I think that currently Che dashboard needs both APIs. @sleshchenko @olexii4 can keep me honest.

sleshchenko: The dashboard needs devfile content since it allows users to modify it before applying to the cluster.

GeekArthur: @l0rd @sleshchenko Thanks for confirming this.

docs/proposals/registry/index-server-REST-API.md
GeekArthur: I will create another PR to update index specification in index generator module.

docs/proposals/registry/index-server-REST-API.md
elsony: That is not part of the formal devfile spec. The node.js stack (originally copied from the odo specific repo) contains some experimental flags that are odo specific for some temporary implementation of the odo deploy function. However, those are supposed to be formally defined and replaced by the devfile 2.2. Feel free to remove the alpha.* attributes from our repo.

GeekArthur: Thanks @elsony for confirming, I will remove them from our registry repo.

docs/proposals/registry/index-server-REST-API.md
GeekArthur: I will remove them from our registry repo.

GeekArthur: PR https://github.com/devfile/registry/pull/2 to remove them from our registry repo.

docs/proposals/registry/index-server-REST-API.md
GeekArthur: PR https://github.com/devfile/registry/pull/2 to remove them from our registry repo.

docs/proposals/registry/index-server-REST-API.md
amisevsk: My question (and I think what Serhii is getting at as well) is if the listed example's links are examples of what they will be in reality or just placeholders. The confusion is -- from these samples -- I would try a flow like
1. `curl http://devfile-registry.192.168.1.1.nip.io/index`
2. Get a self-link, e.g. `devfile-catalog/java-maven:latest` 
3. Try
    ```
    curl http://devfile-registry.192.168.1.1.nip.io/devfile-catalog/java-maven:latest
    ```
    to get the devfile I want.

However, `http://devfile-registry.192.168.1.1.nip.io/devfile-catalog/` will be an invalid path per the spec below:
```
GET http://{registry host}/devfiles/{stack}
```

GeekArthur: @amisevsk The self-link points to the stack devfile image (eg. docker image) instead of stack devfile content, so it's not related to and it's not used by the REST APIs in this PR, the index server can handle the self-link internally to access the stack devfile image.

Consumers or index server REST APIs users don't need to derive the URL to get the devfile content by using self-links, just follow docs `GET http://{registry host}/devfiles/{stack}` to update `registry host` and `stack` would be fine. For example, if user wants to get nodejs devfile content, just run `curl http://devfile-registry.192.168.1.1.nip.io/devfiles/nodejs`.

amisevsk: Ah, my confusion is because my assumption upon seeing a "self" link in an index would be to assume that it's a link to the file being indexed (i.e. the devfile). If it's a link to the image, why not key the link via `"image"` or something similar? Even though it's simple, `http://{registry host}/devfiles/{stack}` is more complicated than a self link in the index (and could break if for whatever reason this spec is changed). 

What will `curl http://devfile-registry.192.168.1.1.nip.io/devfile-catalog/java-maven:latest` return?

GeekArthur: I am not sure why originally we name it "self" instead of "image" and I agree with you it's a little bit misleading, to update it we can change the index generator as index generator generates the "self" field, we make the changes after we confirm there is no breaking changes.

IMO, the complexity of `http://{registry host}/devfiles/{stack}` and `http://{registry host}/{self}` should be the same because the difference between the two ways is to get {stack} vs to get {self}, and both {stack} and {self} can be get from index under specific stack with one query, so the complexity would be the same. However, I think `http://{registry host}/devfiles/{stack}` is more RESTful, which clearly locates the resource by using nouns with different levels.

`curl http://devfile-registry.192.168.1.1.nip.io/devfile-catalog/java-maven:latest` is invalid, because according to this index server REST APIs docs, currently we only serve two REST APIs below:
- To get index content: `GET http://{registry host}/index`
- To get devfile content: `GET http://{registry host}/devfiles/{stack}`

GeekArthur:
@l0rd @sleshchenko @johnmcollier @elsony @amisevsk Thanks for reviewing this PR, comments addressed, can you please review it again?

docs/proposals/registry/index-server-REST-API.md
amisevsk: > IMO, the complexity of http://{registry host}/devfiles/{stack} and http://{registry host}/{self} should be the same because the difference between the two ways is to get {stack} vs to get {self},

The difference here is that `http://{registry host}/devfiles/{stack}` is a convention we define in our spec, whereas `http://{registry host}/{self}` (where `{self}` is `devfiles/{stack}`) is provided to the consumer at the time of getting the index. If we go with the former, all users of the registry will need to be updated if there's e.g. a different structure ( `http://{registry host}/devfiles/v3/{stack}`). With the latter, the registry can put devfiles where it likes and consumers don't need to worry about it.

johnmcollier: Just chiming in here, initially we went with `self` as it meant a minimal number of changes to index.json, and was quick to prototype. I agree that we should change it, though I would say `stack` would be my preference over `image`.

elsony (approved):
lgtm

docs/proposals/registry/index-server-REST-API.md
GeekArthur: > The difference here is that http://{registry host}/devfiles/{stack} is a convention we define in our spec, whereas http://{registry host}/{self} (where {self} is devfiles/{stack}) is provided to the consumer at the time of getting the index. If we go with the former, all users of the registry will need to be updated if there's e.g. a different structure ( http://{registry host}/devfiles/v3/{stack}). With the latter, the registry can put devfiles where it likes and consumers don't need to worry about it.

I get your point, I agree that if the API versions get involve into this, to use the `self` can avoid users to specify the version explicitly, but I am wondering if the API supports multiple versions and when users want to choose which version to use, then we might still need to provide the version option somewhere (eg. selfv2, selfv3) and users still need to update that as well.

To make the change, I think the URL of the REST API design is still `http://{registry host}/devfiles/{stack}`, but if we do need a "self" as the shortcut for users we can the following things:
- Change the **current** `self` field to `image` or `stack`
- Add the `self` field to point to `devfiles/{stack}`

Which means the main URL definition is `http://{registry host}/devfiles/{stack}`, users can use either use `http://{registry host}/devfiles/{stack}` or `http://{registry host}/{self}` to get the content of the devfile, so that we can both provide a clear URL definition to users with RESTful design and let users get the devfile content with `self` placeholder without worrying about the structure, we can use separate issue and PR to implement that.

GeekArthur: @amisevsk Issue has been open: https://github.com/devfile/api/issues/297

johnmcollier (approved):

PR #285, opened 2020-12-18
Add tools for releasing the Devfile schema

johnmcollier:
### What does this PR do?
This PR adds tools for releasing and publishing the Devfile schema, as outlined [here](https://github.com/devfile/api/blob/master/docs/proposals/versioning-and-release.md):
- Adds a script to run that generates the schemas for the version being released (e.g. 2.0.0, 2.1.0, etc), and opens a pull request to merge them to the release branch.
    - Script is based on the `make-release.sh` from https://github.com/che-incubator/chectl/blob/master/make-release.sh
- Adds a GitHub workflow action that publishes the released schemas to the Devfile website when a release is created on GitHub.
    - For the time being, the GitHub release still needs to be manually created at first after merging the release PR to trigger the workflow.
    
Resolves https://github.com/devfile/api/issues/150

Yes.

For the `make-release.sh` script:
1) Ensure the `hub` CLI is installed and the `GITHUB_TOKEN` environment variable is set in your shell
2) Copy the script over from this PR to your fork of the repo
3) Run `./make-release.sh <some-schema-version> v1alpha2`

For the GitHub action, see https://github.com/johnmcollier/api/runs/1558896668?check_suite_focus=true for what a sample run looks like.

#### Docs PR
N/A

### DIFF
--- a/.github/workflows/release-schema.yaml
+++ b/.github/workflows/release-schema.yaml
@@ -1,0 +1,48 @@
+name: Release Devfile Schema
+
+# Triggers the workflow when a release is published on GitHub.
+on:
+  release:
+    types: [published]
+
+jobs:
+  release-json-schema:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout devfile docs
+        uses: actions/checkout@v2
+        with:
+          repository: devfile/docs
+          persist-credentials: false
+          path: docs-repo
+      - name: Checkout devfile api
+        uses: actions/checkout@v2
+        with:
+          path: api-repo
+      - name: Get the version being released
+        id: get_version
+        run: echo ::set-output name=version::$(cat api-repo/schemas/latest/jsonSchemaVersion.txt)
+      - name: Overwrite Stable Json Schema in Docs
+        run: mkdir -p docs-repo/docs/modules/user-guide/attachments/jsonschemas/stable && cp -f api-repo/schemas/latest/devfile.json docs-repo/docs/modules/user-guide/attachments/jsonschemas/stable/devfile.json
+      - name: Copy released Json Schema to Docs
+        run: mkdir -p docs-repo/docs/modules/user-guide/attachments/jsonschemas/${{ steps.get_version.outputs.VERSION }} && cp -f api-repo/schemas/latest/devfile.json docs-repo/docs/modules/user-guide/attachments/jsonschemas/${{ steps.get_version.outputs.VERSION }}/devfile.json
+      - name: Push to the devfile/docs repo
+        working-directory: docs-repo/
+        run: |
+          if [ "$(git status -s)" == "" ]
+          then
+            echo "Nothing to commit, Json schema didn't change"
+            exit 0
+          fi
+          
+          lastCommit="$(cd ../api-repo; git log -1 --format=%H)"
+          lastCommitterName="$(cd ../api-repo; git log -1 --format=%an)"
+          lastCommitterEmail="$(cd ../api-repo; git log -1 --format=%ae)"
+          
+          git config --global user.email "${lastCommitterEmail}"
+          git config --global user.name "${lastCommitterName}"
+          
+          git add docs/modules/user-guide/attachments/jsonschemas/stable/devfile.json docs/modules/user-guide/attachments/jsonschemas/${{ steps.get_version.outputs.VERSION }}/devfile.json 
+          git commit -m "Update devfile schema based on devfile/api@${lastCommit}"
+          git push "https://devfile-ci-robot:${{secrets.DOCS_UPDATE_SECRET}}@github.com/devfile/docs"
+          
--- a/.github/workflows/release-schema.yaml
+++ b/.github/workflows/release-schema.yaml
@@ -22,11 +22,36 @@ jobs:
       - name: Get the version being released
         id: get_version
         run: echo ::set-output name=version::$(cat api-repo/schemas/latest/jsonSchemaVersion.txt)
-      - name: Overwrite Stable Json Schema in Docs
+      - name: Overwrite Stable Json Schema in Docs if needed
         run: |
-          mkdir -p docs-repo/docs/modules/user-guide/attachments/jsonschemas/stable && \
-          cp -f api-repo/schemas/latest/devfile.json \
-            docs-repo/docs/modules/user-guide/attachments/jsonschemas/stable/devfile.json
+          if [ ! -f docs-repo/docs/modules/user-guide/attachments/jsonschemas/stable/jsonSchemaVersion.txt ]; then
+            # Stable version doesn't currently exist, so just copy over the schema we're releasing
+            mkdir -p docs-repo/docs/modules/user-guide/attachments/jsonschemas/stable
+            cp -f api-repo/schemas/latest/{devfile.json,jsonSchemaVersion.txt} \
+              docs-repo/docs/modules/user-guide/attachments/jsonschemas/stable/
+            exit 0
+          fi
+
+          # Parse the schema version that's being released
+          IFS='.' read -a semver <<< "${{ steps.get_version.outputs.VERSION }}"
+          MAJOR=${semver[0]}
+          MINOR=${semver[1]}
+          BUGFIX=${semver[2]}
+          
+          # Parse the version currently set to stable
+          stableVersion=`cat docs-repo/docs/modules/user-guide/attachments/jsonschemas/stable/jsonSchemaVersion.txt`
+          IFS='.' read -a stableSemVer <<< "$stableVersion"
+          stableMajor=${stableSemVer[0]}
+          stableMinor=${stableSemVer[1]}
+          stableBugfix=$(echo ${stableSemVer[2]} | awk -F '-' '{print $1}')
+
+          # Compare the two versions, only update stable if needed
+          if ((stableMajor <= MAJOR)) && ((stableMinor <= MINOR)) && ((stableBugfix <= BUGFIX)); then
+            cp -f api-repo/schemas/latest/{devfile.json,jsonSchemaVersion.txt} \
+              docs-repo/docs/modules/user-guide/attachments/jsonschemas/stable/
+          else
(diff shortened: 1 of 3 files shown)


.github/workflows/release-schema.yaml
```diff
@@ -0,0 +1,48 @@
...
+          path: docs-repo
+      - name: Checkout devfile api
+        uses: actions/checkout@v2
+        with:
+          path: api-repo
+      - name: Get the version being released
+        id: get_version
+        run: echo ::set-output name=version::$(cat api-repo/schemas/latest/jsonSchemaVersion.txt)
+      - name: Overwrite Stable Json Schema in Docs
+        run: mkdir -p docs-repo/docs/modules/user-guide/attachments/jsonschemas/stable && cp -f api-repo/schemas/latest/devfile.json docs-repo/docs/modules/user-guide/attachments/jsonschemas/stable/devfile.json
```
amisevsk, line 26: Minor readability
```suggestion
        run: |
          mkdir -p docs-repo/docs/modules/user-guide/attachments/jsonschemas/stable && \
          cp -f api-repo/schemas/latest/devfile.json \
            docs-repo/docs/modules/user-guide/attachments/jsonschemas/stable/devfile.json
```

.github/workflows/release-schema.yaml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Copy released Json Schema to Docs
+        run: mkdir -p docs-repo/docs/modules/user-guide/attachments/jsonschemas/${{ steps.get_version.outputs.VERSION }} && cp -f api-repo/schemas/latest/devfile.json docs-repo/docs/modules/user-guide/attachments/jsonschemas/${{ steps.get_version.outputs.VERSION }}/devfile.json
```
amisevsk, line 28: Minor readability
```suggestion
        run: |
          mkdir -p docs-repo/docs/modules/user-guide/attachments/jsonschemas/${{ steps.get_version.outputs.VERSION }} && \
          cp -f api-repo/schemas/latest/devfile.json \
            docs-repo/docs/modules/user-guide/attachments/jsonschemas/${{ steps.get_version.outputs.VERSION }}/devfile.json
```

make-release.sh
```diff
@@ -0,0 +1,109 @@
...
+if [[ -z "${GITHUB_TOKEN}" ]]; then
+  echo "[ERROR] The GITHUB_TOKEN environment variable must be set."
+  exit
+fi
+
+VERSION=$1
+API_VERSION=$2
+
+init() {
+  BRANCH=$(echo $VERSION | sed 's/.$/x/')
```
amisevsk, line 38: Not a huge issue now, but this `sed` replacement will fail if we ever get to e.g. `2.0.10`. Maybe something like `sed 's|[0-9]\+$|x|'` is safer?

make-release.sh
```diff
@@ -0,0 +1,109 @@
...
+  git reset --hard
+  git checkout $1
+  git fetch origin --prune
+  git pull origin $1
+}
+
+checkoutToReleaseBranch() {
+  echo "[INFO] Checking out to $BRANCH branch."
+  local branchExist=$(git ls-remote -q --heads | grep $BRANCH | wc -l)
+  if [[ $branchExist == 1 ]]; then
```
amisevsk, line 62: Maybe a cleaner way to do this:
```suggestion
  if git ls-remote -q --heads | grep -q $BRANCH ; then
```

make-release.sh
```diff
@@ -0,0 +1,109 @@
...
+    git push origin master:$BRANCH
+  fi
+  git checkout -B $VERSION
+}
+
+release() {
+  echo "[INFO] Releasing a new $VERSION version"
+
+  # replace nightly versions by release version
+  apply_sed "s#jsonschema:version=.*#jsonschema:version=${VERSION}#g" pkg/apis/workspaces/$API_VERSION/doc.go #src/constants.ts
```
amisevsk, line 77: Is the comment at the end of this line applicable to this repo, or a left-over from adapting the chectl release script?

make-release.sh
```diff
@@ -0,0 +1,109 @@
...
+commitChanges() {
+  echo "[INFO] Pushing changes to $VERSION branch"
+  git add -A
+  git commit -s -m "chore(release): release version ${VERSION}"
+  git push origin $VERSION
+}
+
+createReleaseBranch() {
+  echo "[INFO] Create the release branch based on $VERSION"
+  git push origin $VERSION -f
```
amisevsk, line 92: Seeing `git push -f` in a script worries me -- is force-pushing required for a specific case?

johnmcollier: Nope, just borrowed from the script I adapted this from. I can remove, shouldn't cause an issue 👍

make-release.sh
johnmcollier: Gah, thought I got all of these when I cleaned it up. It should be updated

make-release.sh
johnmcollier: Yeah, good idea.

davidfestal (approved):
Nothing more to comment than what @amisevsk pointed out.

As a side note, I'm just wondering what happens on the master branch after a release of the schemas (that happens in the release branch afaict). Is it bumped to the next alpha version that follows the last release version, or something like that ?
Do we want to define a process here as well ?

johnmcollier:
> As a side note, I'm just wondering what happens on the master branch after a release of the schemas (that happens in the release branch afaict). Is it bumped to the next alpha version that follows the last release version, or something like that ?
Do we want to define a process here as well ?

@davidfestal That's a good point, I think we've got a couple options here?
1) Keep master hardcoded to something like `next`( or `nightly`, etc) and only increment/set the version on the release branch
2) Have the `make-release.sh` open a PR on master that bumps the schema version to the next minor version

WDYT?

make-release.sh
johnmcollier: Removed the force push, it wasn't needed.

make-release.sh
johnmcollier: Updated.

make-release.sh
johnmcollier: I ended up not using sed for this, and took a separate approach that's a little cleaner I think.

johnmcollier:
I've updated this to include additional steps that opens a PR to bump the schema version on master (with a `-alpha` appended)

CC @davidfestal @elsony

make-release.sh
```diff
@@ -0,0 +1,139 @@
...
+set -u
+
+usage ()
+{   echo "Usage: ./make-release.sh <schema-version> <k8s-api-version>"
+    exit
+}
+
+if [[ $# -lt 2 ]]; then usage; fi
+
+VERSION=$1
```
elsony, line 25: Should we name `VERSION` to `SCHEMA_VERSION` and `API_VERSION` to `K8S_VERSION` to avoid confusion when reading the script?

make-release.sh
```diff
@@ -0,0 +1,139 @@
...
+API_VERSION=$2
+
+if ! command -v hub > /dev/null; then
+  echo "[ERROR] The hub CLI needs to be installed. See https://github.com/github/hub/releases"
+  exit
+fi
+if [[ -z "${GITHUB_TOKEN}" ]]; then
```
elsony, line 32: I checked the existing doc in https://github.com/devfile/api/blob/master/docs/proposals/versioning-and-release.md, we didn't update the instructions to set the GITHUB_TOKEN as part of the doc. Should we update it with that info so that anyone who needs to create a release in the future can just follow the instructions without hiccups?

elsony (changes requested):
This script automatically promotes the new version to stable and updating the minor version on the master branch. This script currently should be only be used on promoting code from master to a newer version. This may be sufficient for the time being but in case we need to do a service release on older versions and tries to run this script, we may end up moving stable to an earlier release. In order to avoid this potential problem, I wonder if we should:
1. add an extra check on the new version that we are trying to declare to make sure it is a version later than the existing stable.
2. if the new version that we try to set to is an earlier version than stable, we should probably prompt a warning as part of the script and saying we are trying to declare a release older than the existing stable release. The user can confirm it when running the script and it won't trigger the promotion to the stable nor updating the minor version on the master.

make-release.sh
johnmcollier: Yeah, that's a good point Elson, I've updated the script to be more clear in that regard.

make-release.sh
johnmcollier: Good point, updated the design doc 👍

johnmcollier:
@elsony Thanks for your review comments. I've updated the PR with the following changes:
- (release-schema.yml) Don't publish schema to stable if stable has a never version than what's being published
- (make-release.sh) Don't update master schema version if the master version is newer than what's being published
- Update design doc to clearly state that a Github token is required to run the script
- Clean up the release script for some clarity

.github/workflows/release-schema.yaml
```diff
@@ -22,11 +22,36 @@ jobs:
...
+          
+          # Parse the version currently set to stable
+          stableVersion=`cat docs-repo/docs/modules/user-guide/attachments/jsonschemas/stable/jsonSchemaVersion.txt`
+          IFS='.' read -a stableSemVer <<< "$stableVersion"
+          stableMajor=${stableSemVer[0]}
+          stableMinor=${stableSemVer[1]}
+          stableBugfix=$(echo ${stableSemVer[2]} | awk -F '-' '{print $1}')
+
+          # Compare the two versions, only update stable if needed
+          if ((stableMajor <= MAJOR)) && ((stableMinor <= MINOR)) && ((stableBugfix <= BUGFIX)); then
```
elsony, line 49: For the exact same number, I can see that we don't want to update the minor version in the github action. However, I wonder if we still want to update the schema to the latest on stable in that case, i.e. should this condition be
`if ((stableMajor <= MAJOR)) && ((stableMinor <= MINOR)) && ((stableBugfix < BUGFIX)); then`  (Note: leave the github version check unchanged) ?

elsony (approved):
/lgtm

amisevsk (approved):

Issue #288, opened 2021-01-05
Endpoint Name should also follow DNS1123 rule to have Maxlength=63 chars


yangcao77:
The schema limits endpoint name with Maxlength=15 is because the pod container port name has that limitation

Since we want to use `<portNumber>-<protocol>` as pod container port name (https://github.com/devfile/api/issues/287), the Endpoint Name won't have that 15 chars limitation; it should follow DNS1123 rule, to be with maxlength=63

Issue #287, opened 2021-01-05
use <portNumber>-<protocol> as pod container port name in devfile library generator


yangcao77:
The generator currently is using the devfile Endpoint name as the pod container port name. However, the pod container port name has a 15 characters limit, and as a result, Endpoint name is limited to be less than 15 characters. 

Since Odo now wants to remove that 15 characters truncating for URL Name:https://github.com/openshift/odo/issues/4060, I rechecked the 15 chars limitation only applies to pod container port name.

We can apply how Odo deals with the port name for S2I projects, use `<portNumber>-<protocol>` as the name :https://github.com/openshift/odo/blob/cdfa5450201702406feefe3e4de3cc3f5ee494be/pkg/util/util.go#L575-L578

```
		port := corev1.ContainerPort{
			Name:          fmt.Sprintf("%d-%s", portNumber, strings.ToLower(string(portProto))),
			ContainerPort: portNumber,
			Protocol:      portProto,
			}
```

elsony:
Blocked by #288

amisevsk:
Theia requires multiple routes that lead to the same port for security reasons. How would this change accomodate a devfile such as
```yaml
        container:
          image: "quay.io/eclipse/che-theia:next"
          endpoints:
            - name: "theia"
              exposure: public
              targetPort: 3100
              protocol: http
            - name: "webviews"
              exposure: public
              targetPort: 3100
              protocol: http
```
If my reading of the proposal is correct, this will result in two containerPorts named `"3100-http"`

mik-dass:
Can we close this issue?

yangcao77:
Yes. Forgot to do that.

PR #289, opened 2021-01-05
change endpoint name limit to 63 characters

yangcao77:

### What does this PR do?
This PR changes the length limit of Endpoint Name from 15 to 63 to match DNS1123 and be consistent with the other K8s resources defined in devfile.

the length limit was defined to be 15 was because the container port name defined in pod has to be less than 15 characters
since library is not going to use the endpoint name as the port name(https://github.com/devfile/api/issues/287), the length limit should match the other K8s resources

#288

### DIFF
--- a/crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
+++ b/crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
@@ -4539,7 +4539,7 @@ spec:
                                     - none
                                     type: string
                                   name:
-                                    maxLength: 15
+                                    maxLength: 63
                                     pattern: ^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
                                     type: string
                                   path:
@@ -4708,7 +4708,7 @@ spec:
                                     - none
                                     type: string
                                   name:
-                                    maxLength: 15
+                                    maxLength: 63
                                     pattern: ^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
                                     type: string
                                   path:
@@ -4817,7 +4817,7 @@ spec:
                                     - none
                                     type: string
                                   name:
-                                    maxLength: 15
+                                    maxLength: 63
                                     pattern: ^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
                                     type: string
                                   path:
@@ -5266,7 +5266,7 @@ spec:
                                               - none
                                               type: string
                                             name:
-                                              maxLength: 15
+                                              maxLength: 63
                                               pattern: ^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
                                               type: string
                                             path:
@@ -5425,7 +5425,7 @@ spec:
                                               - none
                                               type: string
                                             name:
-                                              maxLength: 15
+                                              maxLength: 63
                                               pattern: ^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
                                               type: string
                                             path:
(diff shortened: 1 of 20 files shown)


sleshchenko (commented):
I don't have a strong opinion if we should keep 15 or increase it to 63 but another reason why endpoint name mayb have limitation in 15 chars, an implementation may want to use it for the host as part of a subdomain, like `$instanceId-endpointName.$ingressdomain` -> `http://workspacec89506a320d9495d-theia-3100.192.168.39.170.nip.io` <- it's from devworkspace operator, but we should be able to use a dedicated subdomain there, like `http://workspacec89506a320d9495d.theia-3100.192.168.39.170.nip.io`

It's even worse when an implementation does not have permissions to use customHosts and would like to have a meaningful host, then when route is named with endpoint-name, it will get `$endpoint_name-$namespace.$router_host` assigned host.

amisevsk (approved):
Personally I prefer the 15 character limit, since that's the limit on the analogous k8s object (a named container endpoint should map to a named container port if possible). From reading https://github.com/openshift/odo/issues/4060, I understand the motivation, but also don't see the direct link to the changes in this PR, especially in light of https://github.com/devfile/api/issues/287.

From the DevWorkspace operator perspective, increasing this limit has a minor impact on readability; with the 15 char limit, I could have a port named `database-conn` which could be translated to a similarly named container port. With the 63-char limit, all container ports will have to have less intelligible names (e.g. `9000-tcp`) as there's no surefire way to truncate. 

Either way it's not hard to work around, so I'm approving the PR. I totally see how having a single name field that's shorter than the rest could be confusing (as it is for container ports on k8s!)

yangcao77:
since odo is going to increase the URL name limit to 63 chars, odo writes the URL info into devfile Endpoints entry, which has conflict with the schema rule. @kadel 

#287 is just to truncate the container port name inside pod, is it going to affect DevWorkspace or console?  Is DevWorkspace/console going to read generated pod spec and get the container ports?  Note that`devfile.Endpoint` still contains values defined in devfile. Also note that for each devfile endpoint, it does not have to has an exact match of container port, since there can have multiple endpoints sharing the same port number, but the container port & service should only have one defined.

@sleshchenko I can see your point of having 15 chars limit is "safer" in your case, since you are going to include $instanceId and $ingressdomain within the host, but host also has limitation of 63 characters. limiting the endpoint name to be less than 15 characters does not completely resolve that issue, it is still possible that the host length will be >63 chars; which means Devworkspace will still need to verify the length before actually use it.

amisevsk:
> since odo is going to increase the URL name limit to 63 chars, odo writes the URL info into devfile Endpoints entry, which has conflict with the schema rule.
>
> #287 is just to truncate the container port name inside pod, is it going to affect DevWorkspace or console? Is DevWorkspace/console going to read generated pod spec and get the container ports? Note thatdevfile.Endpoint still contains values defined in devfile. Also note that for each devfile endpoint, it does not have to has an exact match of container port, since there can have multiple endpoints sharing the same port number, but the container port & service should only have one defined.

I'm not sure I understand what you mean, but DevWorkspace operator is affected by this change, as container endpoints directly map to containerPorts in the Deployment associated with the DevWorkspace. 

> there can have multiple endpoints sharing the same port number

This is news to me -- it seems like there's some inconsistency there for container components that do not specify `dedicatedPod: true`. Containers in a pod cannot share ports, since they're mapped to a virtual localhost network. 

Also, I ran into another issue today while working on something unrelated -- the proposed way of naming containerPorts in `odo` will result in devfiles with e.g. 
```yaml
          endpoints:
            - name: "theia"
              exposure: public
              targetPort: 3100
            - name: "webviews"
              exposure: public
              targetPort: 3100
```
resolving to invalid k8s specs.

yangcao77:
confirmed with @maysunfaisal that Console is not consuming container port name. #288 and #287 won't affect Console's behavior, merging this PR

PR #301, opened 2021-01-13
Add devfile validation

### LINKED ISSUE #151: Consolidate dev workspace & odo validations
Following the Devfile [discussion](https://docs.google.com/document/d/1Zn6D1gNu1gDidtillni0hNAeU5pscfqwhx4mv3MmmwY/edit#), it was decided that the devfile validation would reside in devfile/api since the schema and go structs reside in devfile/api.

This issue tracks the work where we need to consolidate the validations for devfile present in dev workspace and odo repository. 

The tools' specific validations will remain in their respective repository and only the validations wrt to the devfile spec will be moved to devfile/api.

maysunfaisal:
### What does this PR do?
Implement the consolidated validations documented in https://github.com/devfile/api/blob/master/pkg/validation/validation-rule.md

Fixes #151 

Yes, new go tests

### DIFF
--- a/pkg/apis/workspaces/v1alpha2/keyed_implementations.go
+++ b/pkg/apis/workspaces/v1alpha2/keyed_implementations.go
@@ -1,6 +1,7 @@
 package v1alpha2
 
 import (
+	"fmt"
 	"reflect"
 )
 
@@ -17,4 +18,24 @@ func extractKeys(keyedList interface{}) []Keyed {
 		}
 	}
 	return keys
+}
+
+// CheckDuplicateKeys checks if duplicate keys are present in the devfile objects
+func CheckDuplicateKeys(keyedList interface{}) error {
+	seen := map[string]bool{}
+	value := reflect.ValueOf(keyedList)
+	for i := 0; i < value.Len(); i++ {
+		elem := value.Index(i)
+		if elem.CanInterface() {
+			i := elem.Interface()
+			if keyed, ok := i.(Keyed); ok {
+				key := keyed.Key()
+				if seen[key] {
+					return fmt.Errorf("duplicate key: %s", key)
+				}
+				seen[key] = true
+			}
+		}
+	}
+	return nil
 }
--- a/pkg/validation/commands.go
+++ b/pkg/validation/commands.go
@@ -20,11 +20,13 @@ func ValidateCommands(commands []v1alpha2.Command, components []v1alpha2.Compone
 	for _, command := range commands {
 		// Check if the command is in the list of already processed commands
 		// If there's a hit, it means more than one command share the same ID and we should error out
-		commandID := strings.ToLower(command.Id)
-		if _, exists := processedCommands[commandID]; exists {
+		if isInt(command.Id) {
+			return &InvalidNameOrIdError{id: command.Id, resourceType: "command"}
+		}
+		if _, exists := processedCommands[command.Id]; exists {
 			return &InvalidCommandError{commandId: command.Id, reason: "duplicate commands present with the same id"}
 		}
-		processedCommands[commandID] = commandID
+		processedCommands[command.Id] = command.Id
 
 		parentCommands := make(map[string]string)
 		err = validateCommand(command, parentCommands, commandMap, components)
@@ -62,11 +64,11 @@ func validateCommand(command v1alpha2.Command, parentCommands map[string]string,
 		return validateCommandComponent(command, components)
 	case command.VscodeLaunch != nil:
 		if command.VscodeLaunch.Uri != "" {
-			return validateURI(command.VscodeLaunch.Uri)
+			return ValidateURI(command.VscodeLaunch.Uri)
 		}
 	case command.VscodeTask != nil:
 		if command.VscodeTask.Uri != "" {
-			return validateURI(command.VscodeTask.Uri)
+			return ValidateURI(command.VscodeTask.Uri)
 		}
 	default:
 		err = fmt.Errorf("command %s type is invalid", command.Id)
--- a/pkg/validation/commands.go
+++ b/pkg/validation/commands.go
@@ -1,0 +1,117 @@
+package validation
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
+)
+
+// ValidateCommands validates the devfile commands:
+// 1. if there are commands with duplicate IDs, an error is returned
+// 2. checks if its either a valid exec or composite command
+func ValidateCommands(commands []v1alpha2.Command, components []v1alpha2.Component) (err error) {
+	processedCommands := make(map[string]string, len(commands))
+	commandMap := getCommandsMap(commands)
+
+	for _, command := range commands {
+		// Check if the command is in the list of already processed commands
+		// If there's a hit, it means more than one command share the same ID and we should error out
+		commandID := strings.ToLower(command.Id)
+		if _, exists := processedCommands[commandID]; exists {
+			return &InvalidCommandError{commandId: command.Id, reason: "duplicate commands present with the same id"}
+		}
+		processedCommands[commandID] = commandID
+
+		err = validateCommand(command, commandMap, components)
+		if err != nil {
+			return err
+		}
+	}
+
+	return
+}
+
+// validateCommand validates a given devfile command
+func validateCommand(command v1alpha2.Command, devfileCommands map[string]v1alpha2.Command, components []v1alpha2.Component) (err error) {
+
+	// If the command is a composite command, need to validate that it is valid
+	if command.Composite != nil {
+		parentCommands := make(map[string]string)
+		return validateCompositeCommand(&command, parentCommands, devfileCommands, components)
+	}
+
+	return validateExecCommand(command, components)
+}
+
+// validateExecCommand validates the given exec command, the command should:
+// 1. have a component
+// 2. have a command line
+// 3. map to a valid container component
+func validateExecCommand(command v1alpha2.Command, components []v1alpha2.Component) (err error) {
+
+	if command.Exec == nil {
+		return &InvalidCommandError{commandId: command.Id, reason: "should be of type exec"}
+	}
+
+	// must map to a container component
+	isComponentValid := false
+	for _, component := range components {
+		if component.Container != nil && command.Exec.Component == component.Name {
+			isComponentValid = true
+		}
+	}
+	if !isComponentValid {
+		return &InvalidCommandError{commandId: command.Id, reason: "command does not map to a container component"}
+	}
+
+	return
+}
+
+// validateCompositeCommand checks that the specified composite command is valid. The command:
+// 1. should not reference itself via s subcommand
+// 2. should not indirectly reference itself via a subcommand which is a composite command
+// 3. should reference a valid devfile command
+// 4. should have a valid exec sub command
+func validateCompositeCommand(command *v1alpha2.Command, parentCommands map[string]string, devfileCommands map[string]v1alpha2.Command, components []v1alpha2.Component) error {
+
+	// Store the command ID in a map of parent commands
+	parentCommands[command.Id] = command.Id
+
+	if command.Composite == nil {
+		return &InvalidCommandError{commandId: command.Id, reason: "should be of type composite"}
+	}
+
+	// Loop over the commands and validate that each command points to a command that's in the devfile
+	for _, cmd := range command.Composite.Commands {
+		if strings.ToLower(cmd) == command.Id {
+			return &InvalidCommandError{commandId: command.Id, reason: "composite command cannot reference itself"}
+		}
+
+		// Don't allow commands to indirectly reference themselves, so check if the command equals any of the parent commands in the command tree
+		_, ok := parentCommands[strings.ToLower(cmd)]
+		if ok {
+			return &InvalidCommandError{commandId: command.Id, reason: "composite command cannot indirectly reference itself"}
+		}
+
+		subCommand, ok := devfileCommands[strings.ToLower(cmd)]
+		if !ok {
+			return &InvalidCommandError{commandId: command.Id, reason: fmt.Sprintf("the command %q mentioned in the composite command does not exist in the devfile", cmd)}
+		}
+
+		if subCommand.Composite != nil {
+			// Recursively validate the composite subcommand
+			err := validateCompositeCommand(&subCommand, parentCommands, devfileCommands, components)
+			if err != nil {
+				// Don't wrap the error message here to make the error message more readable to the user
+				return err
+			}
+		} else {
+			err := validateExecCommand(subCommand, components)
+			if err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
--- a/pkg/validation/commands.go
+++ b/pkg/validation/commands.go
@@ -2,6 +2,7 @@ package validation
 
 import (
 	"fmt"
+	"reflect"
 	"strings"
 
 	"github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
@@ -9,9 +10,11 @@ import (
 
 // ValidateCommands validates the devfile commands:
 // 1. if there are commands with duplicate IDs, an error is returned
-// 2. checks if its either a valid exec or composite command
+// 2. checks if its either a valid command
+// 3. checks if commands belonging to a specific group obeys the rule of 1 default command
 func ValidateCommands(commands []v1alpha2.Command, components []v1alpha2.Component) (err error) {
 	processedCommands := make(map[string]string, len(commands))
+	groupCommandMap := make(map[v1alpha2.CommandGroup][]v1alpha2.Command)
 	commandMap := getCommandsMap(commands)
 
 	for _, command := range commands {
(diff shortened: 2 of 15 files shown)


pkg/validation/events.go
```diff
@@ -0,0 +1,136 @@
...
+// 1. event should map to a valid devfile command
+// 2. preStart and postStop events should either map to an apply command or a composite command with apply commands
+// 3. postStart and preStop events should either map to an exec command or a composite command with exec commands
+func isEventValid(eventNames []string, eventType string, commandMap map[string]v1alpha2.Command) error {
+	var invalidCommand, applyEvents, execEvents []string
+
+	for _, eventName := range eventNames {
+		if command, ok := commandMap[strings.ToLower(eventName)]; !ok { // check if event is in the list of devfile commands
+			invalidCommand = append(invalidCommand, eventName)
+		} else {
```
amisevsk, line 64: We could reduce nesting by adding a `continue` and removing the `else` case:
```golang
if command, ok := commandMap[strings.ToLower(eventName)]; !ok { // check if event is in the list of devfile commands
	invalidCommand = append(invalidCommand, eventName)
	continue
}
switch eventType {
```

pkg/validation/events.go
```diff
@@ -0,0 +1,136 @@
...
+		eventErrors = fmt.Sprintf("\n%s does not map to a valid devfile command", strings.Join(invalidCommand, ", "))
+	}
+
+	if len(applyEvents) > 0 {
+		eventErrors += fmt.Sprintf("\n%s should either map to an apply command or a composite command with apply commands", strings.Join(applyEvents, ", "))
+	}
+
+	if len(execEvents) > 0 {
+		eventErrors += fmt.Sprintf("\n%s should either map to an exec command or a composite command with exec commands", strings.Join(execEvents, ", "))
+	}
```
amisevsk, line 101: `applyEvents` and `execEvents` is unclear in this case, consider renaming to indicate that these are lists of apply/exec events added in error

pkg/validation/projects.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"fmt"
+
+	"github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
+)
+
+// ValidateStarterProjects checks if starter project has only one remote configured
+// and if the checkout remote matches the renote configured
+func ValidateStarterProjects(starterProject v1alpha2.StarterProject) (err error) {
```
amisevsk, line 11: I think using a named return here is less clear than returning `nil` explicitly; the created value (`err`) is never referenced.

pkg/validation/projects.go
```diff
@@ -0,0 +1,59 @@
...
+	if starterProject.Git != nil {
+		gitSource = starterProject.Git.GitLikeProjectSource
+	} else if starterProject.Github != nil {
+		gitSource = starterProject.Github.GitLikeProjectSource
+	} else {
+		return
+	}
+
+	if len(gitSource.Remotes) != 1 {
+		return fmt.Errorf("starterProject can have only one remote")
```
amisevsk, line 23: Consider splitting this into a zero remotes and a 2+ remotes case.

pkg/validation/projects.go
```diff
@@ -0,0 +1,59 @@
...
+	} else if gitSource.CheckoutFrom.Remote != "" {
+		return validateRemoteMap(gitSource.Remotes, gitSource.CheckoutFrom.Remote)
+	}
+
+	return
+}
+
+// ValidateProjects checks if the project has more than one remote configured then a checkout
+// remote is mandatory and if the checkout remote matches the renote configured
+func ValidateProjects(project v1alpha2.Project) (err error) {
```
amisevsk, line 33: As above, the named return value is unused.

pkg/validation/validation-rule.md
```diff
@@ -56,3 +56,8 @@ Common rules for all components types:
 
 ### starterProjects:
 - Starter project entries cannot have more than one remote defined
+- if checkout remote is mentioned, validate it against the starter project remote configured map
+
+### projects
+- if more than one remote map is configured, a checkout remote is mandatory
```
amisevsk, line 62:
```suggestion
- if more than one remote is configured, a checkout remote is mandatory
```

pkg/validation/projects.go
```diff
@@ -0,0 +1,59 @@
...
+
+	if len(gitSource.Remotes) > 1 || gitSource.CheckoutFrom.Remote != "" {
+		return validateRemoteMap(gitSource.Remotes, gitSource.CheckoutFrom.Remote)
+	}
+
+	return
+}
+
+// validateRemoteMap checks if the checkout remote is present in the project remote map
+func validateRemoteMap(remotes map[string]string, checkoutRemote string) (err error) {
```
amisevsk, line 52: Unnecessary named return

pkg/validation/commands.go
```diff
@@ -0,0 +1,195 @@
...
+// 2. checks if its either a valid command
+// 3. checks if commands belonging to a specific group obeys the rule of 1 default command
+func ValidateCommands(commands []v1alpha2.Command, components []v1alpha2.Component) (err error) {
+	processedCommands := make(map[string]string, len(commands))
+	groupCommandMap := make(map[v1alpha2.CommandGroup][]v1alpha2.Command)
+	commandMap := getCommandsMap(commands)
+
+	for _, command := range commands {
+		// Check if command id is all numeric
+		if isInt(command.Id) {
```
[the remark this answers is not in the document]
maysunfaisal: @amisevsk wondering if regex is reqd, since `strconv.Atoi(str)` solves the purpose? 🤔 I'm down to either

pkg/validation/commands.go
```diff
@@ -0,0 +1,195 @@
...
+		}
+		processedCommands[command.Id] = command.Id
+
+		parentCommands := make(map[string]string)
+		err = validateCommand(command, parentCommands, commandMap, components)
+		if err != nil {
+			return err
+		}
+
+		commandGroup := *getGroup(command)
```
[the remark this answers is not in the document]
maysunfaisal: ah good find, thx

pkg/validation/commands.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+	if !isComponentValid {
+		return &InvalidCommandError{commandId: command.Id, reason: "command does not map to a container component"}
+	}
+
+	return
+}
+
+// validateCompositeCommand checks that the specified composite command is valid. The command:
+// 1. should not reference itself via s subcommand
```
johnmcollier, line 159:
```suggestion
// 1. should not reference itself via a subcommand
```

pkg/validation/commands.go
```diff
@@ -0,0 +1,195 @@
...
+		if ok {
+			return &InvalidCommandError{commandId: command.Id, reason: "composite command cannot indirectly reference itself"}
+		}
+
+		subCommand, ok := devfileCommands[strings.ToLower(cmd)]
+		if !ok {
+			return &InvalidCommandError{commandId: command.Id, reason: fmt.Sprintf("the command %q mentioned in the composite command does not exist in the devfile", cmd)}
+		}
+
+		err := validateCommand(subCommand, parentCommands, devfileCommands, components)
```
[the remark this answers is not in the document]
maysunfaisal: @amisevsk I am not sure I follow the comment here, are you saying this will fail?

exec1
exec2
composite
 - exec1
 - exec2

pkg/validation/projects.go
maysunfaisal: The schema says this is mandatory, so that will have validated the zero case already 🤔

pkg/validation/endpoints.go
```diff
@@ -0,0 +1,33 @@
...
+		}
+
+		if _, ok := processedEndPointName[endPoint.Name]; ok {
+			return &InvalidEndpointError{name: endPoint.Name}
+		}
+		processedEndPointName[endPoint.Name] = true
+		currentComponentEndPointPort[endPoint.TargetPort] = true
+	}
+
+	for targetPort := range currentComponentEndPointPort {
```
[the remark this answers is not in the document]
maysunfaisal: because endpoints within a component can have the same port but across components cannot.

Valid:
component1
- endpoint1 8080
- endpoint2 8080

Invalid:
component1
-endpoint1 8080
component2
-endpoint2 8080

So by checking outside the endpoint loop, we are effectively cross checking against other components as well

pkg/validation/commands.go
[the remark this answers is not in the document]
amisevsk: It just seem strange to enforce a format on a string by whether or not Go can parse a number from it. The regex you're checking is `[a-zA-Z]`. `strconv.Atoi(str)` will fail for other reasons than that, and it's an additional step in reasoning that's required. 

I won't push too hard on it, but it's a strange approach IMO.

pkg/validation/commands.go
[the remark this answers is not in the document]
amisevsk: From my read, something like
```yaml
commands:
  - name: composite-1
    composite:
      commands: ["composite-a", "composite-b"]
  - name: composite-a
    composite:
      commands: ["plain-command"]
  - name: composite-b
    composite:
      commands: ["plain-command"]
  - name: plain-command
    exec:
      commandLine: "echo 'hello world'"
```
will be an issue.

pkg/validation/endpoints.go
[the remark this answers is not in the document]
amisevsk: Ahh I didn't see that -- good point. Thanks!

pkg/validation/projects.go
amisevsk: In that case I think it'd be clearer if the check was `> 1`, but this is so minor that it's up to you if you want to change it.

pkg/validation/commands.go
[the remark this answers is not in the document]
maysunfaisal: @amisevsk if we're validating `composite-1` in this scenario, it is perfectly fine IMO, even tho its sub-commands ultimately reference the same exec command `plain-command`

I wrote a test case depicting this scenario and it seems fine!

pkg/validation/events_test.go
```diff
@@ -0,0 +1,187 @@
...
+		{
+			name:      "Case 6: Invalid preStop events - Non-Exec and Composite Exec Commands",
+			eventType: preStop,
+			eventNames: []string{
+				"exec1",
+				"apply2",
+				"compositeOnlyExec",
+			},
+			wantErr: true,
+		},
```
amisevsk, line 162: These tests are unclear to me; could we rename the tests to be more precise in what the expected problem is? E.g.
```
Case 5: Invalid postStart events - Composite command contains non-exec command
Case 6: Invalid preStop events - Non-exec command in events
```

pkg/validation/projects_test.go
```diff
@@ -0,0 +1,187 @@
...
+				generateDummyGitStarterProject("project1", "origin", map[string]string{"origin": "originremote"}),
+				generateDummyGitStarterProject("project2", "origin", map[string]string{"origin": "originremote2"}),
+			},
+			wantErr: false,
+		},
+		{
+			name: "Case 2: Invalid Starter Project",
+			starterProjects: []v1alpha2.StarterProject{
+				generateDummyGithubStarterProject("project1", "origin", map[string]string{"origin": "originremote", "test": "testremote"}),
+				generateDummyGithubStarterProject("project2", "origin", map[string]string{"origin": "originremote", "test": "testremote"}),
```
amisevsk, line 92: This test would be clearer if there was only one starter project.

pkg/validation/utils_test.go
```diff
@@ -0,0 +1,71 @@
...
+		{
+			name:       "Case 2: alphanumeric string",
+			arg:        "1234abc",
+			wantResult: false,
+		},
+		{
+			name:       "Case 3: string with numbers and character",
+			arg:        "12_34",
+			wantResult: false,
+		},
```
amisevsk, line 28: Should add a testcase for `arg: "0xff"` to catch in case someone tries to use `ParseInt` instead of `Atoi`. Note also that underscores in IDs are not allowed by regex.

amisevsk (commented):
Generally looks good, a few comments regarding test cases.

pkg/validation/utils.go
```diff
@@ -21,11 +21,9 @@ func getCommandsMap(commands []v1alpha2.Command) map[string]v1alpha2.Command {
 }
 
 // isInt returns true if the string is an integer, false otherwise
-func isInt(str string) bool {
-	if _, err := strconv.Atoi(str); err == nil {
-		return true
-	}
-	return false
+func isInt(str string) (bool, error) {
+	match, err := regexp.MatchString("^[0-9]+$", str)
```
amisevsk, line 25: Two minor issues here:
1. Regex won't match `10-10`; if you want to check that at least one letter exists, just check `"[a-zA-Z]"`
2. Avoid returning an error by using 
    ```
    var alphaRegexp = regexp.MustCompile(`[a-zA-Z]`)
    ```
    at the top of the file.

pkg/validation/events.go
```diff
@@ -0,0 +1,138 @@
...
+		command, ok := commandMap[strings.ToLower(eventName)]
+		if !ok { // check if event is in the list of devfile commands
+			invalidCommand = append(invalidCommand, eventName)
+			continue
+		}
+
+		switch eventType {
+		case preStart:
+			fallthrough
+		case postStop:
```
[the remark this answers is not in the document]
maysunfaisal: TIL!

pkg/apis/workspaces/v1alpha2/keyed_implementations.go
```diff
@@ -18,3 +19,23 @@ func extractKeys(keyedList interface{}) []Keyed {
 	}
 	return keys
 }
+
+// CheckDuplicateKeys checks if duplicate keys are present in the devfile objects
+func CheckDuplicateKeys(keyedList interface{}) error {
```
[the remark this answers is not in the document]
maysunfaisal: @amisevsk Because I cannot pass a component/command struct to an interface of type `Keyed` but I can with `interface{}`. The other implementation func `extractKeys(keyedList interface{})` does the same thing. Am I missing something? 🤔

pkg/validation/commands.go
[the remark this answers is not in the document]
amisevsk: It took me a few tries but I managed to catch the issue I'm describing. Basically, since we're using a flat map to store "visited" command IDs, a command ID that has been processed as part of another branch can cause a false-positive. For example, the references below will fail despite not having any back refs:

self-ref

I pushed a commit with a reproducer [here](https://github.com/amisevsk/devworkspace-api/commit/92d1b19119c61ed44c5a572ccffc21ac1d194056#diff-0321478ca2e7301afce4120f6c3e19ebb5d6401b109284a6cf700dcb1d49476eR1-R19). Excuse the other test cases -- I wanted to make sure my reasoning was right

pkg/apis/workspaces/v1alpha2/keyed_implementations.go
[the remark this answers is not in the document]
amisevsk: Ah you're right, I forgot that you can't cast interfaces like that for Go reasons. Sorry about the confusion!

pkg/validation/commands.go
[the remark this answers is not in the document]
maysunfaisal: Thanks 🙌 

I was able to solve it by deleting the map entry when one composite command is done being validated.

pkg/validation/utils.go
maysunfaisal: +1 on point 2, thx

On point 1, I just chatted with Stephanie who wrote the initial doc for validation and it seems to her `10-10` would be a valid id. I just tried it out and it looks Kube seems to be happy with it. 
Looks like we only want to restrict `1234`. Do you agree or have a counter argument?

amisevsk: If `10-10` is valid but `1234` is not, I wonder where the restriction is coming from. For what it's worth, `1234` is a valid identifier in k8s as well (you have to quote it in yaml otherwise it's interpreted as an integer).

The only checks for "not all numbers" I know of are
1. Places that use RFC1035 DNS labels (must start with an alphabetic character)
2. Port names, for which `10-10` is [invalid](https://github.com/kubernetes/apimachinery/blob/ee96543cc6730a24af1c7323dac47e52c340ba39/pkg/util/validation/validation.go#L317-L345)

In this case, what is the `isInt` check being used for specifically? I tried applying a DevWorkspace with ids like `"1234"` littered throughout and I didn't run into any issues -- the DevWorkspace operator correctly named a container `1234`.

maysunfaisal: Actually I wasnt able to create a svc with `1234`/`12-34`. Not sure what resource the DevWorkspace was creating 🤔. I even quoted it in yaml :
```
The Service "1234" is invalid: metadata.name: Invalid value: "1234": a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')
```

But anyhow, I think we can remove this restriction. I confirmed and it seems like port name cannot take either `1234`/`12-34` like you suggested and linked. And the library generator is [already handling this](https://github.com/devfile/library/blob/741af7b5cad69d4962c8ea1b374d9c80014eb6b9/pkg/devfile/generator/utils.go#L46) for container port names from devfile. 

Command IDs are not representative of any Kube resource and this just leaves us with Component name. I was able to create a pod with container named `"1234"`/`"12-34"`.

WDYT?

amisevsk: Makes sense to me, it seems like `isInt` is unnecessary (I believe the `1234` svc name is covered by in-schema regexes)

maysunfaisal: Updated! I have also addressed the remaining unresolved conversations.

maysunfaisal:
@amisevsk Any updates on the addressed commits? Thx!

pkg/validation/components.go
```diff
@@ -0,0 +1,117 @@
...
+			}
+		}
+
+	}
+
+	// Check if the volume mounts mentioned in the containers are referenced by a volume component
+	var invalidVolumeMountsErr string
+	for componentName, volumeMountNames := range processedVolumeMounts {
+		for _, volumeMountName := range volumeMountNames {
+			if _, ok := processedVolumes[volumeMountName]; !ok {
```
amisevsk, line 106: Minor nit, but this could just be `if !processedVolumes[volumeMountName] {`

pkg/validation/components_test.go
```diff
@@ -0,0 +1,253 @@
...
+			Value: "/some/path/",
+		},
+	}
+
+	endpointUrl18080 := generateDummyEndpoint("url1", 8080)
+	endpointUrl18081 := generateDummyEndpoint("url1", 8081)
+	endpointUrl28080 := generateDummyEndpoint("url2", 8080)
+
+	invalidVolMountErr := ".*\nvolume mount myinvalidvol belonging to the container component.*\nvolume mount myinvalidvol2 belonging to the container component.*"
+	duplicateCmdErr := "duplicate key: component1"
```
amisevsk, line 133: Maybe a copy-paste issue, but `Cmd` does not make sense in this context (maybe `duplicateComponentErr`?)

amisevsk (approved):
Overall LGTM, a few leftover minor nits.

maysunfaisal:
@amisevsk thanks for those nit reviews! 🙌  I've addressed them!

amisevsk (approved):
Approving again to say LGTM! Thanks for your patience @maysunfaisal

maysunfaisal:
@amisevsk no thank you for the thorough review! Appreciate it.

PR #307, opened 2021-01-15
Update endpoint tests

maysunfaisal:

### What does this PR do?
Update the CI tests as they were not rebased before merging

yes

### DIFF
--- a/test/v200/devfiles/components/containerEndpointLongName.yaml
+++ b/test/v200/devfiles/components/containerEndpointLongName.yaml
@@ -2,5 +2,5 @@
     container:
       image: testImage
       endpoints:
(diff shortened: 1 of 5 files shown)


mmulholla:
LGTM

amisevsk (approved):
If the purpose of these tests is to detect incompatible changes, this should not have triggered a failure in the first place -- all devfiles with endpoint names `<=15` chars are valid if the check is `<=63` chars.

maysunfaisal:
@amisevsk true but the tests were failing because the expected outcome was different. The tests were looking for <=15 when master was <=63, this was because the test PR was not rebased with master before merging and got out of sync

PR #319, opened 2021-01-20
Extend volumes

### LINKED ISSUE #310: Extending Volume components to support more features
## Is your feature request related to a problem? Please describe.
Currently, Volume components in a devfile only support a size field. It would be useful to be able to specify more information about what sort of volume is needed:

- Specify that a volume does not require persistent storage (e.g. can be provisioned as an emptyDir volume) -- such volumes are useful for sharing data between containers
- Specify that a volume should exist already (i.e. don't create a volume from this file) -- this would be useful for specifying e.g. a shared ROX assets volume that can be mounted to multiple workspaces
- Specify an external volume's type (e.g. link to a secret that exists on the cluster; link to a configmap on the cluster) -- this would allow mounting secrets/configmaps that already exist into new workspaces.

## Describe the solution you'd like
Introduce additional fields on volume components:
- `persistent` - whether persistent volume is expected to be available after workspace is stopped (default: `true`)
- `external` - defines volume as external to the current workspace (i.e. its lifecycle is separate)
- `external.type` - type of external volume, e.g. "secret", "configmap", or "storage"
- `external.name` - name of existing resource on cluster

Something like
```yaml
spec:
  components:
    - name: my-ephemeral-volume 
      volume:
        persistent: false # Volume my-ephemeral-volume will be mounted as emptyDir

    - name: my-external-volume
      volume:
        external: # Setting fields indicates that this is an external volume
          type: storage
          name: my-pvc
            
    - name: my-existing-secret
      volume:
        external:
          type: secret # or configmap
          name: my-secret
```

## Describe alternatives you've considered
Suggestions are more than welcome

amisevsk:
### What does this PR do?
Implements extensions to Volume components according to proposal in https://github.com/devfile/api/issues/310. Not a final design; I'm hoping for more discussion before merge. I've added this topic to the upcoming call to discuss more widely.

Closes #310 

N/A?


pkg/apis/workspaces/v1alpha2/volumeComponent.go
```diff
@@ -11,4 +11,44 @@ type Volume struct {
...
+	// kubebuilder:validation:Enum="storage,configmap,secret"
+	Type ExistingVolumeType `json:"type"`
+}
+
+// ExistingVolumeType defines the type of an external Volume
+type ExistingVolumeType string
+
+const (
+	// StorageVolumeType specifies persistent storage, e.g. a PersistentVolumeClaim
+	StorageVolumeType ExistingVolumeType = "storage"
```
sleshchenko, line 49: I wonder what is drawback if we name it with k8s term (as we do for other types), e.g. pvc, or persistentvolumeclaim (I know it's too long).

pkg/apis/workspaces/v1alpha2/volumeComponent.go
```diff
@@ -11,4 +11,44 @@ type Volume struct {
 	// +optional
 	// Size of the volume
 	Size string `json:"size,omitempty"`
+	// Persistent defines whether to use persistent storage for this volume. Defaults
+	// to true. When set to false, storage is tied to a pod lifecycle and is erased on
+	// delete.
+	// +optional
+	// +kubebuilder:default=true
+	Persistent bool `json:"persistent,omitempty"`
```
sleshchenko, line 19: Can it be used together with `External`? Should we deny it on the implementation level?

sleshchenko (approved):
I'm OK with the made proposal

pkg/apis/workspaces/v1alpha2/volumeComponent.go
amisevsk: Yeah I agree. One thing we discussed is that semantically it only makes sense to have either [persistent, size] or [external] but not both.

I also need to remove the readonly parameter for now, it doesn't make sense in its current spot.

pkg/apis/workspaces/v1alpha2/volumeComponent.go
amisevsk: PVC probably makes sense to make clear what we're planning to do with the information provided. I chose "storage" as it's more generic, but it may complicate things if we need to extend this in the future.

johnmcollier: What about `persistent`?

amisevsk: I like it! :+1:

amisevsk:
I'm closing this PR in favor of #324 and #325 which represent the same changes but as separate PRs. I think #324 (ephemeral/non-persistent volumes) makes sense to merge but #325 deserves more discussion, as pointed out by @l0rd.

pkg/apis/workspaces/v1alpha2/volumeComponent.go
amisevsk: Applied this change in https://github.com/devfile/api/pull/325

pkg/apis/workspaces/v1alpha2/volumeComponent.go
amisevsk: We can revisit this topic once #324 is merged and if we decide to go ahead on #325.

PR #324, opened 2021-01-26
Ephemeral volumes

amisevsk:
### What does this PR do?
This PR is half of https://github.com/devfile/api/pull/319 (support for ephemeral/emptyDir volume components). These components provide a way to share storage between components -- e.g. sharing a `.m2` volume between a builder container and a Java language server sidecar. 

Compared to https://github.com/devfile/api/pull/319, I've renamed the field from `persistent` to `ephemeral` (following @davidfestal's suggestion) to support having `false` as the default value; otherwise, we're generally required to use a boolean pointer, which is a hassle to work with. 

Fixes 1/2 of #310 

N/A, spec change.

### DIFF
--- a/pkg/apis/workspaces/v1alpha1/volumeComponent.go
+++ b/pkg/apis/workspaces/v1alpha1/volumeComponent.go
@@ -15,4 +15,9 @@ type Volume struct {
 	// +optional
 	// Size of the volume
 	Size string `json:"size,omitempty"`
+
+	// +optional
+	// Ephemeral volumes are not stored persistently across restarts. Defaults
+	// to false
+	Ephemeral bool `json:"ephemeral,omitempty"`
 }
--- a/pkg/apis/workspaces/v1alpha2/volumeComponent.go
+++ b/pkg/apis/workspaces/v1alpha2/volumeComponent.go
@@ -11,4 +11,9 @@ type Volume struct {
 	// +optional
 	// Size of the volume
 	Size string `json:"size,omitempty"`
+
+	// +optional
+	// Ephemeral volumes are not stored persistently across restarts. Defaults
+	// to false
+	Ephemeral bool `json:"ephemeral,omitempty"`
 }
--- a/crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
+++ b/crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
@@ -1657,6 +1657,11 @@ spec:
                                     description: Configuration overriding for a Volume
                                       component in a plugin
                                     properties:
+                                      ephemeral:
+                                        description: Ephemeral volumes are not stored
+                                          persistently across restarts. Defaults to
+                                          false
+                                        type: boolean
                                       name:
                                         description: Mandatory name that allows referencing
                                           the Volume component in Container volume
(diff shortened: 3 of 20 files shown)


sleshchenko (approved):

pkg/apis/workspaces/v1alpha1/volumeComponent.go
```diff
@@ -15,4 +15,9 @@ type Volume struct {
 	// +optional
 	// Size of the volume
 	Size string `json:"size,omitempty"`
+
+	// +optional
+	// Ephemeral volumes are not stored persistently across restarts. Defaults
+	// to false
+	Ephemeral bool `json:"ephemeral,omitempty"`
```
sleshchenko, line 22: I just wonder if it makes sense to use `temporary` here instead of avoid confusing between emptyDir and ephemeral volumes (which is in alpha but still) https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#volume-v1-core

pkg/apis/workspaces/v1alpha2/volumeComponent.go
```diff
@@ -11,4 +11,9 @@ type Volume struct {
 	// +optional
 	// Size of the volume
 	Size string `json:"size,omitempty"`
+
+	// +optional
+	// Ephemeral volumes are not stored persistently across restarts. Defaults
+	// to false
+	Ephemeral bool `json:"ephemeral,omitempty"`
```
maysunfaisal, line 18: @kadel what are your thoughts on this from odo perspective, since odo has a project preference for ephemeral volumes..

pkg/apis/workspaces/v1alpha1/volumeComponent.go
amisevsk: I originally had `persistent` here -- we switched to ephemeral because that's the k8s verbiage in [docs](https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/)

sleshchenko: I see that `Ephemeral` and `EmptyDir` are types of Ephemeral storage in the K8s.
So, I'm fully OK with using Ephemeral without implementation details, and then if needed - we could use like attributes to specify user desired implementation.

pkg/apis/workspaces/v1alpha2/volumeComponent.go
kadel: Ephemeral preference in odo is not related to volume component in devfile. This setting is used for internal odo volume that is used for the directory where the source code is stored (`/projects`).

odo is actually treating all component volumes as not ephemeral (it always creates PVC for each defined volume component)

kadel (approved):

pkg/apis/workspaces/v1alpha2/volumeComponent.go
maysunfaisal: odo preference may want to update the label for ephermeral as users may get confused in the future with the preference label and the devfile property on volumes.

davidfestal (approved):

elsony (approved):
/lgtm

PR #325, opened 2021-01-26
External volumes

amisevsk:
### What does this PR do?
This PR is half of #310 (support for "external" volumes). It adds support for mounting existing resources (e.g. configmaps, secrets, or persistent volumes) into devfile components without the implicit assumption that they're defined by the devfile itself. Use cases are:
- Mounting config files or ssh secrets
- Mounting shared volumes (e.g. static assets)

I've separated it out as this is a change we should consider carefully -- referring to volumes that are assumed to exist on the cluster makes devfiles less portable. Without a defined way of handling volumes we can't find, we risk e.g. my devfile not being compatible with your cluster (unless you happen to have the same configmaps/secrets defined). (Thanks @l0rd for pointing this out.)

Half of #310

N/A


sleshchenko (commented):
```yaml
components:
- name: cfg
  volumes:
    external:
      name: my-app-cfg
      type: secret
```
Looks simple but it makes sense to think if we want to support advanced parameters for each of type, like `optional`, `items` for configmap and secret, and `readonly` for `storage`.
If we want, then we may want to go the following way:

```yaml
components:
- name: cfg
  volumes:
    external:
      secret:
        name: my-app-cfg # can be secretName if we follow k8s
        optional: true # optional is optional =)
        items: ...
```
the only drawback - one more nesting level

pkg/apis/workspaces/v1alpha1/volumeComponent.go
```diff
@@ -15,4 +15,38 @@ type Volume struct {
...
+	// kubebuilder:validation:Enum="persistent,configmap,secret"
+	Type ExistingVolumeType `json:"type"`
+}
+
+// ExistingVolumeType defines the type of an external Volume
+type ExistingVolumeType string
+
+const (
+	// PersistentVolumeType specifies persistent storage, e.g. a PersistentVolumeClaim
+	PersistentVolumeType ExistingVolumeType = "persistent"
```
kadel, line 47: If this is always going reference PVC would it be better to go with full name? `persistentvolumeclaim`?

amisevsk: I prefer the abstraction that `persistent` gives (plus it's shorter), but I'm fine either way. In general, tying our API 1-1 with kubernetes makes the whole thing feel like a weird indirection layer rather than an API.

elsony (commented):
Are `external` and the existing `size` mutually exclusive, i.e. if we are using `external` volume, then the `size` field cannot be specified and vice versa?

amisevsk:
> Are `external` and the existing `size` mutually exclusive, i.e. if we are using `external` volume, then the `size` field cannot be specified and vice versa?

From a k8s perspective, yes, the fields are exclusive -- most of the top-level volume fields are invalid if `external` is specified; I don't know a good solution there.

Issue #339, opened 2021-02-09
nil pointer dereference in Starterproject validation


yangcao77:
I'm seeing this error when I try to validate starterproject in my devfile:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x1988416]

goroutine 1 [running]:
github.com/devfile/api/v2/pkg/validation.ValidateStarterProjects(0xc00009b5f0, 0x2, 0x2, 0x0, 0x0)
	/Users/stephanie/go/pkg/mod/github.com/devfile/api/v2@v2.0.0-20210202172954-6424f4139ac7/pkg/validation/projects.go:28 +0x136
```

It is because the `CheckOutFrom` is nil, but the code is trying to get the value of `CheckOutFrom.Remote`
https://github.com/devfile/api/blob/019d88e3a9102768be6d2e9b37fa418517ac16b0/pkg/validation/projects.go#L28

Need to check if `CheckOutFrom != nil` before deference it. 

Same check should also applys to: https://github.com/devfile/api/blob/019d88e3a9102768be6d2e9b37fa418517ac16b0/pkg/validation/projects.go#L61-L62

PR #341, opened 2021-02-10
project & starterproject validation bug

yangcao77:
### What does this PR do?
This PR fixes the bug in project & starterproject validation. detailed description in #339 

https://github.com/devfile/api/issues/339

Added more unit test cases to increase invalid project/starterproject test coverage

### DIFF
--- a/pkg/validation/projects.go
+++ b/pkg/validation/projects.go
@@ -63,7 +63,7 @@ func ValidateProjects(projects []v1alpha2.Project) error {
 				errString += fmt.Sprintf("\nproject %s has more than one remote defined, but has no checkoutfrom remote defined", project.Name)
 				continue
 			}
-				err := validateRemoteMap(gitSource.Remotes, gitSource.CheckoutFrom.Remote, project.Name)
+			err := validateRemoteMap(gitSource.Remotes, gitSource.CheckoutFrom.Remote, project.Name)
 			if err != nil {
 				errString += fmt.Sprintf("\n%s", err.Error())
 			}
--- a/pkg/validation/projects.go
+++ b/pkg/validation/projects.go
@@ -59,11 +59,11 @@ func ValidateProjects(projects []v1alpha2.Project) error {
 		if len(gitSource.Remotes) == 0 {
 			errString += fmt.Sprintf("\nprojects %s should have at least one remote", project.Name)
 		} else if len(gitSource.Remotes) > 1 || (gitSource.CheckoutFrom != nil && gitSource.CheckoutFrom.Remote != "") {
-			var checkoutRemotes string
-			if gitSource.CheckoutFrom != nil {
-				checkoutRemotes = gitSource.CheckoutFrom.Remote
+			if len(gitSource.Remotes) > 1 && (gitSource.CheckoutFrom == nil || (gitSource.CheckoutFrom != nil && gitSource.CheckoutFrom.Remote == "")) {
+				errString += fmt.Sprintf("\nproject %s has more than one remote defined, but has no checkoutfrom remote defined", project.Name)
+				continue
 			}
-				err := validateRemoteMap(gitSource.Remotes, checkoutRemotes, project.Name)
+				err := validateRemoteMap(gitSource.Remotes, gitSource.CheckoutFrom.Remote, project.Name)
 			if err != nil {
 				errString += fmt.Sprintf("\n%s", err.Error())
 			}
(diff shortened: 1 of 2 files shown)


pkg/validation/projects.go
```diff
@@ -58,7 +58,11 @@ func ValidateProjects(projects []v1alpha2.Project) error {
 
 		if len(gitSource.Remotes) == 0 {
 			errString += fmt.Sprintf("\nprojects %s should have at least one remote", project.Name)
-		} else if len(gitSource.Remotes) > 1 || gitSource.CheckoutFrom.Remote != "" {
+		} else if len(gitSource.Remotes) > 1 || (gitSource.CheckoutFrom != nil && gitSource.CheckoutFrom.Remote != "") {
+			if len(gitSource.Remotes) > 1 && (gitSource.CheckoutFrom == nil || (gitSource.CheckoutFrom != nil && gitSource.CheckoutFrom.Remote == "")) {
+				errString += fmt.Sprintf("\nproject %s has more than one remote defined, but has no checkoutfrom remote defined", project.Name)
+				continue
+			}
```
maysunfaisal, line 65:
```suggestion
		} else if len(gitSource.Remotes) > 1 && gitSource.CheckoutFrom != nil {
			err := validateRemoteMap(gitSource.Remotes, gitSource.CheckoutFrom.Remote, project.Name)
			if err != nil {
				errString += fmt.Sprintf("\n%s", err.Error())
			}
		} else if len(gitSource.Remotes) > 1 && gitSource.CheckoutFrom == nil {
				errString += fmt.Sprintf("\nproject %s has more than one remote defined, but has no checkout from remote defined", project.Name)
		}
```

I would suggest breaking it to different else if conditions. My suggestion handles all the test cases you wrote but if we want to explicitly handle`""` checkoutfrom then we can do it the way you suggested,

pkg/validation/projects.go
```diff
@@ -58,7 +58,11 @@ func ValidateProjects(projects []v1alpha2.Project) error {
 
 		if len(gitSource.Remotes) == 0 {
 			errString += fmt.Sprintf("\nprojects %s should have at least one remote", project.Name)
-		} else if len(gitSource.Remotes) > 1 || gitSource.CheckoutFrom.Remote != "" {
+		} else if len(gitSource.Remotes) > 1 || (gitSource.CheckoutFrom != nil && gitSource.CheckoutFrom.Remote != "") {
+			if len(gitSource.Remotes) > 1 && (gitSource.CheckoutFrom == nil || (gitSource.CheckoutFrom != nil && gitSource.CheckoutFrom.Remote == "")) {
```
amisevsk, line 62: This conditional sequence is kind of a nightmare, something like `if A || (B && C) then: if A && (!B || (B && !C)) then error`, and should be simplified. Maybe something like 
```go
switch len(gitSource.Remotes) {
case 0:
  errString += fmt.Sprintf("\nprojects %s should have at least one remote", project.Name)
case 1:
  if gitSource.CheckoutFrom != nil {
    if err := validateRemoteMap(gitSource.Remotes, gitSource.CheckoutFrom.Remote, project.Name); err != nil {
      errString += fmt.Sprintf("\n%s", err.Error())
    }
  }
default: // len(gitSource.Remotes) >= 2
  if gitSource.CheckoutFrom == nil || gitSource.CheckoutFrom.Remote == "" {
    errString += fmt.Sprintf("\nproject %s has more than one remote defined, but has no checkoutfrom remote defined", project.Name)
    continue
  }
  if err := validateRemoteMap(gitSource.Remotes, gitSource.CheckoutFrom.Remote, project.Name); err != nil {
    errString += fmt.Sprintf("\n%s", err.Error())
  }
}
```
Maybe some of this logic can be moved into `validateRemoteMap`? That function seems to simple to be broken out since it's just a map access.

maysunfaisal (commented):
generally looks good, just a small suggestion

pkg/validation/projects.go
amisevsk: I believe this suggestion doesn't cover
```yaml
git:
  remotes: ["my-remote"]
  checkoutFrom:
    remote: "another-remote"
```
but I didn't read too closely.

:+1: on splitting this logic up though, it's very complicated.

pkg/validation/projects.go
yangcao77: Thanks for the suggestion. Using a case switch does make the code more readable. 
I have also updated the starterProject validation to use case switch as well to be consistent.

amisevsk (approved):
LGTM :+1:

Issue #295, opened 2021-01-11
[Merging] passing in `nil` in MergeDevWorkspaceTemplateSpec causes panic


amisevsk:
## Describe the bug
Passing `nil` for `parentFlattenedContent` or `pluginFlattenedContent` in `overriding.MergeDevWorkspaceTemplateSpec` causes a panic due to nil pointer.

Stacktrace:
```
E0111 12:31:43.096684   57531 runtime.go:78] Observed a panic: "value method github.com/devfile/api/pkg/apis/workspaces/v1alpha2.DevWorkspaceTemplateSpecContent.GetToplevelLists called using nil *DevWorkspaceTemplateSpecContent pointer" (value method github.com/devfile/api/pkg/apis/workspaces/v1alpha2.DevWorkspaceTemplateSpecContent.GetToplevelLists called using nil *DevWorkspaceTemplateSpecContent pointer)
goroutine 679 [running]:
panic(0x1d8d280, 0xc000908b30)
	/usr/lib/golang/src/runtime/panic.go:975 +0x41b
github.com/devfile/api/pkg/apis/workspaces/v1alpha2.(*DevWorkspaceTemplateSpecContent).GetToplevelLists(0x0, 0x0)
	<autogenerated>:1 +0xde
github.com/devfile/api/pkg/utils/overriding.checkKeys(0x20187f0, 0xc000a4b9c0, 0x2, 0x2, 0x0, 0x0)
	/home/amisevsk/go/pkg/mod/github.com/devfile/api@v0.0.0-20201125082321-aeda60d43619/pkg/utils/overriding/keys.go:26 +0x175
github.com/devfile/api/pkg/utils/overriding.ensureNoConflictWithParent(0xc00037cf50, 0x0, 0x0, 0x0)
	/home/amisevsk/go/pkg/mod/github.com/devfile/api@v0.0.0-20201125082321-aeda60d43619/pkg/utils/overriding/merging.go:156 +0xf3
github.com/devfile/api/pkg/utils/overriding.MergeDevWorkspaceTemplateSpec(0xc00037cf50, 0x0, 0xc000492d08, 0x1, 0x1, 0x0, 0x0, 0x0)
	/home/amisevsk/go/pkg/mod/github.com/devfile/api@v0.0.0-20201125082321-aeda60d43619/pkg/utils/overriding/merging.go:32 +0x28f

```

## Expected behavior
Nil check should be done, and that component should be skipped

PR #352, opened 2021-02-18
Add Global Attributes, Global Variable & Substitution and Deprecate Metadata.Attributes

### LINKED ISSUE #239: Top level attribute definition in a devfile
The existing devfile does not have a way to define global attributes. If there are some values that I wanted to reuse in different sections of the devfile, I need to repeat those values in various places within a devfile. 

Sample use case:
I wanted to create a couple of maven stacks and all of them have pretty much the same content except that the commands/images will make use of a different version of java. If I can define global attributes in a devfile, I can use the attribute key in places where I need to reference the java version, e.g. image path, commands, etc. Then, I can just have a  single parent for those maven stacks and each child devfile of them just override the java version attribute in the devfile to provide the specific java version behaviour in different stacks.

Proposal:
1. Define attributes as a map of implementation-dependent free-form YAML attributes on the top level in a devfile.
2. Use `{{attribute-name}}` to refer to the attribute with the key `attribute-name`
3. To the implementation simple and avoiding to the need introducing hierarchy path on attribute reference, only attributes on the top level can be resolved within a devfile. Attributes defined on other levels, e.g. attributes defined under a command, cannot be resolved.

Example:
This example defines a global attribute java version number on the top level and references the value of the variable in the container image field.
```
schemaVersion: 2.0.0
metadata:
  name: java-maven
  version: 1.1.1
attributes:
  javaVersion: 11
components:
  - name: tools
    container:
      image: quay.io/eclipse/che-java{{javaVersion}}-maven:nightly
...
```

Note: Using this mechanism, we can also address the requirement https://github.com/devfile/api/issues/149 where each container can refer to the global attribute value and add the required environment variables in the container definition.

maysunfaisal:

### What does this PR do?
- Introduces a global attribute property (not a top level list since Attribute is a free form yaml map and does not have name/id)
- Implements Attribute overriding and merging as per the current devfile overriding and merging implementation for other fields
- Implements substitution for global attributes
  - global attribute is referenced by `{{attributeKey}}` where `attributeKey` is the key entry in the global attribute
  - valid for all string fields in the devfile except for `schemaVersion` & `metadata`
  - exceptions are top level keys like name/id as they're guarded by regex and also required for overriding & merging and string enums as they're guarded by schema validation
  - exceptions also include int & boolean properties, as fields for these properties cannot accept `{{}}` as value in the devfile
  
####  :pushpin: EDIT POST DEVFILE CABAL AND PR REVIEW DISCUSSION:
Post Devfile cabal call & PR review discussion, the above criteria has been modified, Please refer https://github.com/devfile/api/pull/352#issuecomment-806857030 for the PR acceptance criteria

Fixes #239 

Yes, updated current tests for overriding & merging and added new go tests.


elsony (changes requested):

crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
```diff
@@ -4137,6 +4137,13 @@ spec:
                 description: Structure of the workspace. This is also the specification
                   of a workspace template.
                 properties:
+                  attributes:
+                    description: Map of implementation-dependant free-form YAML attributes.
+                      Attribute values can be referenced through out the devfile in
```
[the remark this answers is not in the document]
maysunfaisal: Addressed

crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
```diff
@@ -4910,6 +4917,12 @@ spec:
                           - required:
                             - kubernetes
                           properties:
+                            attributes:
+                              description: Overrides of attributes encapsulated in
+                                a parent devfile or a plugin. Overriding is done according
```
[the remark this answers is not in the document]
maysunfaisal: I thought about it removing earlier but the plugin PR is still a WIP but I've removed it now

crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
```diff
@@ -5742,6 +5755,12 @@ spec:
                     - required:
                       - kubernetes
                     properties:
+                      attributes:
+                        description: Overrides of attributes encapsulated in a parent
+                          devfile or a plugin. Overriding is done according to K8S
```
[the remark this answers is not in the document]
maysunfaisal: I thought about it removing earlier but the plugin PR is still a WIP but I've removed it now

crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
```diff
@@ -4137,6 +4137,13 @@ spec:
                 description: Structure of the workspace. This is also the specification
                   of a workspace template.
                 properties:
+                  attributes:
+                    description: Map of implementation-dependant free-form YAML attributes.
+                      Attribute values can be referenced through out the devfile in
+                      string type fields in the form {{attribute-key}} except for
+                      schemaVersion and metadata
```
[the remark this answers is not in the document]
maysunfaisal: Addressed

pkg/validation/attributes/attributes.go
```diff
@@ -0,0 +1,62 @@
...
+		// Validate the starter projects
+		if err = ValidateStarterProjects(workspaceTemplateSpec.Attributes, &workspaceTemplateSpec.StarterProjects); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+var globalAttributeRegex = regexp.MustCompile(`\{{2}(.*?)\}{2}`)
```
amisevsk, line 46: Nitpick (feel free to ignore): I feel like using the repetition regexp obscures the meaning here since the char being checked is `{`. Perhaps `\{\{(.*?)\}\}` is clearer? We may also want to exclude `{` and `}` from the inner match, to avoid weirdness around using e.g. `{{my{attribute}}`, but that's probably disallowed elsewhere.

pkg/validation/attributes/attributes.go
```diff
@@ -0,0 +1,62 @@
...
+)
+
+// ValidateGlobalAttribute validates the workspace template spec data for global attribute references
+func ValidateGlobalAttribute(workspaceTemplateSpec *v1alpha2.DevWorkspaceTemplateSpec) error {
+
+	var err error
+
+	if workspaceTemplateSpec != nil {
+		// Validate the components
+		if err = ValidateComponents(workspaceTemplateSpec.Attributes, &workspaceTemplateSpec.Components); err != nil {
```
amisevsk, line 18: I'm not clear on why we're passing references to slices throughout this code -- if the concern is being able to modify the underlying data, passing the slices is sufficient (and cleaner). Is there a reason we need to pass pointers to structs everywhere?

For context, passing `workspaceTemplateSpec.Components` *will* copy the slice struct, but the slice struct is just a bunch of references to memory locations.

pkg/validation/attributes/attributes.go
```diff
@@ -0,0 +1,62 @@
...
+// validateAndReplaceDataWithAttribute validates the string for a global attribute and replaces it. An error
+// is returned if the string references an invalid global attribute key
+func validateAndReplaceDataWithAttribute(val string, attributes apiAttributes.Attributes) (string, error) {
+	matches := globalAttributeRegex.FindAllStringSubmatch(val, -1)
+	for _, match := range matches {
+		var err error
+		attrValue := attributes.GetString(match[1], &err)
+		if err != nil {
+			return "", err
+		}
```
amisevsk, line 57: Are we sure we want to consider this an error case? It basically forbids using `{{text}}` anywhere in a Devfile. I think a safer choice is to leave it unmodified.

pkg/validation/attributes/attributes.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"regexp"
+	"strings"
+
+	"github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
+	apiAttributes "github.com/devfile/api/v2/pkg/attributes"
+)
+
+// ValidateGlobalAttribute validates the workspace template spec data for global attribute references
+func ValidateGlobalAttribute(workspaceTemplateSpec *v1alpha2.DevWorkspaceTemplateSpec) error {
```
amisevsk, line 12: This function (and the subfunctions below) are named `Validate[...]` but in reality they appear to validate *and* replace attributes. We should make this functionality clear (and potentially have purely validation-focussed paths)

pkg/validation/attributes/attributes_event.go
```diff
@@ -0,0 +1,45 @@
...
+
+// ValidateEvents validates the events data for a global attribute
+func ValidateEvents(attributes apiAttributes.Attributes, events *v1alpha2.Events) error {
+	var err error
+
+	if events != nil {
+		switch {
+		case len(events.PreStart) > 0:
+			for i := range events.PreStart {
+				if events.PreStart[i], err = validateAndReplaceDataWithAttribute(events.PreStart[i], attributes); err != nil {
```
amisevsk, line 16: I'm curious on the design here -- `events.Prestart` is a slice of strings referring to command IDs, and command IDs cannot have `{` and `}` in them, which makes it seem impossible to cleanly use this functionality. I'm not sure we should support attributes inside `events` unless we relax the regexes for validation (though that poses further problems)

pkg/validation/attributes/attributes_test.go
```diff
@@ -0,0 +1,257 @@
...
+	tests := []struct {
+		name     string
+		testFile string
+		expected v1alpha2.DevWorkspaceTemplateSpec
+		wantErr  bool
+	}{
+		{
+			name:     "Successful global attribute substitution",
+			testFile: "test-fixtures/all/devfile-good.yaml",
+			expected: v1alpha2.DevWorkspaceTemplateSpec{
```
amisevsk, line 24: For these tests in general, consider putting all input/output in files -- it's hard to track where replacements have occurred between a yaml file and a golang struct initialization.

pkg/validation/attributes/attributes_command_test.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			testExecCommand := v1alpha2.ExecCommand{}
+
+			readFileToStruct(t, tt.testFile, &testExecCommand)
+
+			err := validateExecCommand(tt.attributes, &testExecCommand)
+			if tt.wantErr == (err == nil) {
+				t.Errorf("error: %v", err)
```
amisevsk, line 62: This isn't really descriptive enough IMO -- you could have a test fail with `error: <nil>` which could be frustrating. There's no real need for terseness here:
```go
if tt.wantErr && err == nil {
  t.Errorf("Expected error from test but got nil")
} else if !tt.wantErr && err != nil {
  t.Errorf("Got unexpected error: %s", err)
}
```

amisevsk (commented):
Overall looks good, with a few comments/questions. One more general confusion I could foresee is that we reuse the term `attributes` despite *this* attributes field being special, and others being just plain. Is this confusing?
```yaml
schemaVersion: 2.1.0
metadata:
  name: test-devfile
attributes:
  go-version: "1.14.7"
components:
  name: my-component
  attributes:
    go-version: "1.15.6"
  image: my-sidecar-go:{{go-version}}
```

pkg/apis/workspaces/v1alpha2/devworkspaceTemplateSpec.go
```diff
@@ -12,6 +14,15 @@ type DevWorkspaceTemplateSpec struct {
...
 // +devfile:overrides:generate
 type DevWorkspaceTemplateSpecContent struct {
+	// Map of implementation-dependant free-form YAML attributes.
+	// Attribute values can be referenced throughout the devfile in string type fields in the form {{attribute-key}}
+	// except for schemaVersion and metadata. Exception to the string field include element's key identifiers(command id,
+	// component name, endpoint name, project name, etc.) and string enums(command group kind, endpoint exposure, etc.)
+	// +optional
+	// +patchStrategy=merge
+	// +devfile:overrides:include:omitInPlugin=true,description=Overrides of attributes encapsulated in a parent devfile.
+	Attributes attributes.Attributes `json:"attributes,omitempty" patchStrategy:"merge"`
```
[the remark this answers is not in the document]
maysunfaisal: @amisevsk metadata attributes can store information about data that is not required to be used and substituted in other devfile fields. FWIW, I had the same question as you regd removal of the metadata attribute.

pkg/validation/attributes/attributes.go
maysunfaisal: Hmm interesting thought. So if we go by the suggested convention, it either replaces the valid `{{key}}` to its value or leaves the invalid key `{{invalidKey}}` as it is? Then there would be no error cases at all.. 

What happens when someone commits an unintentional typo or it maybe confusing if an error is not returned since we are allowing the global attributes to overriden and merged 🤔

pkg/validation/attributes/attributes_event.go
maysunfaisal: I think this is a valid point 👍

pkg/validation/attributes/attributes.go
maysunfaisal: Does `ValidateAndReplaceFor[...]` sound ok to you? I'll also update the func desc.

pkg/validation/attributes/attributes.go
maysunfaisal: `my{attribute` is a valid string for an attribute key, if we exclude them, are we restricting it too much? 🤔

pkg/apis/workspaces/v1alpha2/devworkspaceTemplateSpec.go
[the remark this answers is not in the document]
amisevsk: With that in mind then, are the global attributes intended to be _only_ used for substitution? I was kind of imagining using the global attributes for both, at which point `.metadata.attributes` is redundant.

pkg/validation/attributes/attributes.go
amisevsk: We're fully in the weeds of edge cases now :)

How would we parse `{{my{{attribute}}test}}`?

pkg/validation/attributes/attributes.go
amisevsk: Yeah, not requiring replacement does leave the ugliness of trying to execute something like `go{{go-version}}` (though ideally one would notice that pretty quickly). My main worry is that there might be a use case _somewhere_ that requires `{{text}}` and this approach would effectively make us incompatible (e.g. I know helm charts are littered with `{{}}`). 

Good point about parent attributes though -- I think the _most_ confusing outcome would be an attribute from parent replacing `{{my-var}}` in your code despite you never seeing the `my-var` attribute defined. I'm pretty split here.

pkg/validation/attributes/attributes.go
amisevsk: Makes sense

pkg/validation/attributes/attributes.go
maysunfaisal: The first occurrence of `{{` & `}}` ie, `my{{attribute` is the match and remaining `test}}` is considered a string, to keep it fairly simply and consistent 🤔

pkg/validation/attributes/attributes.go
maysunfaisal: @amisevsk regd the parent case, I think if a user is intending to use a parent devfile, they should be aware of the possible consequences of parent data being merged in the main devfile content, in this case a parent global attribute `my-var` replacing all occurances of `{{my-var}}` in the final devfile.

maysunfaisal: @elsony what are your thoughts about future incompatibility arising from using `{{}}` in devfile for global attributes because like @amisevsk says Helm charts uses this convention.

pkg/validation/attributes/attributes_command_test.go
maysunfaisal: Addressed

pkg/validation/attributes/attributes_test.go
maysunfaisal: Addressed

pkg/validation/attributes/attributes_event.go
maysunfaisal: Addressed

pkg/validation/attributes/attributes.go
maysunfaisal: Addressed

pkg/validation/attributes/attributes.go
maysunfaisal: Addressed

pkg/utils/overriding/keys.go
```diff
@@ -27,7 +30,27 @@ func checkKeys(doCheck checkFn, toplevelListContainers ...workspaces.TopLevelLis
...
+		var attributeValue reflect.Value
+		// toplevelListContainers can contain either a pointer or a struct and needs to be safeguarded when using reflect
+		if value.Kind() == reflect.Ptr {
+			attributeValue = value.Elem().FieldByName("Attributes")
+		} else {
+			attributeValue = value.FieldByName("Attributes")
+		}
+
+		if attributeValue.IsValid() && attributeValue.CanInterface() {
+			attributes := attributeValue.Interface().(attributesPkg.Attributes)
```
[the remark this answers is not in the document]
maysunfaisal: Addressed

pkg/validation/attributes/attributes.go
maysunfaisal: @sleshchenko not sure I follow, are we saying convention should be one pair `{}` ie,  `{var}`?

@amisevsk I thought about a bit more, I am not sure how devfile data will give way to incompatibility with a Helm chart. `{{my-var}}` in a devfile is either resolved or not. If its not resolved, we can err out(also depends upon our other conversation whether we should err out) or it will remain as `{{my-var}}`. If this data is being used in Helm, it will most likely not follow the Helm convention which is something like `{{.IMAGE}}`(notice the period) so in those cases will likely result in a Helm failure 🤔

pkg/validation/attributes/attributes_command.go
```diff
@@ -0,0 +1,108 @@
...
+
+	if composite != nil {
+		// Validate composite label
+		if composite.Label, err = validateAndReplaceDataWithAttribute(composite.Label, attributes); err != nil {
+			return err
+		}
+
+		// Validate composite commands
+		for i := range composite.Commands {
+			if composite.Commands[i], err = validateAndReplaceDataWithAttribute(composite.Commands[i], attributes); err != nil {
```
amisevsk, line 82: This seems to be another case of using attribute replacement in a place where the value you're replacing is an ID (and so cannot be an attribute), since `composite.Commands` refers to command IDs.

pkg/validation/attributes/attributes_command.go
```diff
@@ -0,0 +1,108 @@
...
+	if exec != nil {
+		// Validate exec command line
+		if exec.CommandLine, err = validateAndReplaceDataWithAttribute(exec.CommandLine, attributes); err != nil {
+			return err
+		}
+
+		// Validate exec component
+		if exec.Component, err = validateAndReplaceDataWithAttribute(exec.Component, attributes); err != nil {
+			return err
+		}
```
amisevsk, line 47: How would this work? `exec.Component` has to be one of the component IDs in the devfile, and so cannot have any `{{}}` chars. This would mean that you could have a global attribute doing something like
```yaml
commands:
  - id: test
    exec:
      component: {{my-component}}
```
but the value of the `my-component` attribute is now bound by the same rules as the value here (must refer to a component key). This sort of thing would also make validation more difficult, as we would have to replace attributes before validating. 

Is there a use case where we would want to configure the target component in this way?

pkg/validation/attributes/attributes_command.go
```diff
@@ -0,0 +1,108 @@
...
+	var err error
+
+	if apply != nil {
+		// Validate composite label
+		if apply.Label, err = validateAndReplaceDataWithAttribute(apply.Label, attributes); err != nil {
+			return err
+		}
+
+		// Validate apply component
+		if apply.Component, err = validateAndReplaceDataWithAttribute(apply.Component, attributes); err != nil {
```
amisevsk, line 102: Similar comment as above, for `exec.Component`. I'm not sure there's a use-case for this.

pkg/validation/attributes/attributes_component.go
```diff
@@ -0,0 +1,189 @@
...
+		// Validate container env
+		if len(container.Env) > 0 {
+			if err = validateAndReplaceForEnv(attributes, container.Env); err != nil {
+				return err
+			}
+		}
+
+		// Validate container volume mounts
+		for i := range container.VolumeMounts {
+			if container.VolumeMounts[i].Name, err = validateAndReplaceDataWithAttribute(container.VolumeMounts[i].Name, attributes); err != nil {
```
amisevsk, line 86: This has the same problem as command components; `volumeMount.Name` has to refer to a volume component in the DevWorkspace, so any attribute here would have a fixed value (or else the DevWorkspace is invalid).

pkg/validation/attributes/attributes_parent.go
```diff
@@ -0,0 +1,48 @@
...
+
+		// Validate parent registry url
+		if parent.RegistryUrl, err = validateAndReplaceDataWithAttribute(parent.RegistryUrl, attributes); err != nil {
+			return err
+		}
+
+		// Note: No need to substitute parent overrides at this point. Call global attribute validation/substitution
+		// after merging the flattened parent devfile to the main devfile. Parent's global attribute key can
+		// be overridden in parent overrides or the alternative is to mention the attribute as a main devfile
+		// global attribute if parent devfile does not have a global attribute
```
amisevsk, line 44: This is kind of subtle and results in a fairly complicated replacement flow:
1. Do attribute replacement to fill placeholders in `parent` section
    - This *might fail* if a component in the current devfile expects to use an attribute from its parent
2. Flatten parent into Devfile to get a devfile without parent elements
3. Do attribute replacement again, to handle placeholders in parent
4. Validate devfile only at this point

Global attributes + replacement means that there are various places where devfiles are invalid until some other steps happen.

pkg/validation/attributes/attributes.go
amisevsk: I've thought about it a bit more and I'm going towards "this shouldn't be an error". Failing here causes subtle complexity that's hard to resolve, e.g. the devfiles:
```yaml
metadata:
  name: parent-devfile
attributes:
  parent-attribute: "test"
components:
  - name: parent-component
    container:
      image: quay.io/test/image
---
metadata:
  name: main-devfile
attributes:
  parent-devfile-id: "parent-devfile"
parent:
  id: "{{parent-devfile-id}}"
components:
  - name: child-component
    image: quay.io/{{parent-attribute}}/image
```
become impossible to reconcile: doing the replacement in `main-devfile` will fail because `{{parent-attribute}}` does not exist, and we can't flatten in the parent attribute becuase its id is hidden behind `{{parent-devfile-id}}`, which requires attribute replacement.

pkg/apis/workspaces/v1alpha2/devworkspaceTemplateSpec.go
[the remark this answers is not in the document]
amisevsk: Unless we're explicitly forbidding the global attributes introduced here from storing non-replacement attributes, then I think it doesn't make sense to have both -- it'll just be a source of confusion. If global attributes to be used *just* for replacement, then why isn't their type `map[string]string`?

pkg/validation/attributes/test-fixtures/commands/apply.yaml
```diff
@@ -0,0 +1,3 @@
+# Attributes are defined in test-fixtures/attributes/attributes-referenced.yaml
+label: "{{version}}"
+component: "{{foo}}"
```
amisevsk, line 3: Trailing newlines missing in various files here, may as well fix it now in case we get a linter one day.

amisevsk (commented):
There are a few issues that I would like to see resolved before merging these changes (as much as I want to use them in the DevWorkspace Operator :wink: )

1. We should figure out what to do with `.metadata.attributes` and `.attributes`. It doesn't make sense to have both, so I think we should get rid of `metadata.attributes` -- otherwise it'll just be confusing.
2. There's some subtlety around how replacement + validation + flattening has to happen once global attributes (as implemented here) are added (see my review comments below)
3. We're still replacing attributes in a few cross-reference fields (e.g. volumeMounts in container components that have to refer to volume component ids), which we need to be very careful about. I'm not sure it's a valid usecase, and so I think any field that refers to another field in the devfile should not support attributes replacement.

pkg/validation/attributes/attributes_parent.go
maysunfaisal: So I've thought about this when drafting this PR and the way I see it is (copying devfile snipppet from your comment below),

```yaml
metadata:
  name: parent-devfile
attributes:
  parent-attribute: "test"
components:
  - name: parent-component
    container:
      image: quay.io/test/image
---
metadata:
  name: main-devfile
attributes:
  parent-devfile-id: "parent-devfile"
parent:
  id: "{{parent-devfile-id}}"
components:
  - name: child-component
    image: quay.io/{{parent-attribute}}/image
```

To resolve this flow, in order

1. resolve the parent in the main-devfile, so in this case the id `{{parent-devfile-id}}`(uri/kubernetes/registryURL) needs to be resolved to fetch the parent
2. overwrite the parent-devfile with main-devfile's parent overrides (teh regular override convention)
3. merge parent-devfile to main-devfile (the regular merging convention)
4. now validate the flatten devfile for global attributes

So we definitely need to resolve a parent source in the main-devfile to fetch the parent and is the only requirement here, unless we say parent source cannot have a global attribute as well.. WDYT?

As for your comment regd various places in devfile maybe invalid until some steps happen - isn't that true for other devfile objects? For example, my main-devfile can be the following and is invalid by itself in the main-devfile but is valid in a flattened devfile.. 🤔 

```yaml
metadata:
  name: parent-devfile
components:
  - name: parent-component
    container:
      image: quay.io/test/image
---
metadata:
  name: main-devfile
parent:
  id: "parent-devfile-id"
command:
  - id: child-component
     exec:
        component: parent-component
```

pkg/validation/attributes/attributes_command.go
maysunfaisal: makes sense, this was an oversight after your comment regd events in the prev feedback

pkg/validation/attributes/attributes_component.go
maysunfaisal: makes sense, this was an oversight after your comment regd events in the prev feedback

pkg/validation/attributes/attributes_command.go
maysunfaisal: makes sense, this was an oversight after your comment regd events in the prev feedback

pkg/validation/attributes/attributes_command.go
maysunfaisal: makes sense, this was an oversight after your comment regd events in the prev feedback

pkg/validation/attributes/attributes.go
maysunfaisal: @amisevsk I've explained the steps in this comment here regd this case https://github.com/devfile/api/pull/352#discussion_r587591297 pls do share your thoughts

pkg/validation/attributes/attributes_parent.go
amisevsk: This makes sense to me (and good point about unflattened devfiles in general). I'm on the side of "no attributes in parent" except maybe in overrides to make step 1 simpler.

maysunfaisal: @amisevsk Ack! Will remove global attributes from parent source but the overrides can have them(following the agreed upon observation that override object's id/name cannot have global attributes) and they will be resolved in the flattened devfile which is the step 4 mentioned above.

pkg/apis/workspaces/v1alpha2/devworkspaceTemplateSpec.go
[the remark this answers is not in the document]
maysunfaisal: @amisevsk I've talked with Elson. We're definitely interested in keeping metadata attributes as they will be used for information pertaining to devfiles. And they've been used in the past by both odo and Console devfiles.

With that being said, as I see Global Attributes at this point to be used for replacement, I am changing the type to `map[string]string` as you suggested to avoid confusion.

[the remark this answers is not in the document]
metlos: If the intended usage of the global attributes is just for replacement, what about just calling them such? E.g.:

```yaml
schemaVersion: 2.1.0
metadata:
  name: my-devfile
  attributes:
    # free-form yaml
substitutions:
  # string: string
```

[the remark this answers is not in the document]
maysunfaisal: @metlos  Not sure if others will agree on `substitutions` but open to suggestions 😄 
@amisevsk any opinion regd the rename?

pkg/validation/attributes/attributes_parent.go
maysunfaisal: Addressed in new commit :)

pkg/validation/attributes/attributes_command.go
maysunfaisal: Addressed in new commit :)

pkg/validation/attributes/attributes_command.go
maysunfaisal: Addressed in new commit :)

pkg/validation/attributes/attributes_command.go
maysunfaisal: Addressed in new commit :)

pkg/utils/overriding/test-fixtures/patches/global-attributes/result.yaml
```diff
@@ -0,0 +1,9 @@
+attributes:
+  boolAttributeToChange: false
+  stringAttributeToKeep: stringValue
+  objectAttributeToChange:
+
+    newObjectAttributeField:
+      objectAttributeSubField: 11
+      newObjectAttributeSubField: newObjectAttributeFieldValue
+    newField: true
```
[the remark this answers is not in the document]
maysunfaisal: Addressed in new commit :)

pkg/apis/workspaces/v1alpha2/devworkspaceTemplateSpec.go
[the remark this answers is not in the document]
maysunfaisal: Addressed type in new commit :)

pkg/apis/workspaces/v1alpha2/devworkspaceTemplateSpec.go
```diff
@@ -12,6 +12,16 @@ type DevWorkspaceTemplateSpec struct {
...
 type DevWorkspaceTemplateSpecContent struct {
+	// Map of string attributes.
+	// Attribute values can be referenced throughout the devfile in string type fields in the form {{attribute-key}}
+	// except for schemaVersion, metadata, parent source. Exception to the string field also include element's key identifiers
+	// (command id, component name, endpoint name, project name, etc.) and their references(events, command's component, container's
+	// volume mount name, etc.) and string enums(command group kind, endpoint exposure, etc.)
+	// +optional
+	// +patchStrategy=merge
+	// +devfile:overrides:include:omitInPlugin=true,description=Overrides of attributes encapsulated in a parent devfile.
+	Attributes map[string]string `json:"attributes,omitempty" patchStrategy:"merge"`
```
amisevsk, line 23: Moving to `map[string]string` complicates the issue further IMO. Now we have adjacent attributes fields that take different types of data.

maysunfaisal: @amisevsk so do you recommend a rename of this field or stay as Attributes and move back to free-form yaml? Btw I'm not sure if you have read my prev comment here https://github.com/devfile/api/pull/352#discussion_r589697319, but we're definitely interested in keeping metadata.Attributes.

If we want to keep it as `map[string]string`, then it might be a good idea to rename it? 🤔

kadel:
I just realized someting. (Hope it is not too late) 
Wouldn't it be better to call this something else than **attributes**?  **variables** might make more sense.
We already have **_attributes_** in devfile, and it is meant for a different use than this top-level **attributes**.
I'm afraid that this will create confusion. 

ping @elsony @davidfestal

maysunfaisal:
> I just realized someting. (Hope it is not too late)
> Wouldn't it be better to call this something else than **attributes**? **variables** might make more sense.
> We already have **_attributes_** in devfile, and it is meant for a different use than this top-level **attributes**.
> I'm afraid that this will create confusion.
> 
> ping @elsony @davidfestal

@kadel yes this PR is on hold, and that is one of the topics up for discussion in the next cabal call to see what everyone prefers to call it.

l0rd:
@maysunfaisal @kadel in the Che/DevWorkspace context we need attributes that are not in `metadata` (because `metadata` gets lost in the devfile --> devworkspacetemplate transformation). But we agree that having 2 top level `attributes` is confusing. What if we remove `attributes` at `metadata` level? What is this used for?

maysunfaisal:
> @maysunfaisal @kadel in the Che/DevWorkspace context we need attributes that are not in `metadata` (because `metadata` gets lost in the devfile --> devworkspacetemplate transformation). But we agree that having 2 top level `attributes` is confusing. What if we remove `attributes` at `metadata` level? What is this used for?

@l0rd there was one odo stack devfile that used metadata.attributes information previously, though I am not sure if they consume it at the moment - https://github.com/odo-devfiles/registry/blob/master/devfiles/java-openliberty/devfile.yaml#L6-L7. 

The console devfile dev preview also uses this metadata.attributes currently - https://github.com/redhat-developer/devfile-sample/blob/master/devfile.yaml#L5-L7

l0rd:
> @l0rd there was one odo stack devfile that used metadata.attributes information previously, though I am not sure if they consume it at the moment - [https://github.com/odo-devfiles/registry/blob/master/devfiles/java-openliberty/devfile.yaml#L6-L7](https://github.com/odo-devfiles/registry/blob/master/devfiles/java-openliberty/devfile.yaml?rgh-link-date=2021-03-22T17%3A36%3A24Z#L6-L7).

I do not see `metadata.attributes` in the devfile above 🤔 

> The console devfile dev preview also uses this metadata.attributes currently - [https://github.com/redhat-developer/devfile-sample/blob/master/devfile.yaml#L5-L7](https://github.com/redhat-developer/devfile-sample/blob/master/devfile.yaml?rgh-link-date=2021-03-22T17%3A36%3A24Z#L5-L7)

Who's the DevConsole point of contact to verify if we can still change that or if it's too late (i.e. it's used in the wild)?

maysunfaisal:
> > @l0rd there was one odo stack devfile that used metadata.attributes information previously, though I am not sure if they consume it at the moment - [https://github.com/odo-devfiles/registry/blob/master/devfiles/java-openliberty/devfile.yaml#L6-L7](https://github.com/odo-devfiles/registry/blob/master/devfiles/java-openliberty/devfile.yaml?rgh-link-date=2021-03-22T17%3A36%3A24Z#L6-L7).
> 
> I do not see `metadata.attributes` in the devfile above 🤔

My bad, but I think if the devfiles were to be updated, it would be in the metadata.attributes section. At least thats my understanding.

> > The console devfile dev preview also uses this metadata.attributes currently - [https://github.com/redhat-developer/devfile-sample/blob/master/devfile.yaml#L5-L7](https://github.com/redhat-developer/devfile-sample/blob/master/devfile.yaml?rgh-link-date=2021-03-22T17%3A36%3A24Z#L5-L7)
> 
> Who's the DevConsole point of contact to verify if we can still change that or if it's too late (i.e. it's used in the wild)?

That would be Andrew Ballantyne. I think the Console dev preview is waiting for the Devfile 2.2 outerloop support being discussed on the cabal call. I've already put up a topic for this on the next cabal call, we can discuss this then perhaps, with the wider team (I noticed that you also put up a topic on the next cabal call).

l0rd:
@maysunfaisal I added the same topic in the agenda, I didn't notice yours :-) Let remove mine. 

To clarify: we figured out today that we require top level attributes, NOT in metadata. That's necessary for devfile v2 support in Che (hence v2.1 of the spec). Those top level attributes have to be free form `yaml`, not `map[string]string`. That's one problem.

On the other hand what you are introducing with this PR is something distinct that I would rather call `variables`, `args`, `parameters` etc...For clarity I think it would be better to keep it separated for `attributes`.

Let's discuss that on this week cabal.

elsony:
1. Rename this string global replacement field to `variables`
2. Deprecate the existing attributes under `metadata` and recommend using the global attribute on top-level
3. Add freeform global attribute
4. If `variables` are not resolved, issue non-blocking warning
5. Both attribute on top-level and variable field support merging and overwriting

maysunfaisal:
I am planning on addressing all the points mentioned in Elson's comment here https://github.com/devfile/api/pull/352#issuecomment-806857030 with this PR since this PR started out as global attributes free form yaml.

So the code for it can be adjusted from my prev commits on this PR.

maysunfaisal:
@amisevsk will be grateful if you could pls help re-review this PR. It has been updated post the devfile cabal discussion. You may look from commit `Update Cabal Feedback - 1` onwards, thanks! 😄

pkg/validation/variables/utils.go
```diff
@@ -0,0 +1,44 @@
...
+	formatter := &logrus.TextFormatter{
+		FullTimestamp: true,
+	}
+	logrus.SetFormatter(formatter)
+}
+
+// logInvalidVariables logs a warning for all the invalid keys belong to a devfile object.
+//  We do not want to err out but only log a warning in case if variable key is not resolved.
+func logInvalidVariables(objectType, objectName string, invalidKeys []string) {
+	logrus.Warnf("Devfile %s %s has invalid variable references - %s", objectType, objectName, strings.Join(invalidKeys, ","))
```
amisevsk, line 20: I don't think it's acceptable to log in a library function -- if I call this from the controller and there's an issue, is it going to print logs in its own format? Library functions should return errors (or a 'warning' error type) that can be printed however the consumer deems appropriate.

amisevsk (commented):
I'll do a more detailed review on Tuesday next week :+1:

Thanks for your work @maysunfaisal

pkg/apis/workspaces/v1alpha2/devworkspaceTemplateSpec.go
```diff
@@ -12,6 +14,22 @@ type DevWorkspaceTemplateSpec struct {
 
 // +devfile:overrides:generate
 type DevWorkspaceTemplateSpecContent struct {
+	// Map of string variables.
+	// Variable values can be referenced throughout the devfile in string type fields in the form {{variable-key}}
+	// except for schemaVersion, metadata, parent source. Exception to the string field also include element's key identifiers
+	// (command id, component name, endpoint name, project name, etc.) and their references(events, command's component, container's
+	// volume mount name, etc.) and string enums(command group kind, endpoint exposure, etc.)
```
amisevsk, line 21: Wording could be clearer here

> Map of key-value variables used for string replacement in the devfile. Values can can be referenced via `{{variable-key}}` to replace the corresponding value in string fields in the devfile. Replacement cannot be used for 
> - schemaVersion, metadata, parent source
> - element identifiers, e.g. command id, component name, endpoint name, project name
> - references to identifiers, e.g. in events, a command's component, container's volume mount name
> - string enums, e.g. command group kind, endpoint exposure

or something

pkg/devfile/header.go
```diff
@@ -27,7 +27,7 @@ type DevfileMetadata struct {
 	// +kubebuilder:validation:Pattern=^([0-9]+)\.([0-9]+)\.([0-9]+)(\-[0-9a-z-]+(\.[0-9a-z-]+)*)?(\+[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$
 	Version string `json:"version,omitempty"`
 
-	// Map of implementation-dependant free-form YAML attributes.
+	// Map of implementation-dependant free-form YAML attributes. Deprecated, use the global attributes instead.
```
amisevsk, line 30:
```suggestion
	// Map of implementation-dependant free-form YAML attributes. Deprecated, use the top-level attributes field instead.
```

pkg/utils/overriding/merging.go
```diff
@@ -105,6 +106,30 @@ func MergeDevWorkspaceTemplateSpec(
...
+
+		var err error
+		if len(content.Attributes) > 0 {
+			if len(result.Attributes) == 0 {
+				result.Attributes = attributes.Attributes{}
+			}
+			for k, v := range content.Attributes {
+				result.Attributes.FromMap(map[string]interface{}{
+					k: v,
+				}, &err)
```
amisevsk, line 127:
```suggestion
				result.Attributes.Put(k, v, &err)
```

pkg/validation/variables/variables.go
```diff
@@ -0,0 +1,51 @@
...
+		ValidateAndReplaceForProjects(workspaceTemplateSpec.Variables, workspaceTemplateSpec.Projects)
+
+		// Validate the starter projects and replace for global variable
+		ValidateAndReplaceForStarterProjects(workspaceTemplateSpec.Variables, workspaceTemplateSpec.StarterProjects)
+	}
+
+	return
+}
+
+var globalVariableRegex = regexp.MustCompile(`\{\{(.*?)\}\}`)
```
amisevsk, line 30: Global vars should be at the top

amisevsk (commented):
Apart from logging and a few minor comments, generally LGTM.

pkg/validation/variables/utils.go
maysunfaisal: Updated, used a warning struct as discussed on Slack :)

pkg/apis/workspaces/v1alpha2/devworkspaceTemplateSpec.go
maysunfaisal: Updated!

pkg/devfile/header.go
maysunfaisal: Updated!

pkg/utils/overriding/merging.go
maysunfaisal: Updated!

pkg/validation/variables/variables.go
maysunfaisal: Updated!

pkg/validation/variables/variables_command_test.go
```diff
@@ -0,0 +1,149 @@
...
+			testApplyCommand := v1alpha2.ApplyCommand{}
+			readFileToStruct(t, tt.testFile, &testApplyCommand)
+
+			testVariable := make(map[string]string)
+			readFileToStruct(t, tt.variableFile, &testVariable)
+
+			err := validateAndReplaceForApplyCommand(testVariable, &testApplyCommand)
+			verr, ok := err.(*InvalidKeysError)
+			if tt.wantErr && !ok {
+				t.Errorf("Expected InvalidKeysError error from test but got %+v", verr)
```
amisevsk, line 139: If `!ok` here, you can't use the value of `verr` (should be `nil` I think)

pkg/validation/variables/variables_command_test.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			testApplyCommand := v1alpha2.ApplyCommand{}
+			readFileToStruct(t, tt.testFile, &testApplyCommand)
+
+			testVariable := make(map[string]string)
+			readFileToStruct(t, tt.variableFile, &testVariable)
+
+			err := validateAndReplaceForApplyCommand(testVariable, &testApplyCommand)
```
amisevsk, line 136: I don't see where we actually check the value of `err`. The test requires `outputFile` to be identical to `testFile` when we expect an error, which seems like a strange choice.

pkg/validation/variables/variables_component_test.go
```diff
@@ -0,0 +1,212 @@
...
+			testContainerComponent := v1alpha2.ContainerComponent{}
+			readFileToStruct(t, tt.testFile, &testContainerComponent)
+
+			testVariable := make(map[string]string)
+			readFileToStruct(t, tt.variableFile, &testVariable)
+
+			err := validateAndReplaceForContainerComponent(testVariable, &testContainerComponent)
+			verr, ok := err.(*InvalidKeysError)
+			if tt.wantErr && !ok {
+				t.Errorf("Expected InvalidKeysError error from test but got %+v", verr)
```
amisevsk, line 45: Same comments as for `variables_command_test.go`

pkg/utils/overriding/merging_test.go
```diff
@@ -0,0 +1,372 @@
...
+		}
+
+		resultJson, err := json.Marshal(result)
+		if err != nil {
+			t.Error(err)
+		}
+		resultYaml, err := yaml.JSONToYAML(resultJson)
+		if err != nil {
+			t.Error(err)
+		}
```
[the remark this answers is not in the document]
maysunfaisal: this were written prev, i just cut-paste them to a new file, but i'll update it

pkg/utils/overriding/merging_test.go
```diff
@@ -0,0 +1,372 @@
...
+		expectedJson, err := yamlMachinery.ToJSON(expected)
+		if err != nil {
+			t.Error(err)
+		}
+		expectedYaml, err := yaml.JSONToYAML(expectedJson)
+		if err != nil {
+			t.Error(err)
+		}
+
+		assert.Equal(t, string(expectedYaml), string(resultYaml), "The two values should be the same.")
```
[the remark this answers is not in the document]
maysunfaisal: this were written prev, i just cut-paste them to a new file, but i'll update it

pkg/utils/overriding/merging_test.go
```diff
@@ -0,0 +1,372 @@
...
+				return nil
+			}
+			main, err := ioutil.ReadFile(path)
+			if err != nil {
+				t.Error(err)
+				return nil
+			}
+			dirPath := filepath.Dir(path)
+			parent := []byte{}
+			parentFile := filepath.Join(dirPath, "parent.yaml")
```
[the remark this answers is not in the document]
maysunfaisal: this were written prev, i just cut-paste them to a new file, but i'll update it

pkg/validation/variables/variables_command_test.go
maysunfaisal: I dont necessarily check the contents of the err. I just make sure its `verr, ok := err.(*InvalidKeysError)` in the next line. As for the output file, since we decided that any unresolved substitution should remain as it is, I just make sure the devfile is as expected that is they have `{{invalid-var}}` when its unresolved.

pkg/utils/overriding/merging_test.go
```diff
@@ -0,0 +1,372 @@
...
+
+	tests := []struct {
+		name                    string
+		mainContent             *dw.DevWorkspaceTemplateSpecContent
+		parentFlattenedContent  *dw.DevWorkspaceTemplateSpecContent
+		pluginFlattenedContents []*dw.DevWorkspaceTemplateSpecContent
+		expected                *dw.DevWorkspaceTemplateSpecContent
+		wantErr                 *string
+	}{
+		{
```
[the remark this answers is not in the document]
maysunfaisal: I just tried to keep it consistent with the existing overriding_test, but i'll remove these since they're probably redundant

pkg/validation/variables/utils.go
```diff
@@ -0,0 +1,24 @@
...
+		}
+	}
+}
+
+// processInvalidKeys processes the invalid keys and return InvalidKeysError if present
+func processInvalidKeys(invalidKeys map[string]bool) error {
+	var invalidKeysArr []string
+	for key := range invalidKeys {
+		invalidKeysArr = append(invalidKeysArr, key)
+	}
```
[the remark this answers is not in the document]
maysunfaisal: I know, thats why I have used this func here https://github.com/devfile/api/pull/352#discussion_r610777774

pkg/validation/variables/variables_command_test.go
maysunfaisal: FWIW I have a test that does check the warning struct where the keys come from the errors. The test is in https://github.com/devfile/api/pull/352/files#diff-88492713a9572b112b36ac53c64c195ed31f959b4835a58d7e985d96cfecd912R31-R50

pkg/utils/overriding/merging_test.go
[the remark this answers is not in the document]
maysunfaisal: I think the prev author wanted to test the wrapper func for merging(and the same for overriding) which takes bytes[] instead of structs.

pkg/utils/overriding/merging_test.go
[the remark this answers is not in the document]
maysunfaisal: I think the original intention of the prev author was to have the result.yaml file contents sorted when comaring, the yaml file bytes is converted to JSON and then re-converted to YAML to maintain the order

```golang
                expectedJson, err := yamlMachinery.ToJSON(expected)
		if err != nil {
			t.Error(err)
		}
		expectedYaml, err := yaml.JSONToYAML(expectedJson)
		if err != nil {
			t.Error(err)
		}

		assert.Equal(t, string(expectedYaml), string(resultYaml), "The two values should be the same.")
``` 

I noticed that it results in different order otherwise:

```
FAIL: TestMerging/no-parent (0.00s)
        /Users/maysun/dev/redhat/api/pkg/utils/overriding/merging_test.go:40: 
            	Error Trace:	merging_test.go:40
            	Error:      	Not equal: 
            	            	expected: "components:\n  - name: test-plugin-component\n    container:\n      image: test-plugin-image\n  - name: test-component\n    container:\n      image: test-image\n"
            	            	actual  : "components:\n- container:\n    image: test-plugin-image\n  name: test-plugin-component\n- container:\n    image: test-image\n  name: test-component\n"
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -1,8 +1,8 @@
            	            	 components:
            	            	-  - name: test-plugin-component
            	            	-    container:
            	            	-      image: test-plugin-image
            	            	-  - name: test-component
            	            	-    container:
            	            	-      image: test-image
            	            	+- container:
            	            	+    image: test-plugin-image
            	            	+  name: test-plugin-component
            	            	+- container:
            	            	+    image: test-image
            	            	+  name: test-component
            	            	 
            	Test:       	TestMerging/no-parent
            	Messages:   	The two values should be the same.
 ```

File: pkg/utils/overriding/merging_test.go (hunk 33 above)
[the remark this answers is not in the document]
Comment (maysunfaisal) [reply pkg/utils/overriding/merging_test.go]: removed this test, and updated the files on an existing tests to mimic this one.

File: pkg/validation/variables/utils.go [hunk 35]
```diff
@@ -0,0 +1,24 @@
...
+
+	if len(invalidKeysArr) > 0 {
+		return &InvalidKeysError{Keys: invalidKeysArr}
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
maysunfaisal: updated!

pkg/validation/variables/utils.go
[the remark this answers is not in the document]
maysunfaisal: updated, the arr is now sorted before sending it via the error!

pkg/validation/variables/utils_test.go
```diff
@@ -0,0 +1,100 @@
...
+				t.Errorf("Expected error from test but got nil")
+			} else if !tt.wantErr && err != nil {
+				t.Errorf("Got unexpected error: %s", err)
+			}
+		})
+	}
+}
+
+// testStringArrElements checks if string slices have the same elements
+func testStringArrElements(aa, bb []string) bool {
```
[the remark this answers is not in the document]
maysunfaisal: updated, this has been removed

pkg/validation/variables/variables_command_test.go
maysunfaisal: updated to `err` for all the tests

pkg/validation/variables/variables_component_test.go
maysunfaisal: updated to `err` for all the tests

pkg/utils/overriding/merging_test.go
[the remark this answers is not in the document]
amisevsk: Huh, strange. I'm fine with keeping it as is but that should definitely have a doc explaining why it's necessary (otherwise someone like me might remove it when I see it in a year ;))

Another option is to compare structs directly, something along the lines of
```go
var expectedTemplate *dw.DevWorkspaceTemplateSpec
err = yaml.Unmarshal(expected, expectedTemplate)
if err != nil {
	t.Error(err)
}
assert.Equal(t, expectedTemplate, result, "The two values should be the same")
```
this makes some sense, since we're already serializing the result -- why not deserialize the expected instead?

If we want to keep the nice diff showing the error, this can be done using `go-cmp` as well -- replace the `assert.Equal` with
```go
assert.True(t, cmp.Equal(expectedTemplate, result), fmt.Sprintf("Output doesn't match expected\n%s", cmp.Diff(expectedTemplate, result)))
```
(I don't remember if `assert.Equal` prints nice diffs by default)

pkg/utils/overriding/merging_test.go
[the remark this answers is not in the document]
amisevsk: This makes sense, but is kind of ugly IMO. We've already had one panic bug due to testing the `bytes` version over the default: https://github.com/devfile/api/issues/295

I'm fine with leaving it out of this PR as it's probably out-of-scope though.

pkg/validation/variables/variables_command_test.go
amisevsk: :+1: makes sense to me.

amisevsk (commented):
Only one more request :tada: 

Thank you for your patience.

pkg/utils/overriding/merging_test.go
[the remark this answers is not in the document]
maysunfaisal: Thats a good idea, I've just done what you've suggested and used the template struct to compare. Updated for both merging and overriding tests

amisevsk (approved):
:tada: :tada: :tada:

elsony (approved):
/lgtm

crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
```diff
@@ -4140,6 +4140,10 @@ spec:
                 description: Structure of the devworkspace. This is also the specification
                   of a devworkspace template.
                 properties:
+                  attributes:
+                    description: Map of implementation-dependant free-form YAML attributes.
```
maysunfaisal, line 4144: dummy1

crds/workspace.devfile.io_devworkspacetemplates.v1beta1.yaml
```diff
@@ -6813,6 +6823,13 @@ spec:
                   uri:
                     description: Uri of a Devfile yaml file
                     type: string
+                  variables:
+                    additionalProperties:
+                      type: string
+                    description: Overrides of variables encapsulated in a parent devfile.
```
maysunfaisal, line 6829: dummy2

maysunfaisal (commented):
test review

PR #349, opened 2021-02-17
remove vscodetask and vscodelaunch, deprecate project.github

yangcao77:
Signed-off-by: Stephanie <yangcao@redhat.com>

### What does this PR do?
remove vscodetask and vscodelaunch, deprecate project.github
remove vacodetask & vacodelaunch related tests and sample devfiles

### What issues does this PR fix or reference?
#340 
#344

Issue #340, opened 2021-02-10
Remove devfile.projects[].github


benoitf:
**Is your feature request related to a problem? Please describe.**
when referencing a project I have git or github parameter but they're equals so github brings no value there as we don't change the way of defining a project like
```yaml
github:
   - organization: eclipse
   - repository: che
```
but as we specify the full url to the repo it's useless

**Describe the solution you'd like**
remove github as it brings confusion or make it fancier to define a github repository but if it's just to paste the git url there is no value.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

elsony:
git and github are duplicated. So we'll deprecate github and recommend user to use git.

elsony:
The `github` duplicate with the existing `git` setting. Instead of deprecating `github`, we have decided to remove `github` support in devfile 2.1.

PR #360, opened 2021-03-01
add tests for projects and starter projects

mmulholla:
Add test for projects and starter project to the new api test. 
The code added will be shared with the library tests.

### DIFF
--- a/test/v200/utils/common/project_test_utils.go
+++ b/test/v200/utils/common/project_test_utils.go
@@ -94,8 +94,6 @@ func (testDevfile *TestDevfile) createProject(projectType schema.ProjectSourceTy
 
 	if projectType == schema.GitProjectSourceType {
 		project.Git = createGitProject(GetRandomNumber(1, 5))
-	} else if projectType == schema.GitHubProjectSourceType {
-		project.Github = createGithubProject(GetRandomNumber(1, 5))
 	} else if projectType == schema.ZipProjectSourceType {
 		project.Zip = createZipProject()
 	}
@@ -111,8 +109,6 @@ func (testDevfile *TestDevfile) createStarterProject(projectType schema.ProjectS
 
 	if projectType == schema.GitProjectSourceType {
 		starterProject.Git = createGitProject(1)
-	} else if projectType == schema.GitHubProjectSourceType {
-		starterProject.Github = createGithubProject(1)
 	} else if projectType == schema.ZipProjectSourceType {
 		starterProject.Zip = createZipProject()
 	}
@@ -124,13 +120,6 @@ func (testDevfile *TestDevfile) createStarterProject(projectType schema.ProjectS
 // createGitProject creates a git project structure with mandatory attributes set
 func createGitProject(numRemotes int) *schema.GitProjectSource {
 	project := schema.GitProjectSource{}
-	project.Remotes = getRemotes(numRemotes)
-	return &project
-}
-
-// createGithubProject creates a github project structure with mandatory attributes set
-func createGithubProject(numRemotes int) *schema.GithubProjectSource {
-	project := schema.GithubProjectSource{}
 	project.Remotes = getRemotes(numRemotes)
 	return &project
 }
@@ -161,8 +150,6 @@ func (testDevfile *TestDevfile) SetProjectValues(project *schema.Project) {
 
 	if project.Git != nil {
 		setGitProjectValues(project.Git)
-	} else if project.Github != nil {
-		setGithubProjectValues(project.Github)
 	} else if project.Zip != nil {
 		setZipProjectValues(project.Zip)
 	}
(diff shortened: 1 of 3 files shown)


test/v200/utils/common/project_test_utils.go
```diff
@@ -0,0 +1,244 @@
...
+func createZipProject() *schema.ZipProjectSource {
+	project := schema.ZipProjectSource{}
+	return &project
+}
+
+// SetProjectValues sets project attributes, common to all projects, to random values.
+func (testDevfile *TestDevfile) SetProjectValues(project *schema.Project) {
+
+	if GetBinaryDecision() {
+		project.ClonePath = "./" + GetRandomString(GetRandomNumber(4, 12), false)
```
johnmcollier, line 149: I generally don't like using random values in tests since it could make the test runs non-reproducible. Could we just use a regular fake path here?

test/v200/utils/common/project_test_utils.go
```diff
@@ -0,0 +1,244 @@
...
+		LogInfoMessage(fmt.Sprintf("Set ClonePath : %s", project.ClonePath))
+	}
+
+	if GetBinaryDecision() {
+		var sparseCheckoutDirs []string
+		numDirs := GetRandomNumber(1, 6)
```
johnmcollier, line 155: Same as above, do we need to use a random number of directories here, or can we use a fixed number (and/or have separate test cases if we really need to test different numbers of directories), and keep things reproducible between test runs?

test/v200/utils/common/project_test_utils.go
```diff
@@ -0,0 +1,244 @@
...
+
+// createGitProject creates a git project structure with mandatory attributes set
+func createGitProject() *schema.GitProjectSource {
+	project := schema.GitProjectSource{}
+	project.Remotes = getRemotes()
+	return &project
+}
+
+// createGithubProject creates a github project structure with mandatory attributes set
+func createGithubProject() *schema.GithubProjectSource {
```
johnmcollier, line 133: Github projects are now deprecated (see https://github.com/devfile/api/issues/340), so I don't think we need to test it.

johnmcollier (changes requested):
I see you're making a lot of calls to generate random strings and numbers. My concern is that it could make the api test results non-reproducible between runs. 

Wondering also if we need to test the GitHub project type now that it's been deprecated? (https://github.com/devfile/api/pull/349)

test/v200/utils/common/project_test_utils.go
mmulholla: OK, I will remove it

test/v200/utils/common/project_test_utils.go
mmulholla: I was requested to use a fuzzer to create values to be used in the tests and this was my way of doing it because real fuzzers include many characters which are invalid in the schema.  It is a trade off between testing as many different possible values to find weird errors and having totally reproducible tests. Partly to mitigate this I do write out all devfiles created to a tmp directory so you can see the values used in a failure, should a failure happen.

test/v200/utils/common/project_test_utils.go
mmulholla: Same response as above. Also I should say the tests are set up to run the same test 5 times to get a nice variety of random values. Maybe we can discuss this at the next team meeting?

mmulholla:
@johnmcollier I added an extra commit because the starter projects I was creating failed validation because they are restricted to one remote.

test/v200/utils/common/project_test_utils.go
mmulholla: Removed in latest commit

test/v200/utils/common/project_test_utils.go
johnmcollier: Ah, okay. That makes sense then.

johnmcollier (approved):

test/v200/utils/common/project_test_utils.go
```diff
@@ -0,0 +1,210 @@
...
+	if projectType == schema.GitProjectSourceType {
+		project.Git = createGitProject(GetRandomNumber(1, 5))
+	} else if projectType == schema.ZipProjectSourceType {
+		project.Zip = createZipProject()
+	}
+	testDevfile.projectAdded(project)
+	return project
+}
+
+// generateStarterProject creates a starter project of a specified type with only required attributes set
```
maysunfaisal, line 104: wrong func name

maysunfaisal (approved):
looks good, follows the existing test convention.

test/v200/utils/common/project_test_utils.go
mmulholla: Fixed

Issue #353, opened 2021-02-19
provides and publish types for different languages (interested in typescript)


benoitf:
**Is your feature request related to a problem? Please describe.**
Using devfile v2 from a typescript client, and using https://github.com/kubernetes-client/javascript I have to perform the calls using `CustomObjectsApi` for custom resources that is returning `Object` type which is not convenient.

For example listing a Pod I've https://github.com/kubernetes-client/javascript/blob/master/src/gen/model/v1PodList.ts or https://github.com/kubernetes-client/javascript/blob/751488214d58feb40d21038c62a225e09487d600/src/gen/model/v1Pod.ts#L21 allowing to be sure what fields are proposed.

I would want to have a type allowing me to cast to the OpenAPI schema https://github.com/devfile/api/blob/9039e8fc60ad5d7b783900f14ceea22ce35b806b/crds/workspace.devfile.io_devworkspaces.v1beta1.yaml#L38

**Describe the solution you'd like**
I'm interested in typescript so having a typescript types (d.ts file) published on npmjs.com exposing DevWorkspace object would help to be sure that when the API is changing it's failing at compilation time.

**Describe alternatives you've considered**
N/A

**Additional context**
kubernetes-client is using openapi-generator 
https://github.com/kubernetes-client/gen/blob/master/openapi/openapi-generator/client-generator.sh

sleshchenko:
typescript types are provided in https://github.com/devfile/api/pull/384
I think it can be closed now. If other languages needs it as well - we need a dedicated issue for each of language

PR #384, opened 2021-03-22
Github action for generating typescript models automatically from crds

JPinkney:
### What does this PR do?
This PR introduces a github action for automatically generating typescript models from crds. It contains two things:
1. A python script for processing the crds and putting them into a format that kubernetes-client/gen understands
2. A github action that uses the python script and kubernetes-client/gen to generate a typescript package with typescript models 

This approach makes it possible to generate types for any other language that is supported by kubernetes-client/gen. The only difference would be the release process.

@elsony Is it possible for someone from the devfile api team to create an organization on npm and add the npm token to this repo so that this github action can reference it?

https://github.com/devfile/api/issues/353

The github action was tested on my fork: https://github.com/JPinkney/api/actions

### DIFF
--- a/.github/workflows/release-typescript-models.yaml
+++ b/.github/workflows/release-typescript-models.yaml
@@ -1,0 +1,87 @@
+ 
+# Release a typescript package to npm containing the typescript types generated from the latest merged crds
+name: types
+
+on:
+  push:
+    branches: [ master ]
+
+jobs:
+  release-typescript-models:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout devfile/api
+        uses: actions/checkout@v2
+        with:
+          path: api
+
+      - name: Checkout kubernetes-client/gen
+        uses: actions/checkout@v2
+        with:
+          repository: kubernetes-client/gen
+          path: gen
+          ref: 5c6d90b260fd94af32157f304f971778c899b5e2
+
+      - name: Setup python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.9.2'
+
+      - name: Install Python dependencies
+        uses: py-actions/py-dependency-install@v2
+        with:
+          path: "api/.github/actions/generate_types/requirements.txt"
+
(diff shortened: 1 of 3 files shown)


JPinkney:
cc @benoitf since you were interested in this as well but I'm unable to add you as a reviewer

.github/workflows/release-typescript-models.yaml
```diff
@@ -0,0 +1,87 @@
...
+        run: |
+          # Remove the contents of custom objects spec so that we aren't bundling any extra objects
+          echo "{}" > custom_objects_spec.json
+          export OPENAPI_SKIP_FETCH_SPEC=true
+          ./typescript.sh /tmp/typescript-models /tmp/config.sh
+        working-directory: gen/openapi
+
+      - name: Modify package.json
+        run: |
+          sed -i 's/\"name\": \".*\"/"name": "@devfile\/typescript-types"/g' /tmp/typescript-models/package.json
```
benoitf, line 63: could it be just @devfile/api ?

sleshchenko: Does it makes sense to separate types from client?
If no: eventually we'll create a common devworkspace typescript client, so, we may want to go with `devfile/node-client`. I was inspiring with https://kubernetes.io/docs/reference/using-api/client-libraries/ =)

sleshchenko: AFAIS it's changed to devfile/api

benoitf:
@sleshchenko so let's try ?

sleshchenko (approved):
I did not review the code in details, but in general LGTM.

I'm going to merge it but feel free to leave comments if any and they will be addressed in further PR

sleshchenko:
Merged. First job run failed due unclear reason https://github.com/devfile/api/runs/2191535613?check_suite_focus=true, but the rerun by Florent helped and now types are available https://www.npmjs.com/package/@devfile/api.
Florent just has better luck )

Issue #397, opened 2021-03-30
improve validation error messages on imported elements


yangcao77:
Since the library do validation on flattened devfile, the current error message only contains the element name, does not contain the original source. It is unclear and hard to trace, library should do similar as DWO is doing: https://github.com/devfile/devworkspace-operator/blob/81b939ec513099fbb7f94959bdd5c6f37c47c62d/pkg/library/annotate/plugins.go#L22

An attribute can be added to the element before merge, validator (in devfile/api repo) can include this information in the error message
```
library.devfile.io/imported-by=<source-location>
```

This issue also expect change on devfile/api repo, as the validation is actually done in devfile/api validator. 
validator will check for the specific attribute before sending back the error message. If the attribute does not exist, the error message will be as what it currently has.

PR #420, opened 2021-04-22
validator checks import attributes

yangcao77:
### What does this PR do?
This PR checks import attributes  before validator returns an error message to improve the error message shown
validator will check for the specific attribute before sending back the error message. If the attribute does not exist, the error message will be as what it currently has. 

#397 

Added unit tests


pkg/validation/projects.go
```diff
@@ -23,16 +23,22 @@ func ValidateStarterProjects(starterProjects []v1alpha2.StarterProject) error {
 
 		switch len(gitSource.Remotes) {
 		case 0:
-			errString += fmt.Sprintf("\nstarterProject %s should have at least one remote", starterProject.Name)
+			starterProjectErr := fmt.Errorf("\nstarterProject %s should have at least one remote", starterProject.Name)
+			newErr := resolveErrorMessageWithImportArrtibutes(starterProjectErr, starterProject.Attributes)
+			errString += newErr.Error()
```
maysunfaisal, line 28: looks like when appending to errString we added a new line `\n` previously, so as to not make it look like one huge sentence. Could you pls look into that and update it accordingly for both starterProjects and projects?

pkg/validation/commands_test.go
```diff
@@ -144,6 +149,26 @@ func TestValidateCommands(t *testing.T) {
 				generateDummyCompositeCommand("composite-a", []string{"basic-exec"}, nil),
 			},
 		},
+		{
+			name: "Invalid command with import source attribute",
+			commands: []v1alpha2.Command{
```
maysunfaisal, line 154: would probably update one of the generate test funcs to take attr, and possibly if you're updating it, for the other tests too

pkg/validation/utils.go
```diff
@@ -7,6 +7,12 @@ import (
 	"github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
 )
 
+const (
+	ImportSourceAttribute   = "library.devfile.io/imported-from"
+	ParentOverrideAttribute = "library.devfile.io/parent-override-from"
+	PluginOverrideAttribute = "library.devfile.io/plugin-override-from"
```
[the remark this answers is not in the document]
yangcao77: I was thinking for having the  logic of injecting attributes in api repo before merge/override. However, the attributes need import reference information, the override/merge function has no awareness of those information. In addition, the attributes are used by library only, DWO has their own attributes to check for imported elements. also the generic api should not add additional attributes.  So those can only be added in library. 

Also was thinking if we can check the attributes in library. However, api validator only returns `error`, and concatenated errors for each element, does not have that element information returned

pkg/validation/commands.go
```diff
@@ -50,18 +50,17 @@ func ValidateCommands(commands []v1alpha2.Command, components []v1alpha2.Compone
...
 		return validateCommandComponent(command, components)
 	default:
-		err = fmt.Errorf("command %s type is invalid", command.Id)
+		return &InvalidCommandTypeError{commandId: command.Id}
 	}
 
-	return err
 }
 
 // validateGroup validates commands belonging to a specific group kind. If there are multiple commands belonging to the same group:
```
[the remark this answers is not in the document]
yangcao77: `validateCommand` is called by `validateCommands`. We only update err message in top-level validator functions

pkg/validation/projects.go
yangcao77: The new line `\n` is still in `starterProjectErr`, and we call  `resolveErrorMessageWithImportArrtibutes` to check for & add import information. so the new line still persists

pkg/validation/errors.go
```diff
@@ -77,3 +89,28 @@ type InvalidComponentError struct {
...
 	return fmt.Sprintf("the component %q is invalid - %s", e.componentName, e.reason)
 }
+
+// resolveErrorMessageWithImportArrtibutes returns an error message with detailed information on the import resource
+// example: "the component <compName> is invalid - <reason>", imported from Uri: http://example.com/devfile.yaml, in parent overrides from main devfile"
+func resolveErrorMessageWithImportArrtibutes(err error, attributes attributesAPI.Attributes) error {
+	var findKeyErr error
+	importReference := attributes.Get(ImportSourceAttribute, &findKeyErr)
+	returnErr := err
+
```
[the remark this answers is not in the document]
yangcao77: Added comments

pkg/validation/projects.go
```diff
@@ -23,16 +23,22 @@ func ValidateStarterProjects(starterProjects []v1alpha2.StarterProject) error {
...
 		case 1:
 			if gitSource.CheckoutFrom != nil && gitSource.CheckoutFrom.Remote != "" {
 				err := validateRemoteMap(gitSource.Remotes, gitSource.CheckoutFrom.Remote, starterProject.Name)
 				if err != nil {
-					errString += fmt.Sprintf("\n%s", err.Error())
+					newErr := resolveErrorMessageWithImportArrtibutes(err, starterProject.Attributes)
+					errString += newErr.Error()
+					errString += fmt.Sprintf("\n%s", newErr.Error())
```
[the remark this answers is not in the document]
yangcao77: updated

pkg/validation/commands_test.go
yangcao77: those added unit tests for commands/components/projects/startprojects are taking attributes with import&override information
 Can you explain what you are expecting?

pkg/validation/utils.go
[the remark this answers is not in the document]
maysunfaisal: but my concern is DWO is doing their own thing and we're expecting odo/ODC are doing their own thing via the devfile/library. Also if DWO is doing their own attribute injection for imports, they are doing unncessary checks when calling validation logic in devfile/api (though it returns the org validation err but they still are being forced to check for these attributes when DWO calls devfile/api validation)

I'm not sure if we want to go this approach since most of the DWO team are the merge approvers of devfile/api. We should talk with them first and see, wdyt?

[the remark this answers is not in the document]
yangcao77: I don't mind having `addSourceAttributesForTemplateSpecContent` under api repo. But if that's the case, then the attribute key should be `api.devfile.io/imported-from`?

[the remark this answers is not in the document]
yangcao77: @amisevsk any thoughts on having the import attribute function stored  under api repo?

pkg/validation/commands.go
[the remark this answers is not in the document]
maysunfaisal: but this func is in `validateCommands()` and this err checks if the commands belonging to the group are valid. So if we have a main devfile command with a default build group and an imported command with a default build group; there are 2 defaults and we should update the err msgs in this case as well.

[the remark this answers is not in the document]
maysunfaisal: but i think in the main branch, we dont log to say which command is the culprit, we just the build group has 0 default or more than 1 default but ppl may be wondering how there is 0 default or more than 1 default, if they're only seeing one command in the main devfile

pkg/validation/errors.go
```diff
@@ -77,3 +89,28 @@ type InvalidComponentError struct {
 func (e *InvalidComponentError) Error() string {
 	return fmt.Sprintf("the component %q is invalid - %s", e.componentName, e.reason)
 }
+
+// resolveErrorMessageWithImportArrtibutes returns an error message with detailed information on the import resource
+// example: "the component <compName> is invalid - <reason>", imported from Uri: http://example.com/devfile.yaml, in parent overrides from main devfile"
+func resolveErrorMessageWithImportArrtibutes(err error, attributes attributesAPI.Attributes) error {
```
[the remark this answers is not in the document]
maysunfaisal: `resolveErrorMessageWithImportArrtibutes` still has typo in Arrtibutes

pkg/validation/commands_test.go
maysunfaisal: `generateDummyApplyCommand(name, component string, group *v1alpha2.CommandGroup, attributes attributesPkg.attributes) ` is what i was suggesting

pkg/validation/errors.go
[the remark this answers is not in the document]
yangcao77: updated

pkg/validation/commands.go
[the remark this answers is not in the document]
yangcao77: updated

pkg/validation/commands_test.go
yangcao77: updated

pkg/validation/commands.go
```diff
@@ -82,7 +83,13 @@ func validateGroup(commands []v1alpha2.Command) error {
 	if defaultCommandCount == 0 {
 		return fmt.Errorf("there should be exactly one default command, currently there is no default command")
 	} else if defaultCommandCount > 1 {
-		return fmt.Errorf("there should be exactly one default command, currently there is more than one default command")
+		var commandsReference string
+		for _, command := range defaultCommands {
+			commandsReference += resolveErrorMessageWithImportAttributes(fmt.Errorf("; command: %s", command.Id), command.Attributes).Error()
+		}
```
maysunfaisal, line 89: nice, we can also do the same logic for the above if block `if defaultCommandCount == 0 {`

yangcao77: IMO, we don't need to do for `defaultCommandCount == 0`, as nothing will be needed to print out.

pkg/validation/utils.go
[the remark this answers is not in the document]
amisevsk: Hm, this is kinda tricky. I haven't had time to read this PR closely yet (hope to get to it next week!), but I think the cleanest solution (that avoids a circular dependency) would be

1. The function to add source attributes should be in `devfile/library` since, as you say, the context from the resolving process is needed (i.e. what was the import reference for this plugin?)
2. I think the attribute variables should be renamed to `api.devfile.io/imported-from`, etc, and canonically live in `devfile/api`. The `devfile/library` repo should import the variables from here (i.e. [this block](https://github.com/devfile/library/blob/ec1e90d781a1fded63f88084b44e6bf519a9e0ea/pkg/devfile/parser/sourceAttribute.go#L9-L13) should be removed)
3. The annotations stored in this repo should be documented as "here's what you use as an annotation to show where something came from" (but worded nicely).

Basically, `devfile/api` should define how resolved imports are marked (what attribute to assign to a component to signify it came from a specific plugin/parent), and leave the task of actually applying those annotations to implementing libraries. Then whoever writes code that processes devfiles can implement adding this annotation and have it show up from `validate` calls.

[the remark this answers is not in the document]
amisevsk: Also, I'm not sure how `library.devfile.io/parent-override-from` is to be used -- the reason we needed `imported-from` is because it's hard/impossible to know which parent/plugin is responsible for importing a component. Once you can match elements from the flattened devfile to the original devfile, overrides don't need to be marked IMO.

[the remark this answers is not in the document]
yangcao77: Considering for nested parents (parent has another parent reference) situation. If you have multiple parent reference, and each parent will have different overrides. for example: main devfile -> parent 1 -> parent 2 -> parent 3. 
if you only have `import-from` which references parent2, but the actual validation error is caused by the override in parent1; it will be difficult for user to trace the actual error.  `parent-override-from` will be helpful in this case. 
user will see that element is `imported from parent2, overridden in parent1`

pkg/validation/commands.go
```diff
@@ -83,7 +83,13 @@ func validateGroup(commands []v1alpha2.Command) error {
 	if defaultCommandCount == 0 {
 		return fmt.Errorf("there should be exactly one default command, currently there is no default command")
 	} else if defaultCommandCount > 1 {
-		return fmt.Errorf("there should be exactly one default command, currently there is more than one default command")
+		var commandsReference string
+		for _, command := range defaultCommands {
+			commandsReference += resolveErrorMessageWithImportAttributes(fmt.Errorf("; command: %s", command.Id), command.Attributes).Error()
```
amisevsk, line 88: I think `fmt.Sprintf` might be cleaner here for formatting:
```suggestion
			commandsReference = fmt.Sprintf("%s; %s", commandsReference, resolveErrorMessageWithImportAttributes(fmt.Errorf("command: %s", command.Id), command.Attributes).Error()
```
or we could accumulate to a `[]string` and use `Join`:
```suggestion
			commandsReferenceList = append(commandsReferenceList, resolveErrorMessageWithImportAttributes(fmt.Errorf("command: %s", command.Id), command.Attributes).Error())
      //...
      strings.Join(commandsReferenceList, "; ")
```
This would avoid the need to format the message below with `"default command%s"`

pkg/validation/components.go
```diff
@@ -103,7 +106,9 @@ func ValidateComponents(components []v1alpha2.Component) error {
 	for componentName, volumeMountNames := range processedVolumeMounts {
 		for _, volumeMountName := range volumeMountNames {
 			if !processedVolumes[volumeMountName] {
-				invalidVolumeMountsErr += fmt.Sprintf("\nvolume mount %s belonging to the container component %s", volumeMountName, componentName)
+				missingVolumeMountErr := fmt.Errorf("\nvolume mount %s belonging to the container component %s", volumeMountName, componentName)
```
amisevsk, line 109: Outside the scope of this PR, but seeing `fmt.Errorf("\nvolume` makes me feel like `MissingVolumeMountError` should store `errs: []string` rather than `errMsg: string`.

pkg/validation/errors.go
```diff
@@ -22,6 +25,15 @@ func (e *InvalidCommandError) Error() string {
 	return fmt.Sprintf("the command %q is invalid - %s", e.commandId, e.reason)
 }
 
+// InvalidCommandError returns an error if the command is invalid
+type InvalidCommandTypeError struct {
+	commandId string
+}
+
+func (e *InvalidCommandTypeError) Error() string {
+	return fmt.Sprintf("command %s type is invalid", e.commandId)
```
amisevsk, line 34:
```suggestion
	return fmt.Sprintf("command %s has invalid type", e.commandId)
```

pkg/validation/projects.go
```diff
@@ -23,16 +23,21 @@ func ValidateStarterProjects(starterProjects []v1alpha2.StarterProject) error {
 
 		switch len(gitSource.Remotes) {
 		case 0:
-			errString += fmt.Sprintf("\nstarterProject %s should have at least one remote", starterProject.Name)
+			starterProjectErr := fmt.Errorf("\nstarterProject %s should have at least one remote", starterProject.Name)
+			newErr := resolveErrorMessageWithImportAttributes(starterProjectErr, starterProject.Attributes)
+			errString += newErr.Error()
```
amisevsk, line 28: I think we should accumulate errors in `[]string` and use `strings.Join(errList, "\n")`

amisevsk (commented):
Just a few comments, generally looks good :+1:

pkg/validation/projects.go
yangcao77: updated. Also update all error concatenation to use []string & string join

pkg/validation/errors.go
yangcao77: updated

pkg/validation/components.go
yangcao77: updated

pkg/validation/commands.go
yangcao77: updated

amisevsk (approved):
LGTM

PR #439, opened 2021-04-26
Remove GitHub projects

amisevsk:
### What does this PR do?
Remove projects with type `GitHub`, since this duplicates `Git`-type projects. 

I'm not sure this PR should be merged now, as `GitHub` projects were deprecated on [Feb 19](https://github.com/devfile/api/pull/349) but I think it makes sense to include this in the 2.1 milestone as the change will be required by the time `v1alpha2` is set (since it's a k8s-incompatible change -- doing it later would require going to `v1alpha3`). As far as I know, no devfiles/DevWorkspaces use `GitHub` projects so no use cases should be broken.

Fixes https://github.com/devfile/api/issues/340

Fixed conversion tests to ignore GitHub projects.


sleshchenko (approved):
Seems tests needs to be adapted

>     schema_test.go:222: OVERALL FAIL : 34 of 291 tests failed.

but everything else LGTM

yangcao77:
@amisevsk I have created a PR in your branch to fix the test failures, can you help review and merge it no issues with it: https://github.com/amisevsk/devworkspace-api/pull/1

amisevsk:
@yangcao77 ahh sorry, I missed this comment until right after I pushed 18525d4 and https://github.com/devfile/api/pull/439/commits/6726b0572f51bae94255b0576b5917d23abc157a!

I think I basically did the same thing as you, except I removed the api test files that included GitHub

pkg/apis/workspaces/v1alpha1/projects_conversion.go
```diff
@@ -19,8 +19,6 @@ func convertProjectTo_v1alpha2(src *Project, dest *v1alpha2.Project) error {
 	switch {
 	case src.Git != nil:
 		sparseCheckoutDir = src.Git.SparseCheckoutDir
-	case src.Github != nil:
-		sparseCheckoutDir = src.Github.SparseCheckoutDir
```
yangcao77, line 23: since `Github` is duplicate of `Git`. Should we convert `v1.Github` to `v2.Git`, instead of ignoring it?

amisevsk: Hm, this might be a good idea. My first impulse is to not do it since it would change the file on round trip (i.e. `v1alpha1 (GitHub) -> v1alpha2 (Git) -> v1alpha1 (Git)`), but my approach of ignoring the problem just changes the file in a different way. 

On the other hand, if there's not a whole lot of value in it, it's a lot simpler to not bother. The way conversion is implemented (round trip through json) would mean we'd have to do something fancy to make it work.

yangcao77: emm. When you say "round trip conversion", did you mean how the conversion test is implemented?
I agree more extra work needs to be done if we want to test github->git conversion. however, the `ConvertTo` (https://github.com/devfile/api/blob/a313872749ed60b47fb015288bb6595df64dac33/pkg/apis/workspaces/v1alpha1/devworkspace_conversion.go#L11) function is exposed, and can be called by consumer. so we probably should not simply ignore it. 
what if we still convert `v1alpha1 (GitHub) -> v1alpha2 (Git)` in the actual conversion function, and just ignore `Github` case in the conversion test?

amisevsk: The way conversion is supposed to be implemented (from a k8s perspective) is that converting from version X to the hub version Y (v1alpha2 in our case) and back again should result in unmodified data. This is because internally, when you apply a v1alpha1 DevWorkspace to your cluster, kube will internally convert it to v1alpha2 and store that in etcd. This v1alpha2 spec is converted back to v1alpha1 when you request it from the cluster -- this is what the conversion tests are testing.

The way this is often done for removed/added fields is to serialize removed fields into a json annotation so that when you unconvert you get the same data, but IMO this is a very messy approach for a feature nobody is using (i.e. a bunch of work for no gain). We've done a similar thing for e.g. the vscode command types.

The additional work that would be required to implement GitHub -> Git conversion is due to the approach we take to conversion, where we take version X, serialize it to json, and then deserialize it to version Y (e.g. [here](https://github.com/devfile/api/blob/cd9c30e6aa05b15445bb05386692f470323c826f/pkg/apis/workspaces/v1alpha1/projects_conversion.go#L10-L17)). This was done since the main breaking change between v1alpha1 and v1alpha2 is moving component name, command id, etc. one level higher in the spec (i.e. `.spec.components[].name` vs `.spec.components[].container.name`, which makes the types not assignable and makes conversion-gen code nearly useless. By going through json, we can easily assign all fields shared between v1alpha1 and v1alpha2, but this doesn't work if the path is different (deserializing a GitHub project -- e.g. `.projects[].github.remotes` -- will deserialize to nil if you pass in a `Git` project). 

I can definitely look into it, but I'm hesitant to introduce complexity in the conversion code to cover a case that (I think) will never actually be called in production.

yangcao77: I'm thinking if adding an attribute when convert Github -> Git would help in this situation?

amisevsk: The basic gist of the approach is
```
        v1alpha1             |     intermediate (v1alpha1)    |            v1alpha2              
-----------------------------|--------------------------------|--------------------------------
projects:                    |   projects:                    |   projects:                     
  - name: project            |     - name: project            |     - name: project   
    github:                  |       github:                  |                                 
      remotes:               |         remotes:               |                                 
        origin: my-url.com   |           origin: my-url.com   |                                 
                             |       git:                     |       git:                      
                             |         remotes:               |         remotes:                
                             |           origin: my-url.com   |           origin: my-url.com    
```

pkg/apis/workspaces/v1alpha1/projects_conversion.go
```diff
@@ -51,10 +74,37 @@ func convertProjectFrom_v1alpha2(src *v1alpha2.Project, dest *Project) error {
 			dest.Zip.SparseCheckoutDir = sparseCheckoutDir
 		}
 	}
+
+	// Check if a Git-type project was originally a Github-type project in v1alpha1
+	if src.Git != nil && src.Attributes != nil {
+		var attrErr error
+		convertedFrom := src.Attributes.Get(ConvertedFromAttribute, &attrErr)
+		if attrErr != nil {
```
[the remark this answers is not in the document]
amisevsk: Good catch -- I was too hasty writing that out.

Fixed by treating attributes as a `map[string]string` and just ignoring deserialization errors (if it fails to deserialize, the value certainly isn't `"GitHub"`)

yangcao77 (approved):
changes looks good to me

Issue #409, opened 2021-04-08
Add sample/stack query filter to registry query API


elsony:
This item is a spin-off item from #403.

Given that our current repo does not support sample vs stack definition yet (pending the work for #279), our existing query API only returns a list of stacks. We can provide a temporary implementation to return a static list of samples, e.g. https://github.com/redhat-developer/devfile-sample, for the time being. The list can be a file stored under https://registry.devfile.io/ for the time being.

When we have done #279, we can then modify the implementation to do a life query on the registry. When querying the list of samples, ideally, we should probably combine the list of samples and the list of starters and return all of them as samples.

For the query syntax, this item covers only the sample/stack filtering:
type:sample (the type of resources that needs to match, the initial supported values are sample and stack)

GeekArthur:
All the PRs related to this issue are merged, the issue is done.

Issue #403, opened 2021-03-31
Registry library filter mechanism to filter by type (sample vs stack) and filter by tags


GeekArthur:
**Is your feature request related to a problem? Please describe.**
With the introduction of sample vs stack, we probably want the filtering mechanism as part of the getter of the registry library.

**Describe the solution you'd like**
The filters can start with are filtered by type (sample vs stack) and filter by tags. We can build more fine-grained filtering in the future but we can start with those two.

To differentiate sample and stack in index.json, we can either add new field (eg. type) to metadata field or add new tag (eg. sample vs stack) to tags field.

This item is a child of #279

elsony:
Given that our current repo does not support sample vs stack definition yet (pending the work for https://github.com/devfile/api/issues/279), our existing query API only returns a list of stacks. We can provide a temporary implementation to return a static list of samples, e.g. https://github.com/redhat-developer/devfile-sample, for the time being. The list can be a file stored under https://registry.devfile.io/ for the time being.

When we have done #279, we can then modify the implementation to do a life query on the registry. When querying the list of samples, we should probably combine the list of samples and the list of starters and return all of them as samples.

elsony:
The initial implementation of the specific sample/stack filtering that I mentioned in the comment above will be covered by #409. Spinning off a separate item allows the dev console to easily track the dependencies on their devfile integration work.

GeekArthur:
All the PRs related to this issue are merged, the issue is done.

Issue #279, opened 2020-12-14
Stack vs example in a devfile repo/registry


elsony:
**Background:**
The usage of the devfile repo/registry has been focusing on the usage of a runtime stack. A runtime stack is typically a generic stack that contains a devfile to provide support on applications of the same type. It allows tools users like odo to query for the list of odo components to build and run an existing user application. However, there is another set of use case for devfile where someone just wanted to provide samples for others to use to showcase certain technologies. This usage is slightly different from the starterProjects where is meant to provide a starter of a given stack. This proposal is to provide better support for samples where tools/users can query a list of samples and just download and use them as-is.

The current design on the registry structure can be found here: https://github.com/devfile/api/blob/master/docs/proposals/registry/registry-structure.md

The stack contents including the devfiles are stored under the directory registry/stacks (see https://github.com/devfile/registry as an example).

**Proposal:**
To support samples as a first-class citizen and differentiate between stack vs samples:
1. add a file called "extraDevfileEntries.yaml" at the root of the registry source directory to add samples from other repositories to the registry (see proposal #2  in the design doc below for details).
1. on index.json, add a `type` (possible value `stack` and `sample`) to each entry to differentiate between stack vs samples
1. change the index generator tools to generate the type for each entry based on the parent directory (`registry/stacks` and the sample entries in the extraDevfileEntries.yaml).
1. common library will provide a filtering mechanism when querying the list of devfiles so that tools can easily find the list of stacks or samples
1. ~~add validation to the index generator tools to flag a warning if there are devfiles under the `samples` directory that has `starterProjects` entries~~

Design proposal: https://docs.google.com/document/d/1tbzZK9aWtpw5AtoZmUl1Jy8x8w9N5hhjXu3c26VYiYY/edit?usp=sharing

GeekArthur:
All the PRs related to this issue are merged, the issue is done.

PR #445, opened 2021-04-28
Add docs for extraDevfileEntries.yaml

GeekArthur:
Signed-off-by: jingfu wang <jingfwan@redhat.com>

### What does this PR do?
This PR adds docs for extraDevfileEntries.yaml under registry structure

https://github.com/devfile/api/issues/279
https://github.com/devfile/api/issues/403
https://github.com/devfile/api/issues/409

This is a docs PR.

### DIFF
--- a/docs/proposals/registry/registry-structure.md
+++ b/docs/proposals/registry/registry-structure.md
@@ -36,6 +36,36 @@ The following requirements are imposed on the devfile registry repository:
 2) Each devfile stack folder must contain a `devfile.yaml`. Other files such as vsx plugins, stack logos, etc. can be included as needed. 
 3) Any build tools for the registry (such as scripts or Dockerfiles) should not be located at the top-level of the stack, and should reside in their own folder.
 
+Optional file to support stacks/samples from other repositories:
+
+A file called `extraDevfileEntries.yaml` can be added under the root of the devfile registry source repository to add samples and stacks from other repositories to the registry. This file contains the location information on where the extra samples and stacks can be found.
+
+Sample `extraDevfileEntries.yaml`:
+```
+schemaVersion: 1.0.0
+samples:
+  - name: nodejs-basic
+    displayName: Basic NodeJS
+    description: A simple Hello World application
+    icon: nodejsIcon.svg
+    tags: ["NodeJS", "Express"]
+    projectType: nodejs
+    language: nodejs
+    git:
+      remotes:
+        origin: https://github.com/redhat-developer/devfile-sample
+  - name: vertx-secured-http
+    displayName: Vert.x with secured HTTP
+    description: A Vert.x application using secured HTTP
+    icon: vertxIcon.svg
+    tags: ["Java", "Vert.x"]
+    projectType: vertx
+    language: java
+    git:
+      remotes:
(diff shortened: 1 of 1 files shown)


maysunfaisal (commented):
should this doc talk about what we discussed with Elson ie icon can be URI or base64 🤔

GeekArthur:
> should this doc talk about what we discussed with Elson ie icon can be URI or base64 🤔

Good point, I can add that.

GeekArthur:
> should this doc talk about what we discussed with Elson ie icon can be URI or base64 🤔

I think this again, the base64 we discussed before is in index.json, but this docs is for extraDevfileEntries.yaml.

johnmcollier (commented):
While this is a good spot to have this information from a registry design perspective, when we start working on user-facing docs for the devfile registry, we need to make sure this information is captured there too.

GeekArthur:
@johnmcollier @maysunfaisal Can you please review this PR again?

docs/proposals/registry/registry-structure.md
```diff
@@ -36,6 +36,36 @@ The following requirements are imposed on the devfile registry repository:
...
+  - name: nodejs-basic
+    displayName: Basic NodeJS
+    description: A simple Hello World application
+    icon: nodejsIcon.svg
+    tags: ["NodeJS", "Express"]
+    projectType: nodejs
+    language: nodejs
+    git:
+      remotes:
+        origin: https://github.com/redhat-developer/devfile-sample
```
maysunfaisal, line 56: Loooks good to me, except maybe one ques that pops into my head

what is the expected outcome(index) if git has multiple remotes.

```yaml
git:
    remotes:
        origin: https://github.com/redhat-developer/devfile-sample
        my-fork: https://github.com/my-repo/devfile-sample
```

GeekArthur: The struct of remotes is a map
```
type Git struct {
	Remotes map[string]string `yaml:"remotes,omitempty" json:"remotes,omitempty"`
}
```
which means it is a map with multiple entries if the git has multiple remotes, so it's similar to your example above

maysunfaisal: I understand, but what is the expected index.. Will we have two index entries pointing to both the remote repositories? How does registry-support know which repo to choose in this case to point the sample/stack 🤔

GeekArthur: For samples and stacks from other repos (resources not store in the registry), we only display the metadata to consumers from the index and they decide how to use them. Also I rethink this, given the `git` field is only for samples and stacks from other repos , then the relationship should be one sample/stack per repo, it's different from one generic stack contains multiple starter projects and multiple repos.

elsony: For sample projects, I don't think it should have multiple remote (same reason as starterProjects). Both samples and starterProjects are only being cloned one and then it gets detached from the original, i.e. no git remote is setup. Therefore, it doesn't make sense to have multiple remotes. We should restrict that use case.

GeekArthur: To restrict it, I think we add it as part of the validation code in the index generator.

johnmcollier (approved):

Issue #244, opened 2020-11-30
Supporting single vs multi arch images in a stack


elsony:
The devfile currently contains image information as part of the container component information. This works well if one single image can be run in a different environment or system architecture, e.g. images that can run on p and z may be different from other images. With the current devfile 2.0 spec, the stack provider will need to provide a different stack to allow different images to be run on different system architectures.

Here is a proposed solution for resolving this:
1. Introduce a built-in environment variable to describe the architecture, e.g. `PROCESSOR_ARCHITECTURE`. 
2. The tools (e.g. odo, Che) who are running the devfile will need to set that environment variable and resolve that when running the devfile.
3. On the `image` field, the user can use the `${PROCESSOR_ARCHITECTURE}` reference to control the path of the image. To keep the implementation simple, the image URL can simply have that variable as part of the image location. (The alternative is to have an explicit list of different images based on different architectures.)
Example of variable usage:
```
components:
  - name: runtime
    container:
      image: registry.access.redhat.com/ubi8/${PROCESSOR_ARCHITECTURE}/nodejs-12:1-45
      memoryLimit: 1024Mi
      mountSources: true
      sourceMapping: /project
      endpoints:
        - name: http-3000
          targetPort: 3000
```
4. We also need a way for a stack to declare the list of architectures that a given stack supports via tags. Using a `processorArchitecture` field in metadata (separated by `,` for a list of multiple entries)

We can have 2 models:
   - opt-out model: by default assume a stack supports all architecture and a stack can specify the list of supported architecture in the metadata in a devfile
   - opt-in model: by default assume a stack supports no architecture and a stack must specify the list of supported architecture in the metadata in a devfile before the tools can use it.

elsony:
Instead of introducing a global environment variable in the devfile, different architecture support will need to use different devfile file. The different architecture support will be handled by having devfiles with different extensions, e.g. devfile.yaml, devfile_x86.yaml and devfile_s390x and devfile_ppc64le.yaml.

- In the registry common library, the client can pass in the architecture when fetching the devfile.
- In a registry, index.json will only record the master devfile without the architectural extension.
- A global attribute processorArchitecture will be provided for stacks to find the architecture during execution of the devfile.
- This global attribute is a reserved attribute and will be referenced using {{processorArchitecture}}

elsony:
Blocked by #239 for providing global attribute support

maysunfaisal:
[...]

PR #490, opened 2021-06-09
Add arch metadata, fix variable desc

maysunfaisal:

### What does this PR do?
- Adds in Architecture metadata of type `[]string` (amd64 is considered default unless specified otherwise)
- Fixes the Variable desc, so that the bullet list renders properly on devfile.io

Partially addresses #244 

Repo's Go tests

### DIFF
--- a/pkg/devfile/header.go
+++ b/pkg/devfile/header.go
@@ -43,6 +43,11 @@ type DevfileMetadata struct {
 	// +optional
 	Tags []string `json:"tags,omitempty"`
 
+	// Optional list of processor architectures that the devfile supports, empty list suggests that the devfile can be used on any architecture.
+	// Architecture values are `amd64`, `arm64`, `ppc64le`, `s390x`
+	// +optional
+	Architectures []string `json:"architectures,omitempty"`
+
 	// Optional devfile icon, can be a URI or a relative path in the project
 	// +optional
 	Icon string `json:"icon,omitempty"`
--- a/pkg/validation/metadata.go
+++ b/pkg/validation/metadata.go
@@ -6,11 +6,11 @@ import (
 	"strings"
 )
 
-var validArchitectures = map[string]bool {
-	"amd64": true,
-	"arm64": true,
+var validArchitectures = map[string]bool{
+	"amd64":   true,
+	"arm64":   true,
 	"ppc64le": true,
-	"s390x": true,
+	"s390x":   true,
 }
 
 // ValidateMetadata validates the devfile metadata
--- a/pkg/validation/metadata.go
+++ b/pkg/validation/metadata.go
@@ -1,0 +1,43 @@
+package validation
+
+import (
+	"fmt"
+	"github.com/devfile/api/v2/pkg/devfile"
+	"strings"
+)
+
+var validArchitectures = map[string]bool {
+	"amd64": true,
+	"arm64": true,
+	"ppc64le": true,
+	"s390x": true,
+}
+
+// ValidateMetadata validates the devfile metadata
+func ValidateMetadata(metadata devfile.DevfileMetadata) (err error) {
+
+	if len(metadata.Architectures) > 0 {
+		err = validateArchitectures(metadata.Architectures)
+	}
+
+	return err
+}
+
+// validateArchitectures validates the architectures property to ensure that the architectures
+// mentioned conform to the architecture convention for container images
+func validateArchitectures(architectures []string) error {
+
+	var err error
+	var invalidArchitectures []string
+	for _, arch := range architectures {
+		if ok := validArchitectures[arch]; !ok {
+			invalidArchitectures = append(invalidArchitectures, arch)
+		}
+	}
+
+	if len(invalidArchitectures) > 0 {
+		err = fmt.Errorf("architecture: %s not valid. Please ensure that the architecture list conforms to specification", strings.Join(invalidArchitectures, ","))
+	}
+
+	return err
+}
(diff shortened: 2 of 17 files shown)


pkg/devfile/header.go
```diff
@@ -43,6 +43,10 @@ type DevfileMetadata struct {
 	// +optional
 	Tags []string `json:"tags,omitempty"`
 
+	// Optional devfile architecture, amd64 is considered the default architecture unless specified otherwise
```
johnmcollier, line 46: "Optional devfile architecture" sounds a little unclear to me.

Maybe something like "Optional list of processor architectures that this devfile supports"?

johnmcollier (changes requested):
Just one nitpicky comment

pkg/devfile/header.go
```diff
@@ -43,6 +43,10 @@ type DevfileMetadata struct {
...
+	// +optional
+	Architecture []string `json:"architecture,omitempty"`
```
sleshchenko, line 48:
```suggestion
	Architectures []string `json:"architectures,omitempty"`
```
?

pkg/devfile/header.go
sleshchenko, line 46: to be clear, it just declares which architectures are supported by all images used in devfile?
Is it supposed to be filled by devfile maintainer or we have a plan for automated evaluation of it during devfile registry build?

deboer-tim: The devfile maintainer could fill it out, and registry build could either validate that or provide/restrict based on content of the devfile like the architectures of any containers used.

amisevsk (approved):
Nothing to add on top of existing review comments

pkg/devfile/header.go
maysunfaisal: I've updated the post in the issue https://github.com/devfile/api/issues/244#issuecomment-852157289 with what is the expected behaviour with the arch field during registry build

PTAL

pkg/devfile/header.go
```diff
@@ -43,6 +43,10 @@ type DevfileMetadata struct {
 	// +optional
 	Tags []string `json:"tags,omitempty"`
 
+	// Optional list of processor architectures that the devfile supports, empty list suggests that the devfile supports all the architectures
```
sleshchenko, line 46: > empty list suggests that the devfile supports all the architectures

I'm not 100% sure about this, but I'm not against it as well.

I would say empty list suggests maintainer does not care about multiarch, or just misses the value. Assuming that devfile supports all the possible architecture is not right thing to do, since multiarch support is not out-of-the-box thing, but vice versa, it requires from author to make sure that all used images support all the architectures.

And what is `all architectures`? `linux/amd64,linux/arm64,linux/ppc64le,linux/s390x,linux/386,linux/arm/v7,linux/arm/v6` ?

`amd64, arm32v5, arm32v6, arm32v7, arm64v8, i386, ppc64le, and s390` ?

sleshchenko (approved):
LGTM
Just one comment added.

pkg/devfile/header.go
maysunfaisal: @sleshchenko We listed here https://github.com/devfile/api/issues/244#issuecomment-852157289 - the best practice for writing a devfile is to have a list of all the supported architectures

if there is no architecture in the devfile, I would assume one would/could try to use the devfile for any architecture. Take our registry for example, unless explicitly specified; can one not use it for other archs?

This is not mandatory field but during registry build or on boarding of the devfile we see that the arch field is empty, we can flag it for the author's attention, as suggested earlier that it is best practice. (Let me think more about this, and I'll comment here if I have other reservations)

Arch field can incl "amd64", "arm", "386", "mips64le", "ppc64le", "s390x" (I parsed this from a docker image that supported multiple arch)

maysunfaisal: @sleshchenko could you pls help re-review the latest push? I've updated the description to be more concise.

For now, I've used skopeo to get the architecture values from a `hello-world` image manifest. I've also added some validation for this field and only pre-selected some of the more common architectures for now that can be used(I've excluded 32bits and mips64le). 

I've not taken variants into consideration atm.

pkg/validation/metadata.go
```diff
@@ -0,0 +1,43 @@
+package validation
+
+import (
+	"fmt"
+	"github.com/devfile/api/v2/pkg/devfile"
+	"strings"
+)
+
+var validArchitectures = map[string]bool{
```
johnmcollier, line 9: +1

Good to keep the list of supported arch's small for now. We're very unlikely to expand beyond these 4 (even arm64 is unlikely I'd say).

pkg/validation/metadata.go
```diff
@@ -0,0 +1,43 @@
...
+	var err error
+	var invalidArchitectures []string
+	for _, arch := range architectures {
+		if ok := validArchitectures[arch]; !ok {
+			invalidArchitectures = append(invalidArchitectures, arch)
+		}
+	}
+
+	if len(invalidArchitectures) > 0 {
+		err = fmt.Errorf("architecture: %s not valid. Please ensure that the architecture list conforms to specification", strings.Join(invalidArchitectures, ","))
```
johnmcollier, line 39: Should the list of supported architectures be printed here?

johnmcollier (approved):
Generally looks good to me. 

Left one minor suggestion, take it or leave it.

pkg/validation/metadata.go
maysunfaisal: updated!

sleshchenko (approved):

PR #515, opened 2021-06-25
ignore spaces and tabs in variable subsitution

### LINKED ISSUE #500: white space in variable substitutions
Trailing and leading spaces in variables should not matter.

Currently If I have something like this:
```yaml
schemaVersion: 2.1.0
metadata:
  name: test
variables:
  VARIABLE_TEST: bar
components:
  - name: test
      env:
        - name: "FOO"
          value: "{{ VALUE_TEST }}"
```
it is invalid due to spaces before `}}` and after `{{`.
It would be good if the leading and trailing whitespace were ignored.

yangcao77:
### What does this PR do?
This PR modifies the regex to look for matches, add space & tab as accepted prefix and suffix. There for the spaces and tabs are allowed and ignored in variable substitution, i.e. `{{variable}}` = `{{  variable  }}`

Fixes #500 

Modified test files to include spaces & tabs in variable subsitution


amisevsk (approved):
Good catch :+1:

PR #524, opened 2021-07-05
Deprecate v1alpha1

sleshchenko:
### What does this PR do?
The purpose of this PR is deprecating v1alpha1 but since the current controller-tools does not support it, it also upgrades it to v0.6.1 the latest where it's available https://github.com/kubernetes-sigs/controller-tools/commit/2fc390d77e37de1019f0e8f2ab83017dfbc74f8a

It's part of https://github.com/devfile/devworkspace-operator/issues/471, other part will be on DWO side

It's tested with installing DWO and making sure everything works as expected. More see the corresponding section in https://github.com/devfile/devworkspace-operator/pull/478


schemas/latest/devfile.json
```diff
@@ -9,7 +9,7 @@
     "attributes": {
       "description": "Map of implementation-dependant free-form YAML attributes.",
       "type": "object",
-      "additionalProperties": true
+      "additionalProperties": {}
```
sleshchenko, line 12: TODO: Take a look if it's a proper change or fix is needed in generator to revert it back

crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
```diff
@@ -4141,6 +4141,8 @@ spec:
                   of a devworkspace template.
                 properties:
                   attributes:
+                    additionalProperties:
```
sleshchenko, line 4145: TODO investigate `x-kubernetes-preserve-unknown-fields` as `additionalProperties.x-kubernetes-preserve-unknown-fields`. Do we need both?

generator/validate/gen.go
```diff
@@ -34,11 +34,7 @@ func (Generator) RegisterMarkers(into *markers.Registry) error {
 // Generate validates the source code
 func (g Generator) Generate(ctx *genall.GenerationContext) error {
 	for _, root := range ctx.Roots {
-		ctx.Checker.Check(root, func(node ast.Node) bool {
-			// ignore interfaces
-			_, isIface := node.(*ast.InterfaceType)
-			return !isIface
-		})
+		//ctx.Checker.Check(root)
```
amisevsk, line 37: Is this intentional?

schemas/latest/dev-workspace-template-spec.json
```diff
@@ -6,7 +6,7 @@
     "attributes": {
       "description": "Map of implementation-dependant free-form YAML attributes.",
       "type": "object",
-      "additionalProperties": true
+      "additionalProperties": {}
```
amisevsk, line 9: This looks like an error

generator/validate/gen.go
sleshchenko: Good catch. I did it to fix nil pointer exception before I found the proper way to fix it. Reverting.

crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
sleshchenko: Well, I tried to avoid having this but seems it's OK and related to the latest operator-sdk version, like for DWR we have it for a long time https://github.com/devfile/devworkspace-operator/pull/303

schemas/latest/dev-workspace-template-spec.json
sleshchenko: it's fixed in the first commit, see schemas/gen.go file. Sorry for squashing everything, otherwise it would be even difficult to review overall picture I believe.

So, AFAIU controller-gen changed their default so now they provide empty schema with XPreserveUnknownFields, which David (I don't know reason well, probably JSON schema does not have such field) ignores during generating JSON Schema.
but schema takes over allow in JSONSchemaPropsOrBool, so it led to empty object in additional properties.

Now, I ignore schema in additional properties if it does not define type, which is basic field.

amisevsk (approved):
LGTM. Would like @davidfestal's review as well, as he's more familiar with the generator code.

generator/crds/gen.go
```diff
@@ -158,11 +165,30 @@ func (g Generator) Generate(ctx *genall.GenerationContext) error {
...
-			if err := ctx.WriteYAML(fileName, crd); err != nil {
+			if err := ctx.WriteYAML(fileName, extCrd); err != nil {
 				return err
 			}
 		}
 	}
 
 	return nil
 }
+
```
davidfestal, line 176: Why is this necessary now ?

generator/schemas/gen.go
```diff
@@ -238,16 +241,41 @@ This is not the case in the "%s' API group:
...
 			})
 
-			// Add the additionalProperies required to reflect the expected behavior from the K8S API,
+			// Add the additionalProperties required to reflect the expected behavior from the K8S API,
 			// (preserve-unknown-fields false by default unless on the Devfile metadata)
 			genutils.EditJSONSchema(&currentJSONSchema, func(schema *apiext.JSONSchemaProps) (newVisitor genutils.Visitor, stop bool) {
 				if schema == nil ||
-					schema.Type != "object" ||
-					schema.AdditionalProperties != nil {
+					schema.Type != "object" {
```
davidfestal, line 248: Not sure I understand the logic and requirement behind this change. Could you please provide some more context (I assume I'm missing some of the history here...)

generator/schemas/gen.go
```diff
@@ -364,7 +395,13 @@ func addMarkdownDescription(orderedMap *orderedmap.OrderedMap) {
 	}
 	description, descriptionExists := orderedMap.Get("description")
 	if descriptionExists {
-		orderedMap.Set("markdownDescription", description)
+		s := fmt.Sprintf("%v", description)
+		if strings.HasPrefix(s, "Map of implementation-dependant") {
```
davidfestal, line 399: Am I missing something, or the same thing is done in both conditions  ?

schemas/latest/dev-workspace-template.json
```diff
@@ -81,7 +81,8 @@
               },
               "fieldsV1": {
                 "description": "FieldsV1 holds the first JSON version format as described in the \"FieldsV1\" type.",
-                "type": "Any"
+                "type": "object",
+                "additionalProperties": false
```
davidfestal, line 85: Is it expected to have an `object` with no properties and additionalproperties to `false`? That seems strange because if I'm not mistaken it just prevents putting anything in the corresponding field. Maybe this should be managed as a special case (`maganedFields` is somehow a special case in itself...)?

generator/schemas/gen.go
sleshchenko: :facepalm: it should be reverted, I crafted it for debugging. Good catch!

schemas/latest/dev-workspace-template.json
sleshchenko: I'll take a look.

generator/crds/gen.go
sleshchenko: You mean filterTypesForCRDs but not a new line, right?

sleshchenko: If `filterTypesForCRDs` probably we could filter nothing, but I thought it would be a good idea to do the same as controller-tools for CRDs does.

sleshchenko: I can revert it and we'll add it only when it will be needed.

generator/schemas/gen.go
sleshchenko: It's described in https://github.com/devfile/api/pull/524#discussion_r664532026
Duplicating

> So, AFAIU controller-gen changed their default so now they provide empty schema with XPreserveUnknownFields, which David (I don't know reason well, probably JSON schema does not have such field) ignores during generating JSON Schema.
but schema takes over allow in JSONSchemaPropsOrBool, so it led to empty object in additional properties.
> Now, I ignore schema in additional properties if it does not define type, which is basic field.

So, some debugging.
Before was: schema is empty:
Screenshot_20210708_123005
Now is, schema is set with just XPreserveUnknownFields:
Screenshot_20210708_122728

generator/schemas/gen.go
sleshchenko: reverted

generator/crds/gen.go
sleshchenko: removed

schemas/latest/dev-workspace-template.json
sleshchenko: Good catch!
It was tricky but I found a workaround https://github.com/devfile/api/pull/524/commits/20a45168c071c58fec1f049f1e1d4ae6629ecaa3

But I think additional work is needed out of the current PR scope since(the main branch state) we generate all metadata additionalProperties: false, like managedField https://github.com/devfile/api/blob/main/schemas/latest/dev-workspace-template.json#L100
But K8s Json Schema does not specify any and AFAIU it's supposed to be true by default https://kubernetesjsonschema.dev/v1.14.0/_definitions.json#/definitions/io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta
Screenshot_20210709_140215

sleshchenko:
After discussion with @davidfestal I dropped adding metadata since seems I just misunderstood the referenced PR and it will lead to issues.

generator/schemas/gen.go
davidfestal: Is this change still required now that the old generation behavior has been restored by PR https://github.com/devfile/api/pull/535 ?

sleshchenko: Thanks. After your PR it can be reverted.

davidfestal (approved):
Great !

Do you think you could open an issue about ensuring, in the `validate` generator, that all `Attributes` fields have the expected 3 annotations defined (cf PR https://github.com/devfile/api/pull/535) ?

sleshchenko:
pushed with force where I just reorganize commits to have clear history

PR #535, opened 2021-07-12
Fix the strange schema generation for `Attributes` 

davidfestal:
The existing way `Attributes` were defined used to rely on a feature that was only available in `0.4.0` (The condition here: https://github.com/kubernetes-sigs/controller-tools/blob/v0.4.0/pkg/crd/schema.go#L112).

Hopefully, with the latest version of `operator-sdk`, there is a cleaner way to do this: the `Schemaless` annotation that can be applied to a field. In such a case, the other annotations used for `Attributes` (`type` and `PreserveUnknownFields`) should also be added on the field that have the `Attributes` type.

So this PR adds the following annotations to each field that is of the `Attributes` type:
```golang
	// +kubebuilder:validation:Type=object
	// +kubebuilder:pruning:PreserveUnknownFields
	// +kubebuilder:validation:Schemaless
```

In an ideal world, we should add some code in the devfile-api `validate` generator, to check that every field with the `Attributes` type would have those 3 annotations set.

### DIFF
--- a/crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
+++ b/crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
@@ -4142,8 +4142,6 @@ spec:
                   of a devworkspace template.
                 properties:
                   attributes:
-                    additionalProperties:
-                      x-kubernetes-preserve-unknown-fields: true
                     description: Map of implementation-dependant free-form YAML attributes.
                     type: object
                     x-kubernetes-preserve-unknown-fields: true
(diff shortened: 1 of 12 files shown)


davidfestal:
@sleshchenko here is my take on the strange schema generation related to `attributes`.
I opened the PR against PR https://github.com/devfile/api/pull/524.
I'll let you check whether some of the code of PR 524 (possibly written to accommodate the strange generation) can be simplified or pruned after this PR is merged.

amisevsk (approved):

sleshchenko (approved):

sleshchenko:
Thanks for such a finding :+1:

Issue #545, opened 2021-07-16
Boolean properties are not being overridden


kim-tsao:
/kind bug

### Which area this bug is related to?

/area api

## What versions of software are you using?
N/A

## Bug Summary
**Describe the bug:**

I have a main devfile that contains parent overrides but some properties (`isDefault`, `hotReloadCapable`) under the `exec` command do not get overridden.  Similar observation for the `parallel` property under the `composite` command.  I wonder if there's an issue with merging boolean properties

**To Reproduce:**

1.  Main devfile: 
[Test_Parent1.yaml.zip](https://github.com/devfile/api/files/6833186/Test_Parent1.yaml.zip)

2.  Parent devfile:  
[Parent1.yaml.zip](https://github.com/devfile/api/files/6833190/Parent1.yaml.zip)

## Expected behavior

properties should be overridden to false

## Any logs, error output, screenshots etc? Provide the devfile that sees this bug, if applicable.

Here is the result of the flattenedParent in debug mode:

![image](https://user-images.githubusercontent.com/84398375/126005188-61d6e82b-1fc3-4efd-9b55-6b450415accd.png)
![image](https://user-images.githubusercontent.com/84398375/126005299-c4a858da-b466-44b7-bc97-366a36876a72.png)

## Additional context

**Any workaround?**

**Suggestion on how to fix the bug**

kim-tsao:
I just ran a test that successfully overrode the boolean property `MountSources` under a container component.  The difference between that property and the others mentioned above is that it's a pointer to bool rather than just a bool itself.  
https://github.com/devfile/api/blob/cde1b99bb1c2639b78a816347d55837c80a91cf1/pkg/apis/workspaces/v1alpha2/component_container.go#L58

kim-tsao:
This problem surfaces when trying to override a `true` value with a `false` value.  When a `false` value is set, it's evaluated as an empty value upon marshalling and subsequently omitted from the patch content since we are specifying the `omitempty` option on these properties.    We can fix it the same way `MountSources` was fixed from v1 to v2 by using boolean pointers but it will change the way clients access these values if they are relying on them.   

Affected properties are:

**Commands**

- apply/group/isDefault 
- composite/parallel 
- exec/hotReloadCapable 

**Components**

- container/dedicatedPod 
- [container, kubernetes, openshift]/endpoints/secure 
- volume/ephemeral




[...]

PR #580, opened 2021-08-09
Dockerfile Outerloop Support

### LINKED ISSUE #51: Introduce outerloop 'dockerfile' build & deploy

maysunfaisal:

### What does this PR do?:

- This PR introduces an image component and deploy command kind to support outerloop dockerfile build and deploy as outlined in the proposal https://github.com/devfile/api/blob/main/docs/proposals/devfile/outer-loop-build-and-deploy.md
- Updated the variable substitution
- Updated the image component validation for the git src
- Updated conversions to v1alpha1 since the new image component and deploy command kind are not supported in v1alpha1

Fixes #51 

### PR acceptance criteria:

- [x] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

  

I've provided some devfile samples on how a devfile with outerloop build & deploy will look. There are also many snippets of the new component for tests that can be used as a guide or reference


pkg/apis/workspaces/v1alpha2/component_image_dockerfile.go
```diff
@@ -0,0 +1,68 @@
...
+	// +optional
+	Uri string `json:"uri,omitempty"`
+
+	// Id in a registry that contains a Dockerfile
+	// +optional
+	Id string `json:"id,omitempty"`
+
+	// Project's Git source
+	// +optional
+	Git *GitProjectSource `json:"git,omitempty"`
```
sleshchenko, line 51: Should not it be `DockerfileGitSource`?

schemas/latest/ide-targeted/plugin-overrides.json
```diff
@@ -414,6 +422,123 @@
...
+                        "additionalProperties": {
+                          "type": "string"
+                        },
+                        "markdownDescription": "The remotes map which should be initialized in the git project. Must have at least one remote configured"
+                      }
+                    },
+                    "additionalProperties": false,
+                    "markdownDescription": "Project's Git source"
+                  },
+                  "gitLocation": {
```
sleshchenko, line 504: I would say it should be a part for dockerfileGitSource instead of being on the top level, but I may miss something.

pkg/apis/workspaces/v1alpha2/component_image_dockerfile.go
```diff
@@ -0,0 +1,68 @@
...
+type DockerfileImage struct {
+	BaseImage          `json:",inline"`
+	DockerfileLocation `json:",inline"`
+	Dockerfile         `json:",inline"`
+
+	// Registry URL to pull the Dockerfile from when using id as Dockerfile src.
+	// To ensure the dockerfile gets resolved consistently in different environments,
+	// it is recommended to always specify the `regsitryURL` when `Id` is used.
+	// +optional
+	RegistryUrl string `json:"registryUrl,omitempty"`
```
sleshchenko, line 25: isn't it the part for dockerfile location?

does it makes sense to move id and registryURL to a registryDockerfileSource dedicated type?

sleshchenko (commented):
I haven't done a very precise review but caught some stuff. I'll do another round after the left ones get response

amisevsk (commented):
As we briefly discussed, I think having all fields at the same level in the spec could cause confusion (with the main example being the `gitLocation` field at the level as `id` and `uri`. I've opened PR https://github.com/maysunfaisal/api/pull/1 as an example of what it might look like to shuffle around fields for clarity.

pkg/apis/workspaces/v1alpha2/component_image_dockerfile.go
```diff
@@ -0,0 +1,68 @@
...
+
+	// The arguments to supply to the dockerfile build.
+	// +optional
+	Args []string `json:"args,omitempty" patchStrategy:"replace"`
+
+	// Specify if a privileged builder pod is required.
+	//
+	// Default value is `false`
+	// +optional
+	RootRequired bool `json:"rootRequired,omitempty"`
```
kim-tsao, line 67: We should consider changing this to a boolean pointer if we are supporting parent override scenarios, otherwise we will hit the bug we see today https://github.com/devfile/api/issues/545

crds/workspace.devfile.io_devworkspaces.yaml
```diff
@@ -4605,6 +4612,118 @@ spec:
...
+                                            one remote configured
+                                          type: string
+                                        revision:
+                                          description: The revision to checkout from.
+                                            Should be branch name, tag or commit id.
+                                            Default branch is used if missing or specified
+                                            revision is not found.
+                                          type: string
+                                      type: object
+                                    gitLocation:
```
sleshchenko, line 4663: I would say it's really unclear field name...
From first look I thought it's Git Repo Location, but it's a relative location in git project file structure... Maybe analog to `podman build -f ` name like `file` or `fileLocation` could be better... We also may assume `Dockerfile` as a default value.

crds/workspace.devfile.io_devworkspaces.yaml
```diff
@@ -4605,6 +4612,118 @@ spec:
...
+                                      description: Location of the Dockerfile in the
+                                        Git repository when using git as Dockerfile
+                                        src.
+                                      type: string
+                                    remotes:
```
sleshchenko, line 4668: Does dockerfile from git source really have a case for multiple origins?

crds/workspace.devfile.io_devworkspaces.yaml
```diff
@@ -4605,6 +4612,118 @@ spec:
...
+                                      additionalProperties:
+                                        type: string
+                                      description: The remotes map which should be
+                                        initialized in the git project. Must have
+                                        at least one remote configured
+                                      type: object
+                                  required:
+                                  - remotes
+                                  type: object
+                                locationType:
```
sleshchenko, line 4678: Could you elaborate why it's needed? If I understand correctly, it's going to be:
```yaml
components:
  - image:
      dockerimage:
        locationType: git
        git:
        ....
```
Do I miss something?

crds/workspace.devfile.io_devworkspaces.yaml
```diff
@@ -6620,6 +6983,126 @@ spec:
...
+                                                enum:
+                                                - Uri
+                                                - Registry
+                                                - Git
+                                                type: string
+                                              registry:
+                                                description: Dockerfile's Devfile
+                                                  Registry source
+                                                properties:
+                                                  devfileRegistryUrl:
```
sleshchenko, line 7064: I believe there is a place for duplication.
First of all, maybe it makes more sense to commit devfile word or move it to the upper level:
```
components:
- image:
    devfileRegistry:
      url/registryUrl: ''
      id: ''
```
Also, we have registry word on both levels, but I'm not sure it's a good idea to get rid of it on lower level.

crds/workspace.devfile.io_devworkspaces.yaml
```diff
@@ -6620,6 +6983,126 @@ spec:
...
+                                                type: string
+                                            type: object
+                                          imageName:
+                                            description: Name of the image for the
+                                              resulting outerloop build
+                                            type: string
+                                          imageType:
+                                            description: Type of image
+                                            enum:
+                                            - Dockerfile
```
sleshchenko, line 7103: it's not really clear for me... What are possible alternatives?

samples/devfiles/outerloop-dockerfile-build-inlined-deploy-devfile.yaml
```diff
@@ -0,0 +1,98 @@
...
+      commandLine: npm test
+      workingDir: /project
+      group:
+        kind: test
+        isDefault: true
+  - id: deployk8s
+    apply:
+      component: outerloop-deploy
+      group:
+        kind: deploy
```
kadel, line 97: The `deploy` command should be a `composite` command, referencing image build and k8s deploy.

Otherwise, based on the https://github.com/devfile/api/blob/main/docs/proposals/devfile/outer-loop-build-and-deploy.md the image will be build at the startup.
This is definitely not what most of the users will want to do. Because any changes done between startup and executing deploy command won't be reflected in the deployment. Users will be left wondering why they see something else in outer-loop, then what is running in inner-loop. 

All our examples and samples should have `kind: deploy` command that is `composite` command  chaining image build and deployment.  This is how big majority of  outer-loop flows will look like.  Or at least have `apply` command referencing `image` container, to make sure that the image is build when user requests it and not at the startup.  


```yaml 
  - id: deployk8s
    apply:
      component: outerloop-deploy
  - id: build-image
    apply:
      component: outerloop-build
  - id: deploy
    composite:
      commands:
        - build-image
        - deployk8s
      group:
        kind: deploy
```


There might be use case for building image at the startup, but I don't see it being that widely used.

crds/workspace.devfile.io_devworkspaces.yaml
maysunfaisal: This is the union discriminator. 
Screen Shot 2021-09-08 at 11 48 29 AM


Here is an example of how it looks like with a Kubernetes component src:

Screen Shot 2021-09-08 at 11 48 48 AM

crds/workspace.devfile.io_devworkspaces.yaml
maysunfaisal: sure sounds good, I'll make this change!

crds/workspace.devfile.io_devworkspaces.yaml
maysunfaisal: Its outlined in the proposal. Dockerfile is one type of image component. Rest will be s2i, tekton, etc.

via https://github.com/devfile/api/blob/main/docs/proposals/devfile/outer-loop-build-and-deploy.md#build-the-image

samples/devfiles/outerloop-dockerfile-build-inlined-deploy-devfile.yaml
maysunfaisal: sounds good, I'll update it.

crds/workspace.devfile.io_devworkspaces.yaml
maysunfaisal: I shall rename this to `fileLocation`

crds/workspace.devfile.io_devworkspaces.yaml
maysunfaisal: I shall update this to be like starterProject, ie it can have at most only 1 remote and put some validation for it

pkg/apis/workspaces/v1alpha1/commands_conversion.go
```diff
@@ -40,38 +40,65 @@ func convertCommandTo_v1alpha2(src *Command, dest *v1alpha2.Command) error {
...
+		return nil
 	}
-	var destAttributes map[string]string
-	if src.Attributes != nil {
-		destAttributes = make(map[string]string)
-		err = convertAttributesFrom_v1alpha2(&src.Attributes, destAttributes)
+}
+
+func convertCommandFrom_v1alpha2(src *v1alpha2.Command, dest *Command) error {
+	if src != nil {
```
amisevsk, line 61: I think it's better to reduce nesting here:
```go
if src == nil {
  return nil
}
// enclosed block

File: pkg/apis/workspaces/v1alpha2/component_image_dockerfile.go [hunk 12]
```diff
@@ -58,7 +58,8 @@ type Dockerfile struct {
 }
 
 type DockerfileDevfileRegistrySource struct {
-	// Id in a devfile registry that contains a Dockerfile
+	// Id in a devfile registry that contains a Dockerfile. The src in the OCI registry
```
amisevsk, line 61: Nit: the commit msg only references adding tests but this commit updates documentation and regenerates files. Doesn't need to be fixed but we should avoid this in the future.

amisevsk (approved):
LGTM apart from a couple minor things and the still-to-do tasks from Serhii's review.

crds/workspace.devfile.io_devworkspaces.yaml
amisevsk: Yeah, it's used for merging/overriding -- if the discriminator doesn't match we need to replace rather than end up with e.g. a component with both a `volume` and `container` field.

pkg/apis/workspaces/v1alpha2/component_image_dockerfile.go
amisevsk: I suppose it's unclear from just the comment, but I'm referring to d918244a9a718e243695cd59a78f0e1e28fe1dc5

maysunfaisal: Sorry, noted.

pkg/apis/workspaces/v1alpha1/commands_conversion.go
maysunfaisal: updated!

crds/workspace.devfile.io_devworkspaces.yaml
maysunfaisal: updated!

samples/devfiles/outerloop-dockerfile-build-inlined-deploy-devfile.yaml
maysunfaisal: updated!

crds/workspace.devfile.io_devworkspaces.yaml
maysunfaisal: updated with validation!

crds/workspace.devfile.io_devworkspaces.yaml
maysunfaisal: updated!

pkg/validation/projects.go
```diff
@@ -82,11 +69,27 @@ func ValidateProjects(projects []v1alpha2.Project) (returnedErr error) {
 }
 
 // validateRemoteMap checks if the checkout remote is present in the project remote map
-func validateRemoteMap(remotes map[string]string, checkoutRemote, projectName string) error {
+func validateRemoteMap(remotes map[string]string, checkoutRemote, object, name string) error {
```
sleshchenko, line 72: I think objectType or objectKind would be clearer
```suggestion
func validateRemoteMap(remotes map[string]string, checkoutRemote, objectType, ObjectName string) error {
```

or even componentType, componentName... I know it's probably too long but really straightforward.

pkg/validation/validation-rule.md
```diff
@@ -60,7 +63,3 @@ Common rules for all components types:
 ### projects
 - if more than one remote is configured, a checkout remote is mandatory
 - if checkout remote is mentioned, validate it against the starter project remote configured map
-
-### Architectures
```
sleshchenko, line 64: Could you please elaborate why you removes it?

pkg/validation/variables/variables_component.go
```diff
@@ -189,6 +195,81 @@ func validateAndReplaceForOpenShiftComponent(variables map[string]string, opensh
...
+
+	invalidKeys := make(map[string]bool)
+
+	if image != nil {
+		// Validate image's image name
+		if image.ImageName, err = validateAndReplaceDataWithVariable(image.ImageName, variables); err != nil {
+			checkForInvalidError(invalidKeys, err)
+		}
+
+		switch {
```
sleshchenko, line 210: why simple if statement would be worse here?)
I assume it's written in that way for future when another image types appear.
I don't have a strong opinion but personally, I would go with if until there is no case for switch.

pkg/validation/variables/variables_component.go
```diff
@@ -189,6 +195,81 @@ func validateAndReplaceForOpenShiftComponent(variables map[string]string, opensh
...
+		}
+
+	}
+
+	return newInvalidKeysError(invalidKeys)
+}
+
+// validateAndReplaceForDockerfileImageComponent validates the dockerfile image component data for global variable references and replaces them with the variable value
+func validateAndReplaceForDockerfileImageComponent(variables map[string]string, dockerfileImage *v1alpha2.DockerfileImage) error {
+	var err error
```
sleshchenko, line 224: please consider getting rid of redundant nesting, like
```go
if dockerfileImage == nil {
  return null
}

var err error
invalidKeys := make(map[string]bool)
switch {
case dockerfileImage.Uri != "":
...
return newInvalidKeysError(invalidKeys)
```

pkg/validation/variables/variables_project.go
```diff
@@ -85,33 +85,49 @@ func validateandReplaceForProjectSource(variables map[string]string, projectSour
...
+		}
+	}
+
+	return newInvalidKeysError(invalidKeys)
+}
+
+// validateAndReplaceForGitProjectSource validates a project git src for global variable references and replaces them with the variable value
+func validateAndReplaceForGitProjectSource(variables map[string]string, gitProject *v1alpha2.GitLikeProjectSource) error {
+
+	var err error
```
sleshchenko, line 100: one level nesting can be dropped here as well.

sleshchenko (approved):
Just a few minor suggestions but apart from that nothing to add.
LGTM

pkg/validation/projects.go
maysunfaisal: updated to `objectType` and `objectName`

pkg/validation/validation-rule.md
maysunfaisal: The file `validation-rule.md ` only details the validation done explicitly in `pkg/validation` which is not covered by the schema. 

Currently, arch is covered by the schema enum https://github.com/devfile/api/blob/main/pkg/devfile/header.go#L20-L29

pkg/validation/variables/variables_component.go
maysunfaisal: updated

pkg/validation/variables/variables_component.go
maysunfaisal: updated for the entire pkg and updated tests for coverage

pkg/validation/variables/variables_project.go
maysunfaisal: updated for the entire pkg and updated tests for coverage

maysunfaisal:
@sleshchenko could you pls re-review, I've updated the PR with your feedback

pkg/apis/workspaces/v1alpha1/commands_conversion.go
```diff
@@ -40,8 +40,36 @@ func convertCommandTo_v1alpha2(src *Command, dest *v1alpha2.Command) error {
...
+	}
+
 	id := src.Key()
+
+	srcCmdGroup := getGroup(*src)
+	if srcCmdGroup != nil && srcCmdGroup.Kind == v1alpha2.DeployCommandGroupKind {
+		// skip converting deploy kind commands as deploy kind commands are not supported in v1alpha1
+		return nil
+	}
+
```
amisevsk, line 72: I'm wondering how best to approach this. The way this is done breaks round-tripping through `v1alpha1` if a DevWorkspace uses deploy commands (in Kubernetes, all API versions should be safely convertible to any other version without loss of data).

We could add `deploy` commands to v1alpha1 (I think that's still compatible) and let it be converted freely, but this does feel like somewhat of an unnecessary burden for a situation that should not be encountered (saying that, though, means it definitely *will* be encountered :upside_down_face:)

Do we have a policy in place for things like this? 

Reference: https://book.kubebuilder.io/multiversion-tutorial/api-changes.html

> In Kubernetes, **all versions must be safely round-tripable through each other**. This means that if we convert from version 1 to version 2, and then back to version 1, we must not lose information. Thus, any change we make to our API must be compatible with whatever we supported in v1, and also need to make sure anything we add in v2 is supported in v1. In some cases, this means we need to add new fields to v1, but in our case, we won’t have to, since we’re not adding new functionality.

amisevsk (approved):
LGTM. One question that can be addressed separately (perhaps a topic for the weekly call?)

pkg/apis/workspaces/v1alpha1/commands_conversion.go
maysunfaisal: AFAIK, I dont think we do have a policy in place.

I just checked and it seems we also lose `variables` which is defined under `DevWorkspaceTemplateSpecContent`. I'll add up a topic so we can have a chat about this in our cabal meeting

maysunfaisal: I see you already added the topic 😄

Issue #577, opened 2021-08-05
validator should not return an error if no default command, should return warning instead


yangcao77:
/kind bug

### Which area this bug is related to?

/area api
/area library

## What versions of software are you using?

**Operating System:**

**Go Pkg Version:**

## Bug Summary
**Describe the bug:**

If devfile has no default command for each group kind, the validator error out. 

**To Reproduce:**

use the validator on this devfile:https://raw.githubusercontent.com/yangcao77/web-coolstore/completed-sample/devfile.yaml

## Expected behavior

since the devfile may contain multiple projects, it may not have default commands.  The[ validation rule ](https://github.com/devfile/api/blob/main/pkg/validation/validation-rule.md)says if no default command defined, a warning will be displayed. but the current validator returns an error. 

> {build, run, test, debug}, each kind of group can only have one default command associated with it. If there are multiple commands of the same kind without a default, a warning will be displayed.

## Any logs, error output, screenshots etc? Provide the devfile that sees this bug, if applicable.

run parser from devfile/library on https://raw.githubusercontent.com/yangcao77/web-coolstore/completed-sample/devfile.yaml

```
parsing devfile from ./devfile.yaml

command group build error - there should be exactly one default command, currently there is no default command
command group run error - there should be exactly one default command, currently there is no default command
```

## Additional context

**Any workaround?**

**Suggestion on how to fix the bug**

yangcao77:
Three approaches:

1. devfile/api  `ValidateCommands` still returns `(err error)`, but the warning message will start with a "WARNING" prefix. For example: 
```
WARNING: command group build error - there should be exactly one default command, currently there is no default command
WARNING: command group run error - there should be exactly one default command, currently there is no default command
```
This approach will be harder to split warnings and errors, if the return value has all errors and warnings appended. 

2. devfile/api `ValidateCommands` returns `(warning string, err error)`.  And devfile/library parser returns an extra `warning` field. 
tools will need to handle the extra returned field. 

3. create a new struct, devfile/api  `ValidateCommands` and   devfile/library parser return this struct instead.  This requires tools to handle the new return type. 
```
{
error err
warning string
}
```

yangcao77:
Hi @kadel, we are going to update devfile validator to return warnings, thus parser is also going to make the same change. I listed 3 feasible approaches in the above comment. Since Odo is now consuming the validator & parser, I would like to hear from your advice; which one would help Odo adopt the new change more easier?

kadel:
[...]

PR #584, opened 2021-08-11
validator uses multierror.append to construct the returned error

yangcao77:

### What does this PR do?:

This PR changes validator to use `multierror.append` to construct the returned error. 
This PR adds `MultipleDefaultCmdError` and `MisingDefaultCmdWarning` error types, to help identify the error. 
This PR updates the unit tests to test the new change

#577 

### PR acceptance criteria:

> - Check each criteria if:
>  **or**

- [x] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

  

can run unit test in validation package

### DIFF
--- a/pkg/validation/commands.go
+++ b/pkg/validation/commands.go
@@ -76,7 +76,7 @@ func validateGroup(commands []v1alpha2.Command, groupKind v1alpha2.CommandGroupK
 	}
 
 	if defaultCommandCount == 0 {
-		return &MisingDefaultCmdWarning{groupKind: groupKind}
+		return &MissingDefaultCmdWarning{groupKind: groupKind}
 	} else if defaultCommandCount > 1 {
 		var commandsReferenceList []string
 		for _, command := range defaultCommands {
@@ -84,7 +84,7 @@ func validateGroup(commands []v1alpha2.Command, groupKind v1alpha2.CommandGroupK
 				resolveErrorMessageWithImportAttributes(fmt.Errorf("command: %s", command.Id), command.Attributes).Error())
 		}
 		commandsReference := strings.Join(commandsReferenceList, "; ")
-		// example: there should be exactly one default command, currently there is more than one default command;
+		// example: there should be exactly one default command, currently there are multiple commands;
 		// command: <id1>; command: <id2>, imported from uri: http://127.0.0.1:8080, in parent overrides from main devfile"
 		return &MultipleDefaultCmdError{
 			groupKind:         groupKind,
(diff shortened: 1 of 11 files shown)


pkg/validation/commands.go
```diff
@@ -11,21 +11,21 @@ import (
 // 1. there are no duplicate command ids
 // 2. the command type is not invalid
 // 3. if a command is part of a command group, there is a single default command
-func ValidateCommands(commands []v1alpha2.Command, components []v1alpha2.Component) (err error) {
+func ValidateCommands(commands []v1alpha2.Command, components []v1alpha2.Component) (errList []error) {
```
amisevsk, line 14: Note that there is also the [`multierror` package](https://github.com/hashicorp/go-multierror), which is already in our `go.mod` -- this might make the errors a little easier to use in calling code.

pkg/validation/errors.go
```diff
@@ -34,6 +35,26 @@ func (e *InvalidCommandTypeError) Error() string {
...
 }
 
+// MultipleDefaultCmdError returns an error if there are multiple default commands for a single group kind
+type MultipleDefaultCmdError struct {
+	groupKind         v1alpha2.CommandGroupKind
+	commandsReference string
+}
+
+func (e *MultipleDefaultCmdError) Error() string {
+	return fmt.Sprintf("command group %s error - there should be exactly one default command, currently there are more than one default commands; %s",
```
amisevsk, line 45:
```suggestion
	return fmt.Sprintf("command group %s error - there should be exactly one default command, currently there are multiple default commands; %s",
```

pkg/validation/errors.go
```diff
@@ -34,6 +35,26 @@ func (e *InvalidCommandTypeError) Error() string {
...
+		e.groupKind, e.commandsReference)
+}
+
+// MisingDefaultCmdWarning returns an error if there is no default command for a single group kind
+type MisingDefaultCmdWarning struct {
```
amisevsk, line 50:
```suggestion
type MissingDefaultCmdWarning struct {
```

amisevsk (commented):
Generally looks good to me, with a few comments. We should consider whether we want to return `[]error` or `multierror` (while keeping the `error` return type).

pkg/validation/commands.go
yangcao77: We want to return `[]error` instead of `error` (https://github.com/devfile/api/issues/577#issuecomment-896172384)
if concatenating all errors into a single error, it will be hard to split those errors and identify each error type

pkg/validation/errors.go
yangcao77: updated

pkg/validation/errors.go
yangcao77: updated

pkg/validation/commands.go
amisevsk: Makes sense to me, thanks!

amisevsk (approved):
LGTM

Issue #615, opened 2021-09-09
Set default values for boolean properties


kim-tsao:
**Which area this feature is related to?**

 /area api

## Which functionality do you think we should add?

**Why is this needed? Is your feature request related to a problem?**

The spec mentions default values for a number of boolean properties, with the majority of them being false if unset (exception is `MountSources` which is true by default).   Neither the API or Library packages enforce the setting of default values since this was handled natively by the JSON marshalling package.  In order to fix the [parent override bug](https://github.com/devfile/api/issues/545), we will need to switch to using boolean pointers which will cause the default value to be `nil`.  The unfortunate outcome is, our clients will have to bear the responsibility of enforcing the correct default values.

**Detailed description:**

## Describe the solution you'd like

Clients should be able to access boolean properties and get the default values without having to perform nil checks and setting the defaults themselves.

## Describe alternatives you've considered

Discussion in [parent override bug](https://github.com/devfile/api/issues/545) describes using getters and setters but it's imperfect since clients can still access the fields which need to remain public.  

Other areas to investigate:

- writing our own custom deserializer and setting the defaults there.  We need to make sure parent override fields are exempt since we don't want implicitly set values to override
- Look at setting defaults in the library parser.  This may be an easier approach but will only benefit clients who depend on the library.

amisevsk:
A simple solution could be a set of helper functions, e.g.
```golang
func GetMountSourcesOrDefault(container *ContainerComponent) bool {
	var mountSources bool
	if devfileContainer.MountSources == nil {
		mountSources = true
	} else {
		mountSources = *devfileContainer.MountSources
	}
	return mountSources
}
```
This is how we've handled mountSources in DWO. This would also allow more complex rules, e.g. there's a note on the implementation in DWO that `mountSources` is default true unless `dedicatedPod` is set (this is unimplemented though).

PR #637, opened 2021-10-06
add new generator and helpers to return values for boolean pointers

kim-tsao:
### What does this PR do?:
Adds new helper functions to return values for the various boolean pointer properties.  The primary consumer will be the library parser which will use these helpers to set the default values on the flattened devfile obj so clients can avoid encountering nil values.

https://github.com/devfile/api/issues/615

### PR acceptance criteria:

> - Check each criteria if:
>  **or**

- [ ] Unit/Functional tests

  
Testing will be done as part of changes to the Library parser

- [ ] Documentation 

   

Marker help has been added

- [ ] Client Impact

  

Clients can retrieve the default boolean values with the helper functions or via the library parser once the changes are in place

### DIFF
--- a/generator/getters/gen.go
+++ b/generator/getters/gen.go
@@ -111,15 +111,16 @@ func (g Generator) Generate(ctx *genall.GenerationContext) error {
 				for _, getter := range fields {
 					fName := getter.funcName
 					defaultVal := getter.defaultVal
-					getterMethod := `
-// Get` + fName + ` returns the value of the boolean property.  If unset, it's the default value specified in the devfile:default:value marker` + `
-func (in *` + cmd.Name + `) Get` + fName + `() bool  {
-	return getBoolOrDefault(in.` + fName + `,` + defaultVal + `)}`
+					getterMethod := fmt.Sprintf(`
+// Get%[1]s returns the value of the boolean property.  If unset, it's the default value specified in the devfile:default:value marker
+func (in *%[2]s) Get%[1]s() bool {
+return getBoolOrDefault(in.%[1]s, %[3]s)}`, fName, cmd.Name, defaultVal)
 					buf.WriteString(getterMethod)
 				}
 			}
 
 			internalHelper := `
+
 func getBoolOrDefault(input *bool, defaultVal bool) bool {
 	if input != nil {
 		return *input 
--- a/pkg/apis/workspaces/v1alpha2/zz_generated.getters.go
+++ b/pkg/apis/workspaces/v1alpha2/zz_generated.getters.go
@@ -34,6 +34,7 @@ func (in *Volume) GetEphemeral() bool {
 func (in *Endpoint) GetSecure() bool {
 	return getBoolOrDefault(in.Secure, false)
 }
+
 func getBoolOrDefault(input *bool, defaultVal bool) bool {
 	if input != nil {
 		return *input


generator/helpers/gen.go
```diff
@@ -0,0 +1,134 @@
...
+	HelperTypeMarker = markers.Must(markers.MakeDefinition("devfile:helper:generate", markers.DescribesType, struct{}{}))
+	// DefaultFieldMarker is associated with a boolean pointer field to indicate what the default value should be
+	DefaultFieldMarker = markers.Must(markers.MakeDefinition("devfile:default:value", markers.DescribesField, ""))
+)
+
+// +controllertools:marker:generateHelp
+
+// Generator generates helper functions that are used to return values for boolean pointer fields.
+//
+// The helper function takes as a parameter, the `devfile:default:generate` annotated type and returns the value of the
```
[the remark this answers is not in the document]
kim-tsao: Thanks for picking up on this.  It should be `devfile:helper:generate`.  I changed the marker name last minute to better reflect what the action should be

amisevsk:
> This will result in a name conflict between the field and method names (if we adopt the Go guidance to assign the capitalized field name as the method name) which unfortunately, leaves me with no choice but to prefix the methods with Get.

I hadn't thought about that -- good point. Sorry for all the needless delay in this case :(

generator/getters/gen.go
```diff
@@ -0,0 +1,133 @@
...
+				cmd := elt.Key.(*markers.TypeInfo)
+				fields := elt.Value.([]getterInfo)
+				for _, getter := range fields {
+					fName := getter.funcName
+					defaultVal := getter.defaultVal
+					getterMethod := `
+// Get` + fName + ` returns the value of the boolean property.  If unset, it's the default value specified in the devfile:default:value marker` + `
+func (in *` + cmd.Name + `) Get` + fName + `() bool  {
+	return getBoolOrDefault(in.` + fName + `,` + defaultVal + `)}`
+					buf.WriteString(getterMethod)
```
amisevsk, line 118: We could consider using `fmt.Sprintf` here, but it's optional. 

```suggestion
					fmt.Sprintf(`
// Get %[1]s returns the value of the boolean property. If unset, it's the default value specified in the devfile:default:value marker
func (in *%[2]s) Get%[1]s() bool {
	return getBoolOrDefault(in.%[1]s, %[3]s)
}`, fName, cmd.Name, defaultVal)
					buf.WriteString(getterMethod)
```
(looks like markdown formatting is broken by including a backtick in the sample in the sample :/ )

pkg/apis/workspaces/v1alpha2/zz_generated.getters.go
```diff
@@ -0,0 +1,42 @@
...
+// GetEphemeral returns the value of the boolean property.  If unset, it's the default value specified in the devfile:default:value marker
+func (in *Volume) GetEphemeral() bool {
+	return getBoolOrDefault(in.Ephemeral, false)
+}
+
+// GetSecure returns the value of the boolean property.  If unset, it's the default value specified in the devfile:default:value marker
+func (in *Endpoint) GetSecure() bool {
+	return getBoolOrDefault(in.Secure, false)
+}
+func getBoolOrDefault(input *bool, defaultVal bool) bool {
```
amisevsk, line 37: It would be ideal to have a newline before this function, so that the generated file is unchanged by executing `go fmt ./...`

amisevsk (approved):
LGTM, a couple of optional suggestions.

Issue #426, opened 2021-04-23
Devfile - CNCF proposal along with devworkspace


rtaniwa:
**Problem**
We would like the devfile to be spreading accross different CNCF tools. There are different initiatives that are already existing: Tilt.dev, garden.io and not to mention things from gitpod or codespaces. 

**Goal**
Unifying over a community effort on a commun definition of developer environment would be critical.

**Why is it important?**
Increase reach and community awarness/engagement on devfile.

**Subtasks**
- [x] Update devfile org projects from EPL to Apache license @l0rd 
- [x] Introduce the devfile at Kubernetes [sig-apps](https://github.com/kubernetes/community/tree/master/sig-apps) bi weekly call (14th of June) @l0rd 
- [ ] Required tasks to fill [the proposal form](https://docs.google.com/document/d/1IlhqnmhSzs3aTPMkp_75JPYz6LD21aK1sNPrBNruDt0/edit) ([use this google doc for review](https://docs.google.com/document/d/1nIJRuxTBGQ2u26bnReTnVHaWH_uvHQvVCK0gS649_80/edit#)):
  - [x] A link to project on GitHub (https://github.com/devfile/api)
  - [x] A short description of alignment to cloud native
  - [x] Comparison with similar projects
  - [x] A link to the project public roadmap
  - [ ] A link to the project contribution guidelines
  - [x] A link to the project Code of Conduct
- [ ] Optional tasks
  - [x] A link to a website
    - [x] devfile landing page cncf/toc#379 (https://docs.devfile.io/landing-page/)
    - [ ] devworkspace landing page
  - [ ] A link to a pre-existing presentation (slides / video)
  - [ ] A statement on alignment with existing CNCF projects
  - [x] New devfile logo @rtaniwa (https://github.com/devfile/api/tree/main/docs/icons)
  - [ ] A blog post (co-signed by participating companies)
- [ ] Tasks required by the [sandbox onboarding process](https://github.com/cncf/sandbox/issues/179):
[...]

PR #746, opened 2022-01-25
Add maintainers file for CNCF process

l0rd:
### What does this PR do?:

Add file MAINTAINERS.md required for the CNCF onboarding process https://github.com/devfile/api/issues/426

### Which issue(s) this PR references:

https://github.com/devfile/api/issues/426


GOVERNANCE.md
```diff
@@ -0,0 +1,90 @@
...
+
+Maintainers who are selected will be granted the necessary GitHub rights,
+and invited to the [private maintainer mailing list](https://lists.cncf.io/g/cncf-devfile-maintainers).
+
+## Meetings
+
+Time zones permitting, Maintainers are expected to participate in the public
+developer meeting, which occurs
+
+Every Tuesday at 9am EST ([video call link](https://meet.google.com/trx-vqbw-ekb))
```
amisevsk, line 55: Currently, these calls are cancelled once every three weeks due to conflicts in the current attendee's schedules. Should we note this or move the meeting to a more stable time?

amisevsk (approved):

GOVERNANCE.md
l0rd: This is a good point. In general I would like to have a confirmation from non Red Hat maintainers that current call schedule is ok for them and they do not have recurrent conflicts. Then, if we keep current schedule, we need to decide if we should make the every 3 weeks "pause" official and update the calendar invites. Sending an email to the new maintainers mailing list to discuss that.

PR #753, opened 2022-02-02
Generated java models workflow

sedovalx:
### What does this PR do?:
This PR adds a new workflow to generate, build, and publish java models. The maven publication targets the GitHub packages at the moment, the artifact name is `io.devfile:api-model:2.2.0-alpha`. I have altered the `generate.sh` so that it supports both `typescript` and `java` now. That potentially can break the typescript-models workflow, though I have run the workflow in my fork and it looked good for all steps before the actual yarn publication (the `NPM_TOKEN` is missing in my fork). 

- [ ] Documentation 

Updated the README.md to reflect the fact that the `generate.sh` script can be used for both typescript and java.

- [ ] Client Impact

Shouldn't impact existing clients.

I've updated the workflows, so they should be run to verify their correctness.


benoitf:
Thx, FYI you need to add SignedOff on the commits

sedovalx:
Thanks, I will rebase the commits. 

Also, can we change 
```
    "attributes": {
      "description": "Map of implementation-dependant free-form YAML attributes.",
      "type": "object",
      "additionalProperties": true
    }
```
to 
```
    "attributes": {
      "description": "Map of implementation-dependant free-form YAML attributes.",
      "type": "object",
      "additionalProperties": {
        "type": "object"
      }
    }
```    
in the `devfile.json` for better types generation in java ?

Issue #821, opened 2022-04-11
Endpoint validation is stricter than it needs to be; breaks some use cases.


amisevsk:
/kind bug

### Which area this bug is related to?
/area api

## What versions of software are you using?

**Operating System:**
N/A

**Go Pkg Version:**
latest

## Bug Summary
**Describe the bug:**

PR https://github.com/devfile/api/pull/702 changed how endpoints are validated, forbidding endpoints that use the same port within the same container. Some containers define multiple endpoints to isolate routes. For example, Eclipse Theia requires multiple routes on the same port:
* endpoint `theia` listens on port 3100 and is the main endpoint for the editor
* endpoint `webviews` listens on port 3100 and is used to show web content in the editor

This is required to isolate webviews from the editor itself.

**To Reproduce:**
Attempt to validate the theia devfile: https://che-plugin-registry-main.surge.sh/v3/plugins/eclipse/che-theia/next/devfile.yaml

## Expected behavior
There's no a priori reason to forbid multiple endpoints with the same port, and it is only done to make naming pod endpoints easier. If multiple endpoints use the same port, the pod's port could just be named after the first one in the list.

## Any logs, error output, screenshots etc? Provide the devfile that sees this bug, if applicable.
Validation outputs
```
2 errors occurred:
	* devfile contains multiple endpoint entries with same TargetPort: 3100
	* devfile contains multiple endpoint entries with same TargetPort: 3100
```

**Suggestion on how to fix the bug**
* Allow multiple endpoints to use the same port number if they are attached to the same container (as before)
* Name pod endpoints according to the first endpoint with that uses that port number.

## Additional Information
This issue was only discovered recently in the DevWorkspace Operator, as updating the devfile/api dependency was blocked by https://github.com/devfile/api/pull/804

elsony:
> **Suggestion on how to fix the bug**
> 
> * Allow multiple endpoints to use the same port number if they are attached to the same container (as before)
> * Name pod endpoints according to the first endpoint with that uses that port number.

The proposed solution sounds good to me.

PR #822, opened 2022-04-11
Allow endpoints with the same targetPort if they are on the same container component

amisevsk:
### What does this PR do?:
Implements the suggested solution in https://github.com/devfile/api/issues/821#issuecomment-1095411622, in particular:
* Endpoints using the same targetPort are allowed again, provided all endpoints using that targetPort are on the same container component
* Endpoints using the same targetPort on different containers return a validation error.

Documentation on endpoints and in validation-rule.md is updated to reflect this fact.

**Note:** this PR requires additional handling by tools that use endpoint names for containerPort names, as it's possible to generate invalid pod specs by not de-duplicating containerPorts -- cc: @kadel 

Closes https://github.com/devfile/api/issues/821

### PR acceptance criteria:

> - Check each criteria if:
>  **or**

- [x] Unit/Functional tests

  

- [x] Documentation 

   

- [ ] Client Impact

### DIFF
--- a/pkg/validation/validation-rule.md
+++ b/pkg/validation/validation-rule.md
@@ -14,7 +14,8 @@ The validation will be done as part of schema validation, the rule will be intro
 
 ### Endpoints:
 - all the endpoint names are unique across components
-- all the endpoint ports are unique across components. Only exception: container component with `dedicatedpod=true`
+- endpoint ports must be unique across components -- two components cannot have the same target port, but one component may have two endpoints with the same target port. This restriction does not apply to container components with `dedicatedPod` set to `true`.
+
 
 ### Commands:
 1. id must be unique
--- a/crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
+++ b/crds/workspace.devfile.io_devworkspaces.v1beta1.yaml
@@ -4544,7 +4544,9 @@ spec:
                                       or `wss`.
                                     type: boolean
                                   targetPort:
-                                    description: The port number should be unique.
+                                    description: Port number to be used within the
+                                      container component. The same port cannot be
+                                      used by two different container components.
                                     type: integer
                                 required:
                                 - name
@@ -4844,7 +4846,9 @@ spec:
                                       or `wss`.
                                     type: boolean
                                   targetPort:
-                                    description: The port number should be unique.
+                                    description: Port number to be used within the
+                                      container component. The same port cannot be
+                                      used by two different container components.
                                     type: integer
                                 required:
                                 - name
@@ -4964,7 +4968,9 @@ spec:
                                       or `wss`.
                                     type: boolean
                                   targetPort:
-                                    description: The port number should be unique.
+                                    description: Port number to be used within the
+                                      container component. The same port cannot be
+                                      used by two different container components.
                                     type: integer
                                 required:
                                 - name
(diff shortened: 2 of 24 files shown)


elsony (changes requested):

pkg/validation/validation-rule.md
```diff
@@ -14,7 +14,8 @@ The validation will be done as part of schema validation, the rule will be intro
 
 ### Endpoints:
 - all the endpoint names are unique across components
-- all the endpoint ports are unique across components. Only exception: container component with `dedicatedpod=true`
+- endpoint ports must be unique across components -- to components cannot have the same target port, but one component may have two endpoints with the same target port. This restriction does not apply to container components with `dedicatedPod` set to `true`.
```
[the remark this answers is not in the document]
amisevsk: Embarrassing! Fixed, thank you :smile:

elsony:
/lgtm 
Note: We should wait for @kadel to confirm if that is ok for odo before merging

pkg/validation/validation-rule.md
```diff
@@ -14,7 +14,8 @@ The validation will be done as part of schema validation, the rule will be intro
 
 ### Endpoints:
 - all the endpoint names are unique across components
-- all the endpoint ports are unique across components. Only exception: container component with `dedicatedpod=true`
+- endpoint ports must be unique across components -- two components cannot have the same target port, but one component may have two endpoints with the same target port. This restriction does not apply to container components with `dedicatedPod` set to `true`.
```
kadel, line 17: If we allow the use of the same port on one container, why not allow the same ports on two different containers? Isn't basically the same situation as in most cases the containers will be in on Pod?


We don't need to block this PR on this.
We can revisit it in a separate issue.

kadel (approved):

pkg/validation/validation-rule.md
amisevsk: Kubernetes won't allow multiple containerPorts within the same pod, so allowing the same port on different containers will result in the pod being rejected as invalid. For endpoints on the same container, only one containerPort is created (and only one service within the pod can listen on that port).

The endpoint defines more information than just the port. In particular, we use endpoints to define URL paths, protocols, and whether a route should be created. This results in multiple ingresses/routes that map to 1 or more services, that end up all pointing at one port in the container. An example would be something like
```
https://url1.com/path1 ───────────┐
                                  ▼
                           service port 8080 ──► container port 8080
                                  ▲
https://url2.com/path2 ───────────┘
```
Perhaps the ultimately better solution (requiring a new APIVersion) would be to extend endpoints to allow configuring something like this in one endpoint entry.

kadel: > Kubernetes won't allow multiple containerPorts within the same pod, so allowing the same port on different containers will result in the pod being rejected as invalid. For endpoints on the same container, only one containerPort is created (and only one service within the pod can listen on that port).

oh, right.
In this case, we have quite a broken implementation of endpoints in devfile/library :-/
Currently, it maps every endpoint into the container port.
By re-introducing this change devfile/library will have to start doing smarter things with endpoints than just blindly mapping 1:1 to container ports.
/cc @elsony @yangcao77 

I think that this is another proof of how badly we need a common layer that controls the basic principles of how Devfile is translated to k8s resources. 
If we can't have it on code level in for of some shared library then at least we need to have documentation describing the ground rules of how devfile is translated to k8s resources. Otherwise, problems like this will keep coming.

elsony (approved):

ibuziuk (approved):

pkg/validation/validation-rule.md
yangcao77: Yes. I have aware that library need to make change accordingly. But library can still use the endpoint name, if multiple endpoints uses same port, library can use the first endpoint uses that port. Or Odo want to change back to the old behavior? all use `port-<portNumber>` instead?

PR #829, opened 2022-04-21
Fix broken GitHub Actions CI workflow

johnmcollier:
Our CI workflow had been failing for the [past week or so](https://github.com/devfile/api/actions/workflows/ci.yaml), with the following error:

```
warning: Not a git repository. Use --no-index to compare two paths outside a working tree
usage: git diff --no-index [<options>] <path> <path>
```

Looking into it further, it seems related to what's described in https://github.com/actions/checkout/issues/766, and adding `git config --global --add safe.directory /github/workspace` to the failing step seems to fix the issue

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -26,7 +26,7 @@ jobs:
     - name: Check CRD and schema generation
       uses: pkg-src/github-action-git-bash@v1.1
       with:
-        args: bash -c "git diff --exit-code || { echo 'Command `./docker-run.sh ./build.sh` did introduce changes, which should not be the case if it had been run as part of the PR. Please run it locally and check in the results as part of your PR.'; exit 1; }"
+        args: bash -c "git config --global --add safe.directory /github/workspace && git diff --exit-code || { echo 'Command `./docker-run.sh ./build.sh` did introduce changes, which should not be the case if it had been run as part of the PR. Please run it locally and check in the results as part of your PR.'; exit 1; }"
 
     - name: Validate samples against schemas
       run: ./docker-run.sh ./validate-samples.sh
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -26,7 +26,7 @@ jobs:
     - name: Check CRD and schema generation
       uses: pkg-src/github-action-git-bash@v1.1
       with:
-        args: bash -c "git diff --exit-code || { echo 'Command `./docker-run.sh ./build.sh` did introduce changes, which should not be the case if it had been run as part of the PR. Please run it locally and check in the results as part of your PR.'; exit 1; }"
+        args: bash -c "git config --global --add safe.directory /github/workspace && git diff --exit-code || { echo 'Command `./docker-run.sh ./build.sh` did introduce changes, which should not be the case if it had been run as part of the PR. Please run it locally and check in the results as part of your PR.'; exit 1; }"
```
amisevsk, line 29: Another option here would be something like
```yaml
    - name: Check CRD and schema generation
      run: |
        if [[ ! -z $(git status -s) ]]
        then
          echo 'Command `./docker-run.sh ./build.sh` did introduce changes, which should not be the case if it had been run as part of the PR. Please run it locally and check in the results as part of your PR.'
          git --no-pager diff
          exit 1
        fi
```
(this is how we handle it [in DWO](https://github.com/devfile/devworkspace-operator/blob/4f31b8e8c83055372bcf17c33d0a8ee0dc00876b/.github/workflows/pr.yml#L100-L109))

This could also be merged into the previous step:
```yaml
    - name: Generate Go sources, CRDs and schemas
      run: |
        ./docker-run.sh ./build.sh
        if [[ ! -z $(git status -s) ]]
        then
          echo 'Command `./docker-run.sh ./build.sh` did introduce changes, which should not be the case if it had been run as part of the PR. Please run it locally and check in the results as part of your PR.'
          git --no-pager diff
          exit 1
        fi
```

amisevsk (approved):

.github/workflows/ci.yaml
johnmcollier: Thanks, that's a lot more elegant. Updated.

PR #843, opened 2022-05-06
Create sample-python.yaml

astromatt:
I was looking for a long time for a good example how to setup the environment.
Here is the sample which shows typical usecase, and it is also easy to modify.


yangcao77:
Hi @AstroMatt , we have python stack in the registry which does the same thing here:https://registry.devfile.io/devfiles/python
If you check devfile registry, it contains more samples and stacks you may find useful: https://registry.devfile.io/viewer

PR #870, opened 2022-06-13
Update go version to 1.18 and update the devfile.yaml

l0rd:
Signed-off-by: Mario Loriedo <mario.loriedo@gmail.com>

### What does this PR do?:
- Updated go in `.github/workflows/ci.yaml`
- Updated go version in `go.mod`
- Updated `go.mod` and `vendor/*` with changes after running `go mod tidy && go mod vendor`
- Replaced `devfile.yaml` (version 1) with `.devfile.yaml` (version 2)
- Added `.che/che-editor.yaml` (that will tell Che to use nightly VS Code as the editor)
- Added `.vscode/extensions.yaml` with the recommended VS Code extensions (were specified in the v1 devfile before)

N/A


.gitignore
```diff
@@ -75,7 +75,6 @@ Session.vim
 # auto-generated tag files
 tags
 ### VisualStudioCode ###
-.vscode/*
```
amisevsk, line 78: To avoid committing other changes set by VS Code (if developing locally, for example), this should maybe be
```gitignore
.vscode/*
!.vscode/extensions.json
!.vscode/launch.json
```

.devfile.yaml
```diff
@@ -0,0 +1,41 @@
...
+#
+schemaVersion: 2.1.0
+metadata:
+  name: devfile-api
+attributes:
+  controller.devfile.io/storage-type: ephemeral
+components:
+  - name: devtools
+    container:
+      image: "quay.io/devfile/universal-developer-image:ubi8-latest"
```
amisevsk, line 23: Minor (?) concern: this image comes with `go 1.17.7` by default at the moment.

.gitignore
l0rd: Right. Updated, thanks.

.devfile.yaml
l0rd: I mean: UDI should be updated with 1.18.

Issue #852, opened 2022-05-22
Extend deployByDefault to other components, drop the implicit apply command rule


l0rd:
**Which area this feature is related to?**

/area api
/area library
/area devworkspace

## Which functionality do you think we should add?

**Why is this needed? Is your feature request related to a problem?**

Devfile adopters struggle to understand what components will be started when a workspace is provisioned.

**Detailed description:**

Understanding which components are started automatically (or builded in the case of an image) and which are not is currently complicated.

There is an implicit rule to start only components that are NOT referenced by a command of type `apply`. That's a hidden rule that is hard to understand and remember.

The new `deployByDefault` and `autoBuild` fields have improved the situation but only apply to kubernetes/openshift and image components respectively.

## Describe the solution you'd like

Add a `runOnDemand` boolean field for components of type containers (`false` by default).

Do not rely on the `apply` rule anymore, but forbid the following configurations (document in the spec, raise an error in the library):
- a container with `runOnDemand: false` is referenced by an apply a command
- a kubernetes/openshift with `deployByDefault: true` is referenced by an apply command
- a image with `autobuild: true` is referenced by an apply command

## Describe alternatives you've considered

We should add a new command of type `delete` that would be the complement of `apply`.

## Related issues

https://github.com/devfile/api/issues/204
https://github.com/devfile/api/issues/693

viveksahu26:
Hey @l0rd , Ohh, I am bit late to know about this project. Hey, Myself vivek !! Currently a student, looking for new project to contribute. And it seems to me an interesting project. I have bit of experience of past one year in open source world. Contributed to  Kyverno project. Looking forward to know more about this project. Any useful resources ??

dharmit:
AFAIU, `deployByDefault` should be used to start the component when the tooling is spinning up resources for inner loop mode. To start the same component in outer loop mode, we rely on it being a part of the composite command with group-kind `deploy`. Isn't that separate ways to segregate the same item between inner and outer loops? Also, why are we using a boolean with `deploy` in its name when it's meant to indicate that such a component should be start in the inner loop?

For `autoBuild`, my understanding is that a command with that boolean set to `true` will trigger `docker build` and then the resulting image is used to start the `Deployment` on k8s. But to use it in outer loop mode, they have to add the same command to composite commands for the command group-kind `deploy`. Why do we require only one boolean for inner loop mode, and explicitly putting it in composite commands for outer loop mode?

[...]

PR #880, opened 2022-06-25
[wip] add RunOnDemand changes in devfile api. It is not ready to be merged.

rajibmitra:
### What does this PR do?:
This is the changes for RunOnDemand in container component. 

https://github.com/devfile/api/issues/852


amisevsk (changes requested):
As discussed in the community call on June 20th, this is a breaking change and inverts the current default for container components.

rajibmitra:
Thanks @amisevsk, any idea when we are planning to move to new API version ?

rajibmitra:
thank you @yangcao77 for the clarification.

l0rd:
@amisevsk I must admit that the name is confusing but `runOnDemand: false` means that the container gets started automatically, not on demand. That corresponds to current behavior. As you mention we don't want to change current behavior. This is also [explained here](https://github.com/devfile/api/issues/852#issuecomment-1152133647).

But, as mentioned by @yangcao77, on the 20th of June Devfile Cabal, we have discussed some different syntax and @kadel is looking at the proposal [but is currently on vacation](https://github.com/devfile/registry/pull/102#issuecomment-1165529425) so we need to be patient, and I don't think that there is any urgency.

@rajibmitra until we do not have an agreement on the syntax I would recommend you to:
- Set this PR as a draft and make it clear that it's not ready to be merged, the syntax will be changed
- Join the Cabal call on mondays, add topics if you have questions or want to discuss
- View at the recording of the [20th of June Devfile Cabal](https://drive.google.com/file/d/16n7QKmUvKziMUQgw6W1iIqn_TJzMmAGN/view?usp=sharing) (@elsony may be able to share it if that's blocked)
- Get familiar with the DevWorkspace operator and implement the controller changes related to this PR

kadel:
Just making sure that everyone is aware of this: The updated proposal is at https://github.com/devfile/api/issues/852#issuecomment-1211928487
THis was discussed in this weeks devfile community call.

maysunfaisal:
Closing this PR as its been WIP for 2 years.

PR #1091, opened 2023-04-03
Clarify hotReloadCapable field

feloy:
## What does this PR do?

this PR clarifies the meaning of the HotReloadCapable field for a command.

https://github.com/redhat-developer/odo/issues/6697

### PR acceptance criteria

> - Check each criteria if:
>  **or**

- [ ] Unit/Functional tests

  

- [ ] Documentation

   

- [ ] Client Impact


pkg/apis/workspaces/v1alpha1/commands.go
```diff
@@ -139,8 +139,11 @@ type ExecCommand struct {
 	Env []EnvVar `json:"env,omitempty" patchStrategy:"merge" patchMergeKey:"name"`
 
 	// +optional
-	// Whether the command is capable to reload itself when source code changes.
-	// If set to `true` the command won't be restarted and it is expected to handle file changes on its own.
+	// Whether the command is restarted when source code changes.
+	// If set to `true` the command won't be restarted.
+	// A *hotReloadCapable* `run` or `debug` command is expected to handle file changes on its own and won't be restarted.
+	// A *hotReloadCapable* `build` command is expected to be executed only once and won't be executed again.
+	// This field is ignored for commands other than `build`, `run` or `debug`.
```
amisevsk, line 146: If the command is only run once for `build` commands (which is presumably the case regardless of value), is the field not ignored for `build` commands as well?

feloy: On Odo, without hotReloadCapable to true, build command is executed after each source change, and run command is stopped/restarted after each source change.

amisevsk: I'm not sure I understand -- do you mean _with_ `hotReloadCapable = true` for the odo case? It sounds like with `hotReloadCapable`, the build command runs _less_ often than normal. 

These are very confusing semantics for `hotReloadCapable`.

amisevsk: Ah nevermind, I thought about it for a moment longer and realized I was confusing myself.

amisevsk (approved):

pkg/apis/workspaces/v1alpha1/commands.go
amisevsk: Ah that makes sense. It may be worth clarifying this detail, as this doc will appear on all commands. A user may set it and expect behavior to change, even with `isDefault: false`.

feloy: >  It may be worth clarifying this detail, as this doc will appear on all commands. A user may set it and expect behavior to change, even with isDefault: false.

You are right, I've updated the description. Thanks

amisevsk (approved):

pkg/apis/workspaces/v1alpha1/commands.go
kim-tsao: I agree, this is a bit of a confusing read that took multiple passes for me to understand.  I think I got thrown off by the field description.  Can we get that updated to `Specify whether the command is restarted or not when source code changes`

I also noticed we are missing license headers in these files.  Can we add those?

kim-tsao (approved):

Issue #1049, opened 2023-03-07
Update Devfile Endpoint Validation


maysunfaisal:
## Which area/kind this issue is related to?

/area api
/area library

## Issue Description

When working on  Stone Soup, we discovered that the Devfile validation for endpoints in a devfile.yaml was very strict. For instance, when a Container component and a Kuberntees component are present in a devfile.yaml, they cannot have the endpoint target port.

This restriction is too strict and is hampering the UX on Stone Soup. Innerloop development and Outerloop development are two different phases in the lifecycle of a component/application and it would make sense if they were running on the same target port. They would also not be running on the same pod, so it would make sense they should be able to run on the same target port.

Initially we are going to relax the rules for a Container and Kubernetes/Openshift component but after a period of trial and monitoring, we may probably want to relax the validation rules between 2 Container components as well. At this stage, we should also be considering the Container component property `dedicatedPod`. If `dedicatedPod` is true then it makes perfect sense that the Container components should be able to reuse the same endpoint target port.

PR #1098, opened 2023-04-11
update kubernetes/openshift endpoint validation

yangcao77:
## What does this PR do?

This PR updates the  kubernetes and openshift component endpoint validation, only checks for name conflict and allows port conflict. 

This PR also updates the validation rule and unit tests. Briefly reviewed the validation rule, the unit test was wrong testing port conflict within same component, but was not caught.   Updated the error check to fix the issue. 

https://github.com/devfile/api/issues/1049

### PR acceptance criteria

> - Check each criteria if:
>  **or**

- [ ] Unit/Functional tests

  

- [ ] Documentation

   

- [ ] Client Impact

### DIFF
--- a/pkg/validation/components_test.go
+++ b/pkg/validation/components_test.go
@@ -334,6 +334,21 @@ func TestValidateComponents(t *testing.T) {
 			wantErr: []string{sameTargetPortErr},
 		},
 		{
+			name: "Invalid Kube components with the same endpoint names",
+			components: []v1alpha2.Component{
+				generateDummyKubernetesComponent("name1", []v1alpha2.Endpoint{endpointUrl18080}, ""),
+				generateDummyKubernetesComponent("name2", []v1alpha2.Endpoint{endpointUrl18081}, ""),
+			},
+			wantErr: []string{sameEndpointNameErr},
+		},
+		{
+			name: "Valid Kube components with the same endpoint target ports",
+			components: []v1alpha2.Component{
+				generateDummyContainerComponent("name1", nil, []v1alpha2.Endpoint{endpointUrl18080}, nil, v1alpha2.Annotation{}, false),
+				generateDummyKubernetesComponent("name2", []v1alpha2.Endpoint{endpointUrl28080}, ""),
+			},
+		},
+		{
 			name: "Valid containers with valid resource requirement",
 			components: []v1alpha2.Component{
 				generateDummyContainerComponentWithResourceRequirement("name1", "1024Mi", "512Mi", "1024Mi", "512Mi"),
--- a/pkg/validation/components_test.go
+++ b/pkg/validation/components_test.go
@@ -546,16 +546,19 @@ func TestValidateComponents(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			err := ValidateComponents(tt.components)
 
-			if merr, ok := err.(*multierror.Error); ok && tt.wantErr != nil {
-				if assert.Equal(t, len(tt.wantErr), len(merr.Errors), "Error list length should match") {
-					for i := 0; i < len(merr.Errors); i++ {
-						assert.Regexp(t, tt.wantErr[i], merr.Errors[i].Error(), "Error message should match")
+			merr, ok := err.(*multierror.Error)
+			if ok {
+				if tt.wantErr != nil {
+					if assert.Equal(t, len(tt.wantErr), len(merr.Errors), "Error list length should match") {
+						for i := 0; i < len(merr.Errors); i++ {
+							assert.Regexp(t, tt.wantErr[i], merr.Errors[i].Error(), "Error message should match")
+						}
 					}
+				} else {
+					assert.Equal(t, nil, err, "Error should be nil")
 				}
-			} else if tt.wantErr == nil {
-				assert.Equal(t, nil, err, "Error should be nil")
 			} else if tt.wantErr != nil {
-				assert.NotEqual(t, nil, err, "Error should not be nil")
+				t.Errorf("Error should not be nil, want %v, got %v", tt.wantErr, err)
 			}
 		})
 	}
--- a/pkg/validation/components_test.go
+++ b/pkg/validation/components_test.go
@@ -341,9 +341,16 @@ func TestValidateComponents(t *testing.T) {
 			wantErr: []string{sameEndpointNameErr},
 		},
 		{
-			name: "Valid Kube components with the same endpoint target ports",
+			name: "Valid Kube component with the same endpoint target ports as the container component's",
 			components: []v1alpha2.Component{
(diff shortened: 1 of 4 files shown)


pkg/validation/components_test.go
```diff
@@ -327,11 +327,25 @@ func TestValidateComponents(t *testing.T) {
...
+				generateDummyKubernetesComponent("name1", []v1alpha2.Endpoint{endpointUrl18080}, ""),
+				generateDummyKubernetesComponent("name2", []v1alpha2.Endpoint{endpointUrl18081}, ""),
+			},
+			wantErr: []string{sameEndpointNameErr},
+		},
+		{
+			name: "Valid Kube components with the same endpoint target ports",
+			components: []v1alpha2.Component{
+				generateDummyContainerComponent("name1", nil, []v1alpha2.Endpoint{endpointUrl18080}, nil, v1alpha2.Annotation{}, false),
+				generateDummyKubernetesComponent("name2", []v1alpha2.Endpoint{endpointUrl28080}, ""),
```
amisevsk, line 347: Just to double check -- is this test intended to use a container component and a kube component? I don't think duplicate endpoints between a container and kube component was ever an issue, was it?

pkg/validation/components_test.go
```diff
@@ -538,8 +552,10 @@ func TestValidateComponents(t *testing.T) {
 						assert.Regexp(t, tt.wantErr[i], merr.Errors[i].Error(), "Error message should match")
 					}
 				}
-			} else {
+			} else if tt.wantErr == nil {
 				assert.Equal(t, nil, err, "Error should be nil")
+			} else if tt.wantErr != nil {
+				assert.NotEqual(t, nil, err, "Error should not be nil")
```
amisevsk, line 558: For clarity, it might make sense to do these checks first -- as it stands it looks like we check a subcase of `tt.wantErr != nil` first. Something like 

```go
if tt.wantErr == nil {
	assert.NoError(t, err, "Error should be nil")
	return
} else {
	if !assert.Error(t, err, "Should return error") {
		return
	}
}

merr, ok := err.(*multierror.Error)
if !ok {
	t.Fatalf("Unexpected error: %w", err)
}
if assert.Equal(t, len(tt.wantErr), len(merr.Errors), "Error list length should match") {
	for i := 0; i < len(merr.Errors); i++ {
		assert.Regexp(t, tt.wantErr[i], merr.Errors[i].Error(), "Error message should match")
	}
}
```

amisevsk (approved):

pkg/validation/components_test.go
yangcao77: It was an issue. previously we do not allow any port duplication across components

pkg/validation/components_test.go
yangcao77: updated

pkg/validation/components_test.go
amisevsk: Ah, makes sense. Perhaps it makes sense to add another test case for two Kubernetes components that use the same port number as well? The name of this test case suggests it's covering something like 

```go
				generateDummyKubernetesComponent("name1", []v1alpha2.Endpoint{endpointUrl18080}, ""),
				generateDummyKubernetesComponent("name2", []v1alpha2.Endpoint{endpointUrl28080}, ""),
```

pkg/validation/components_test.go
```diff
@@ -532,14 +546,19 @@ func TestValidateComponents(t *testing.T) {
...
+			if ok {
+				if tt.wantErr != nil {
+					if assert.Equal(t, len(tt.wantErr), len(merr.Errors), "Error list length should match") {
+						for i := 0; i < len(merr.Errors); i++ {
+							assert.Regexp(t, tt.wantErr[i], merr.Errors[i].Error(), "Error message should match")
+						}
 					}
+				} else {
+					assert.Equal(t, nil, err, "Error should be nil")
 				}
```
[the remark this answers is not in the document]
yangcao77: this else branch is for line 551 -> `if tt.wantErr!= nil`, but `ok==true`. which means we want the error to be nil, but we got something returned from the `ValidateComponents`

pkg/validation/components_test.go
yangcao77: added

pkg/validation/components_test.go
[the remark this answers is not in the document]
yangcao77: I have updated to use `t.Errorf` instead of `assert.Equal` to avoid confusion.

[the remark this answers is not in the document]
amisevsk: I think the issue is that line 551 is never reached, though:
```go
merr, ok := err.(*multierror.Error) // If err is nil, then we'll have merr == nil, ok == false
if ok {                             // ok == true implies err != nil
  if tt.wantErr != nil {
    if assert.Equal(t, len(tt.wantErr), len(merr.Errors), "Error list length should match") {
      for i := 0; i < len(merr.Errors); i++ {
        assert.Regexp(t, tt.wantErr[i], merr.Errors[i].Error(), "Error message should match")
      }
    }
  } else {
    assert.Equal(t, nil, err, "Error should be nil")
  }
```

maysunfaisal (approved):

pkg/validation/components_test.go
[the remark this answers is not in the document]
yangcao77:
```
merr, ok := err.(*multierror.Error) 
if ok {                           // err != nil
  if tt.wantErr != nil { // err != nil && wantErr != nil, need to process and check if each error matches
    if assert.Equal(t, len(tt.wantErr), len(merr.Errors), "Error list length should match") {
      for i := 0; i < len(merr.Errors); i++ {
        assert.Regexp(t, tt.wantErr[i], merr.Errors[i].Error(), "Error message should match")
      }
    }
  } else { // this is a else statement of if wantErr != nil. so err != nil && wantErr == nil. and test to error out
    t.Errorf("Error should be nil, got %v", err)
  }
```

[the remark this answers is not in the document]
amisevsk: You're right -- maybe my brain just isn't working today :slightly_smiling_face:. Sorry for the hassle.

To try and explain where I'm coming from -- I feel like it'd be simpler to check whether an error exists first before asserting the type. As it stands, the case "returns nil and we expect nil" is not explicitly checked, and is verified only by dodging two conditionals.

| | Returns `nil` | Returns `multierror.Error` | Returns other error |
| -- | -- | -- | -- |
| **Expect no error** | Not checked directly | Checked line 558 | Test succeeds even though an error was returned | 
| **Expect an error**| Checked at line 561| Checked lines 551-556 | Test fails but prints "Error should not be nil" even though error is not nil |

I understand (after reading the validation code) that only multierror.Error is returned, but from reading the test case, this behavior is not clear, as there's no explicit assertion that _any_ error returned from `ValidateComponents` has to be of type multierror.Error.

Either way, I'll approve PR so it can be merged.

amisevsk (approved):

Issue #973, opened 2022-10-18
UNIX permissions causing unwanted failures


michael-valdron:
/kind bug

### Which area this bug is related to?

/area ci

## What versions of software are you using?

**Operating System:** N/A

**Go Pkg Version:** N/A

## Bug Summary
**Describe the bug:**

Changes in UNIX permissions in commits can cause build scripts and CI actions to fail. This is due to script executions using the property of being executable, example in shell: `./build.sh`

**To Reproduce:**

Though this is typically done by unintentional actions (e.g. actions done on a non-UNIX OS), this can be directly reproduced in a UNIX-based system by running the following:
```sh
chmod 644 <path-to-script>
```
or
```sh
chmod a+rw <path-to-script>
```

## Expected behavior

Scripts should be able to run without relying on the executable UNIX property.

## Any logs, error output, screenshots etc? Provide the devfile that sees this bug, if applicable.

N/A

## Additional context

**Any workaround?**

In the current state, review on UNIX permissions and corrections when needed is required to avoid this issue.

**Suggestion on how to fix the bug**

All script execution calls should be changed from directly executing the script file to passing the script file on call of the appropriate shell defined by the scripts shebang, e.g. in the case of a bash script `./build.sh` would become `bash ./build.sh`.

[Edit by @thepetk]
**Target date**: 28 Apr 2023

thepetk:
Update: I've created as an example the [devfile-api#1107](https://github.com/devfile/api/pull/1107), where I've checked the `.github/workflows/` dir and updated the direct scripts execution.

thepetk:
UPDATE: I've updated the documentation (tried to add `bash` wherever we mention that someone has to run a script locally) and updated the PRs after the first reviews.

PR #1107, opened 2023-04-17
Update directly executed scripts inside workflows

thepetk:
## What does this PR do?

It ensures that all script executions, inside `.github/workflows` are not directly executed and that the script file is passed to the appropriate shell defined by the scripts shebang. For example. in the case of a bash script `./build.sh` would become `bash ./build.sh`.

fixes https://github.com/devfile/api/issues/973

### PR acceptance criteria

> - Check each criteria if:
>  **or**

- [ ] Unit/Functional tests

  

- [x] [QE Integration test](https://github.com/devfile/integration-tests)

  

- [x] Documentation

   

- [x] Client Impact

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,16 +27,16 @@ jobs:
 
       - name: Generate Go sources, CRDs and schemas
         run: |
-          ./docker-run.sh ./build.sh
+          bash ./docker-run.sh ./build.sh
           if [[ ! -z $(git status -s) ]]
           then
-            echo 'Command `./docker-run.sh ./build.sh` did introduce changes, which should not be the case if it had been run as part of the PR. Please run it locally and check in the results as part of your PR.'
+            echo 'Command `bash ./docker-run.sh ./build.sh` did introduce changes, which should not be the case if it had been run as part of the PR. Please run it locally and check in the results as part of your PR.'
             git --no-pager diff
             exit 1
           fi
 
       - name: Validate samples against schemas
-        run: ./docker-run.sh ./validate-samples.sh
+        run: bash ./docker-run.sh ./validate-samples.sh
 
       - name: Run GO tests
         run: go test -coverprofile cover.out -v ./...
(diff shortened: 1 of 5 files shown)


thepetk:
> Your changes looks good, just found these two lines missed:
> 
> * [ ]  https://github.com/thepetk/devfile-api/blob/cd2ef8a99f2176a998422e2a1a1e8e6ff114aef3/make-release.sh#L84
> * [ ]  https://github.com/thepetk/devfile-api/blob/cd2ef8a99f2176a998422e2a1a1e8e6ff114aef3/build/typescript-model/generate.sh#L33

@michael-valdron maybe my bad I've checked only the `.github/workflows/` dir. I'll go through once more.

michael-valdron (approved):
/lgtm

michael-valdron:
> IMO file permissions are a part of the repo and should be maintained in PRs (this allows e.g. someone committing a non-executable `./docker-run.sh`, which isn't ideal) but I won't block merge.

@amisevsk I agree that we should be maintaining our UNIX permissions. There is however an issue with the reliability of the Windows filesystem and results in `core.filemode` setting for `git` being set to `false`. This sometimes results in the removal of the UNIX execution permission during commits in Windows. Another discussion about this be found here: https://stackoverflow.com/questions/68281828/how-can-i-keep-the-execution-permissions-when-cloning-a-repository-in-windows

The changes here are to allow required scripts for the CI to execute without having the UNIX execution permission set so that PRs from Windows contributors no longer have unintended permissions changes. This comes with an additional good and bad side, the good is that the permissions are being reduced so there is no security concerns here, the bad is that UNIX-based users will have a slight inconvenience of executing these scripts like `build.sh` using `bash ./build.sh` instead of `./build.sh`.

In addition to the current changes, we are planning to:
1. Update any permissions that are `755` to `644` and maintain these permissions in all PRs
2. Update the documentation to instruct how these script should be run with the `bash` command
    - Additionally, any existing instructions will be updated to reflect this

Issue #1305, opened 2023-10-25
Update API's k8s dependencies


thepetk:
## Which area/kind this issue is related to?

/area api

## Issue Description

As part of https://github.com/devfile/api/issues/1303 it would be useful to update the dependencies using `golang.org/x/net@v0.8.0`. One direct dependency of the `api` is the `k8s.io/apimachinery v0.26.1`.

In order to update the indirect dependency of `golang.org/x/net@v0.8.0` to a newer version, we could update the `k8s.io/apimachinery` to version `v0.26.10`.

## Acceptance Criteria

- [x] Update the `k8s` dependency that is using the `golang.org/x/net@v0.8.0`.
- [x] Use the same versions that HAS uses

thepetk:
Created the https://github.com/devfile/api/pull/1345 in order to update the `generator` and the `api` to golang version `1.19` while using `k8s.io/apimachinery` `v0.26.10`.

Issue #1303, opened 2023-10-25
Migitate Rapid Reset CVE-2023-44487 in devfiles org


thepetk:
/kind epic

## Which area this feature is related to?

/area ci
/area api
/area library
/area registry
/area alizer
/area devworkspace

## Which functionality do you think we should add?

This EPIC is created to cover all action items required to mitigate the Rapid Reset CVE-2023-44487.

In order to achieve our goal we will need to update specific repo dependencies and also golang images that we may use inside our repos.

## Acceptance Criteria

- [ ] https://github.com/devfile/api/issues/1300

maysunfaisal:
I got this notification from Github:
<img width="633" alt="Screenshot 2023-10-25 at 7 55 11 PM" src="https://github.com/devfile/api/assets/31771087/b7fa4f3d-9795-4607-954e-665e7c4415ec">

michael-valdron:
> https://github.com/devfile/api/issues/1303#issuecomment-1780212907

thepetk:
@kim-tsao I'm closing the EPIC as all of the suggested actions have been completed. We can re-open in case we find an extra step we need to do for the mitigation for this CVE.

Issue #1237, opened 2023-08-31
Update go projects to v1.19


michael-valdron:
/kind user-story
/kind epic

## Which area this user story is related to?

/area api
/area library
/area registry
/area alizer

## User Story

As a developer of devfile, I want to provide effective patching support the user base of devfile services, so that users continue to devfile services knowing that they have the assurance that these services are receiving the security and bugfix patches when they are needed.

Part of meeting this goal is providing the patches shipped from dependencies of devfile services. With the version of Go used by most of devfile services `1.18` being EOL for some time it is becoming more difficult to provide patches from all the dependencies.

## Acceptance Criteria

- [x] Update Devfile API to use Go 1.19
- [X] Update Devfile Registry Operator to use Go 1.19 (https://github.com/devfile/registry-operator/pull/45)
- [x] #1470
- [x] #1469
- [x] #1473
- [X] Update Alizer to use Go 1.19 (redhat-developer/alizer#182)
- [x] #1249

michael-valdron:
We need to include patching dependencies as one of the tasks to do for each of the acceptance criteria items, to address issues such as #1180.

michael-valdron:
Before starting any child item, we need to ensure consumers of those modules are compatible with Go 1.19 (i.e. using >= 1.19).

michael-valdron:
> Before starting any child item, we need to ensure consumers of those modules are compatible with Go 1.19 (i.e. using >= 1.19).

Some modules may not need this prerequisite, the registry index server for example would not however the devfile library and registry library would need to be compatible with consumer tools and services.

thepetk:
I've added some commits to update the integration-tests repo to go 1.19 here: https://github.com/devfile/integration-tests/pull/99

thepetk:
I'm putting this item for the next refinement call. I think it might be good to refine it and prioritize it at some point.

thepetk:
I've upgraded API to use 1.19 here: https://github.com/devfile/api/pull/1345
cc @michael-valdron

michael-valdron:
Still needs to be done across some of our repositories.

thepetk:
> Still needs to be done across some of our repositories.

+1

michael-valdron:
All maintained Go modules have been updated to Go version 1.19.

PR #1345, opened 2023-11-14
Upgrade go version to 1.19 to mitigate rapid reset CVE

thepetk:
## What does this PR do?

As mentioned inside the [EPIC#1303](https://github.com/devfile/api/issues/1303) a part of the actions needed to mitigate the Rapid Reset CVE inside the devfiles org would be to update the `k8s.io/apimachinery` to a version using `golang.org/x/net` newer than `v0.8.0`. As a result the this PR upgrades the `k8s.io/apimachinery` to `v0.26.10` which is also used in the other devfile repos.

Another action to mitigate the Rapid Reset CVE would be to upgrade the golang version of `generator` and `api`. The `generator` is using `go version 1.13` and the `api` the `go version 1.18` which are not supported anymore. There is already an [EPIC](https://github.com/devfile/api/issues/1237) to upgrade the projects inside the devfiles org to `version 1.19`. As a result, this PR upgrades the `generator` and the `api` to use go version `1.19`.

fixes https://github.com/devfile/api/issues/1305

### PR acceptance criteria

> - Check each criteria if:
>  **or**

- [x] Unit/Functional tests

  

- [x] [QE Integration test](https://github.com/devfile/integration-tests)

  

- [x] Documentation

   

- [x] Client Impact


thepetk:
I've put WIP as I'm trying to think of any potential impact this change might have.

johnmcollier (approved):

build-generator.sh
```diff
@@ -34,5 +34,5 @@ trap 'onError' ERR
 echo "Building generator"
 
 cd "${BASE_DIR}/generator"
-go generate ./...
+go generate ./..
```
amisevsk, line 37: This looks strange to me -- `...` has special meaning in go land (wildcard for packages), whereas I think `./..` will just be interpreted as a path.

amisevsk (commented):
Generally looks fine to me, at the level of the devfile/api, as long as the generator still works and outputs the same schemas, we should be safe to make this change.

build-generator.sh
thepetk: @amisevsk thanks for the hint :)

I've followed the same workaround proposed here (also the reason of the error I was experiencing): https://github.com/golang/mock/issues/415#issuecomment-602684505 which is a similar case of upgrading from `1.13` to `> 1.13`.

My issue previously was that when I was trying to run `go generate ./...` I was receiving the following error:
```
cannot find package "." in:
        /home/tpetkos/github.com/thepetk/devfiles/devfile-api/generator/vendor/sigs.k8s.io/controller-tools/cmd/helpgen
crds/gen.go:34: running "go": exit status 1
cannot find package "." in:
        <my-path>/generator/vendor/sigs.k8s.io/controller-tools/cmd/helpgen
getters/gen.go:31: running "go": exit status 1
cannot find package "." in:
       <my-path>/generator/vendor/sigs.k8s.io/controller-tools/cmd/helpgen
interfaces/gen.go:33: running "go": exit status 1
cannot find package "." in:
        <my-path>/generator/vendor/sigs.k8s.io/controller-tools/cmd/helpgen
overrides/gen.go:41: running "go": exit status 1
cannot find package "." in:
        <my-path>/generator/vendor/sigs.k8s.io/controller-tools/cmd/helpgen
schemas/gen.go:46: running "go": exit status 1
cannot find package "." in:
        <my-path>/generator/vendor/sigs.k8s.io/controller-tools/cmd/helpgen
validate/gen.go:32: running "go": exit status 1
```

As suggested in the similar case, the `-mod=mod` flag is fixing that issue and we are able again to use `go generate ./...`

amisevsk: Hmm. I'm concerned that using `-mod=mod` makes the vendor directory meaningless (if I understand correctly, this will direct go to _not_ use our vendor directory, defeating the purpose of having it).

Looking at the error, it seems to be coming from controller-tools. I found the following issue that looks to be the same problem -- https://github.com/kubernetes-sigs/controller-tools/issues/327 -- though it was closed due to inactivity, not due to an actual fix.

thepetk: @amisevsk getting back to you after I've tried various stuff while changing the codebase the least possible (having in mind that the scope of this issue is the upgrade of the go version).

I think the issue you've mentioned is very similar with our issue here. What I've noticed is that we cannot vendor the `cmd/helpgen` directory as it is under `package main` so upon `go mod vendor` we only vendor the `pkg` directory. As a result we cannot find the `helpgen` module we are trying to run the `go:generate` commands we have across the `generator` directory.

What I've tried is to do is to add the `-mod=mod` flag while using a fixed version of helpgen (same with the one that the API is using). It might not be the best approach, but I thought maybe we could create a separate issue to resolve this in a better way in the future (as the `generator` atm uses a very old version of go). WDYT?

amisevsk: I suppose it's the best we can do, makes sense.

amisevsk (approved):

michael-valdron (approved):
/lgtm

thepetk:
I've created issue https://github.com/devfile/api/issues/1359 to investigate any potential workaround regarding the usage of `vendor` dir during `go generate`. cc @amisevsk

Issue #1341, opened 2023-11-09
Implement podman support for devfile/api


Jdubrick:
/kind user-story

## Which area this user story is related to?

/area api

## User Story

Issue opened as a result of https://github.com/devfile/api/issues/1334 for implementing podman support across our repositories.

As a contributor I want to be able to run the build scripts using podman, so that I can develop for devfile with the use of podman instead of docker.

https://github.com/devfile/registry-support currently supports podman and the `setenv.sh` script may be of use for solving the issue for devfile/api.

## Acceptance Criteria

- [ ] Implement podman support for devfile/api build and run scripts.
- [ ] Update documentation to reflect changes in the build/deploy process.
- [ ] Sufficiently test changes made.

PR #1357, opened 2023-11-20
Implementation for podman to be used for container images

Jdubrick:
## What does this PR do?

This PR enables the use of podman for container images instead of explicitly docker.

resolves https://github.com/devfile/api/issues/1341 

### PR acceptance criteria

> - Check each criteria if:
>  **or**

- [ ] Implement podman support for devfile/api build and run scripts.
- [ ] Update documentation to reflect changes in the build/deploy process.
- [ ] Sufficiently test changes made.

### How to test changes / Special notes to the reviewer
Run `bash ./docker-run.sh ./build.sh` with `export USE_PODMAN=true` for podman testing and `export USE_PODMAN=false` for docker.


README.md
```diff
@@ -34,8 +34,8 @@ Release details and process are found in [Devfile Release](RELEASE.md)
 
 ## How to build
```
yangcao77, line 35: this section is duplicating the content in contributing doc. We shouldn't spent effort to maintain same content separate doc. Please use the link to contributing.md, e.g. `./CONTRIBUTING.md#building`

yangcao77 (commented):
generally looks good, left a comment on the doc

README.md
thepetk: I agree about the duplication :) Thinking that in general the contribution guidelines are more related to information about how someone could open a PR or an issue, I was wondering if the `contributing.md` is the right file to maintain this section? WDYT?

Jdubrick: Thank you for pointing that out! To remove the duplication I agree with contributing.md probably not being the right spot for the build instructions, I'll remove it from contributing and keep in in the `readme.md` only

Jdubrick: @yangcao77 @thepetk please see the latest commit for my updated changes.

docker-run.sh
```diff
@@ -23,6 +26,9 @@ WORKDIR="/projects/src/${GO_MODULE}"
 # Container image
 IMAGE_NAME="quay.io/devfile/kubernetes-api-build-prerequisites:latest"
 
+# Run script to set env var for podman if necessary
+. ./setenv.sh
+
```
amisevsk, line 31: I'm not sure I like hiding the alias in a separate script and controlled by an env var -- it's hiding a lot of functionality here (if you don't notice these two lines at the top it looks like it just uses docker).

setenv.sh
```diff
@@ -0,0 +1,25 @@
...
+# limitations under the License.
+
+# This script aliases the docker cli if the environment variable USE_PODMAN is set to true.
+
+# default value is false if USE_PODMAN is unset or null
+podman=${USE_PODMAN:-false}
+if [ ${podman} == true ]; then
+  alias docker=podman
+  echo "setting alias docker=podman"
+fi
```
amisevsk, line 25: Why not inline this?

amisevsk (approved):

README.md
yangcao77: should directly link to the `building` section. not the top of the contributing doc

setenv.sh
yangcao77: I do have the same question, why using a separate script just for setting alias?

Jdubrick: The reasoning for having it in a separate script was only because I was going off of prior work done [here](https://github.com/devfile/registry-support) by someone else. That repo utilizes the same `setenv.sh` script being used alongside an env variable to toggle between podman or docker.

If it's preferred to be inline that change can definitely be made.

amisevsk: The way I've usually done it in other repos is to just use a variable, e.g.
```
DOCKER=docker
if [ <use podman condition> ]; then
  DOCKER=podman
fi

$DOCKER build -t <image> -f <file> <context>
```
this also avoids the need for setting `shopt -s expand_alias`.

Jdubrick: In my last commit I added the setting of the alias inline so we don't need another script to be doing that work. I initially tried to go with your way @amisevsk with the use of a variable but found that when it was trying to call the `./build.sh` script it was failing with a seg fault, this was not happening with setting the alias. I'm not sure why so more investigation will have to be done into that if we want to not use an alias and rather a variable.

yangcao77 (approved):
look good to me. thx for the changes

PR #1614, opened 2024-07-08
fix: bump operations per stale run by 2x

Jdubrick:
## What does this PR do?
This PR bumps the operations allowed by the stale workflow run from `30 -> 60`. The logs are showing that we aren't moving through the stale issues as quickly as desired because we are hitting the operation limit on every run: https://github.com/devfile/api/actions/runs/9831302610/job/27138492511#step:2:456

The default value is `30` that's why we need to add the new field to bump it. This value may need to increase further if we determine `60` is insufficient

N/A
### PR acceptance criteria

> - Check each criteria if:
>  **or**

- [ ] Unit/Functional tests

  

- [ ] Documentation

   

- [ ] Client Impact


michael-valdron (approved):
/lgtm

PR #1666, opened 2024-12-23
fix: compile nmp library as 'commonjs' module

vitaliy-guliy:
# Description of Changes

Compile generated `@devfile/api` node package as `commonjs` module instead of `node:module`.

It was around two years ago when the `@devfile/api` package was published as `commonjs` module https://www.npmjs.com/package/@devfile/api/v/2.2.1-alpha-1667236163?activeTab=code

Since that time all the node libraries generated by openapi-generator are being compiled as `node:module`.
( go to https://www.npmjs.com/package/@devfile/api?activeTab=code, open `package.json` and look at `type` property )

Both Che Dashboard and Che Code projects are compiled as `commonjs`.

There is a well known restriction, that `node:modue` cannot be included into `commonjs` module using `import`, `require` should be used instead.

Che Dashboard and Che Code projects both use `@devfile/api`. When we package dashboard and che-code with webpack, we do not have any issues ( btw, this is a reason why we didn't have the problem for a long time). But when we try to compile the project without using webpack and then run it in a development mode (browser downloads a thousand of separate files), we face a problem with loading some files from `@devfile/api`.

The original issue highlighting a problem https://github.com/eclipse-che/che/issues/23163
A workaround after which I decided to open this PR https://github.com/che-incubator/che-code/pull/472

# Tests Performed
_Explain what tests you personally ran to ensure the changes are functioning as expected._

# How To Test
_Instructions for the reviewer on how to test your changes._

# Notes To Reviewer
_Any notes you would like to include for the reviewer._

### DIFF
--- a/build/typescript-model/generate.sh
+++ b/build/typescript-model/generate.sh
@@ -53,12 +53,49 @@ EOF
     export OPENAPI_GENERATOR_COMMIT="v6.3.0"
     bash $WORK_DIR/gen/openapi/typescript.sh $WORK_DIR/typescript-models $WORK_DIR/config.sh
 
-    apply_sed 's/\"name\": \".*\"/"name": "@devfile\/api"/g' $WORK_DIR/typescript-models/package.json
-    apply_sed 's/\"description\": \".*\"/"description": "Typescript types for devfile api"/g' $WORK_DIR/typescript-models/package.json
-    apply_sed 's/\"repository\": \".*\"/"repository": "devfile\/api"/g' $WORK_DIR/typescript-models/package.json
-    apply_sed 's/\"license\": \".*\"/"license": "Apache-2.0"/g' $WORK_DIR/typescript-models/package.json
-    apply_sed 's/\"@types\/bluebird\": \".*\"/"@types\/bluebird": "3.5.21"/g' $WORK_DIR/typescript-models/package.json
-    
+    local workdir=$(pwd)
+    cd "$WORK_DIR/typescript-models"
+
+    ######################################################################################################
+    echo "[INFO] preparing package.json"
+    ######################################################################################################
+
+    echo "$(jq '. += {"name": "@devfile/api"}' package.json)" > package.json
+    echo "$(jq '. += {"description": "Typescript types for devfile api"}' package.json)" > package.json
+    echo "$(jq '.repository += {"url": "https://github.com/devfile/api"}' package.json)" > package.json
+    echo "$(jq '. += {"homepage": "https://github.com/devfile/api/blob/main/README.md"}' package.json)" > package.json
+    echo "$(jq '. += {"license": "Apache-2.0"}' package.json)" > package.json
+
+    echo "$(jq 'del(.main)' package.json)" > package.json
+    echo "$(jq 'del(.type)' package.json)" > package.json
+    echo "$(jq 'del(.module)' package.json)" > package.json
+    echo "$(jq 'del(.exports)' package.json)" > package.json
+    echo "$(jq 'del(.typings)' package.json)" > package.json
+
+    echo "$(jq '. += {"main": "dist/index.js"}' package.json)" > package.json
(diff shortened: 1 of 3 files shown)


build/typescript-model/generate.sh
```diff
@@ -53,12 +53,49 @@ EOF
     export OPENAPI_GENERATOR_COMMIT="v6.3.0"
     bash $WORK_DIR/gen/openapi/typescript.sh $WORK_DIR/typescript-models $WORK_DIR/config.sh
 
-    apply_sed 's/\"name\": \".*\"/"name": "@devfile\/api"/g' $WORK_DIR/typescript-models/package.json
-    apply_sed 's/\"description\": \".*\"/"description": "Typescript types for devfile api"/g' $WORK_DIR/typescript-models/package.json
-    apply_sed 's/\"repository\": \".*\"/"repository": "devfile\/api"/g' $WORK_DIR/typescript-models/package.json
-    apply_sed 's/\"license\": \".*\"/"license": "Apache-2.0"/g' $WORK_DIR/typescript-models/package.json
-    apply_sed 's/\"@types\/bluebird\": \".*\"/"@types\/bluebird": "3.5.21"/g' $WORK_DIR/typescript-models/package.json
```
dkwon17, line 60: Do we know if it's okay to remove
```
    apply_sed 's/\"@types\/bluebird\": \".*\"/"@types\/bluebird": "3.5.21"/g' $WORK_DIR/typescript-models/package.json
```
?

vitaliy-guliy: I could have deleted it by accident.
I will double check it.

vitaliy-guliy: I recalled why did I remove it at all, there was no dependency on `bluebird` in `package.json`.

To check, go to https://www.npmjs.com/package/@devfile/api?activeTab=code and open existing `package.json`

dkwon17 (approved):

svor (approved):
lgtm

build/typescript-model/generate.sh
```diff
@@ -53,12 +53,49 @@ EOF
...
-    
+    local workdir=$(pwd)
+    cd "$WORK_DIR/typescript-models"
+
+    ######################################################################################################
+    echo "[INFO] preparing package.json"
+    ######################################################################################################
+
+    echo "$(jq '. += {"name": "@devfile/api"}' package.json)" > package.json
```
michael-valdron, line 63: With the addition of `jq` to this script I wonder if it makes sense to include it to the list of [prerequisites](https://github.com/devfile/api/blob/main/CONTRIBUTING.md#prerequisites)? Also `node` though I feel like this should of been there before for generating typescript models.

michael-valdron (commented):
@vitaliy-guliy Thank you for your contribution!

Just have one suggestion about adding to the documentation, looks good to me otherwise.

build/typescript-model/generate.sh
vitaliy-guliy: If we speak about the action that builds and then publishes the typescript models, it runs on ubuntu and there is a dedicated section to install node https://github.com/devfile/api/blob/main/.github/workflows/release-typescript-models.yaml#L48-L53. So no issues with it.

If you want to build typescript models locally, you definitely need to have `jq` and `node` installed.
I think it make sense to add both `jq` and `node` to the list of prerequisites (I'll do it)

Btw, typescript models are generated in a separate container. Node is used then to build and publish the generated library.

michael-valdron: @vitaliy-guliy Sounds good :+1:

vitaliy-guliy: @michael-valdron I've added two items to the prerequisites. Could you take a look?

vitaliy-guliy: And please merge the PR if everything's OK.
I don't have enough rights to do this.

michael-valdron (approved):
/lgtm

PR #1711, opened 2025-05-05
fix (utils) : [CRW-8603] Use slice of string instead of apimachinery Set to merge events

rohanKanojia:
# Description of Changes
We were previously using [apimachinery Set](https://github.com/kubernetes/apimachinery/blob/master/pkg/util/sets/string.go#L25) that is internally based on Map, an unordered data structure. Due to this we were losing the original insertion order defined in the devfile while merging it, `List()` call returns the elements in non-deterministic order, as it internally ranges over the map.

Use string slice to preserve insertion order of all event types.

# Related Issue(s)
https://issues.redhat.com/browse/CRW-8603

# Acceptance Criteria

_Testing and documentation do not need to be complete in order for this PR to be approved. However, tracking issues must be opened for missing testing/documentation._

New testing and documentation issues can be opened under [`devfile/api/issues`](https://github.com/devfile/api/issues).

You can check the respective criteria below if either of the following is true:
- There is a separate tracking issue opened and that issue is linked in this PR.
- Testing/documentation updates are contained within this PR.

If criteria is left unchecked please provide an explanation why.

- [ ] Unit/Functional tests

  

- [ ] Documentation

   

- [ ] Client Impact

  

# Tests Performed
I have added unit tests that verify that insertion order is preserved.
Before these changes, for a devfile that contained postStart events in order like this:
```yaml
    - "event-one"
    - "event-two"
    - "event-three"
```
After merging, this order was altered like this:
```yaml
    - "event-one"
    - "event-three"
    - "event-two"
    - "init-che-code-command"
```
After making these changes, order is preserved (check unit test files)

```yaml
   - "event-one"
    - "event-two"
    - "event-three"
    - "init-che-code-command"
```
# How To Test
You can run unit tests that are added for this change.

# Notes To Reviewer
This change is required for a fix in devworkspace operator for an issue where postStart event order is different than specified: https://issues.redhat.com/browse/CRW-8603

While debugging, I found that we're using [MergeDevWorkspaceTemplateSpec](https://github.com/devfile/api/blob/22740cbc69a8bce15dbc6572db9a78002f261c6b/pkg/utils/overriding/merging.go#L40) and it seems to be changing the order of `postStart` events.

### DIFF
--- a/pkg/utils/overriding/merging.go
+++ b/pkg/utils/overriding/merging.go
@@ -108,19 +108,19 @@ func MergeDevWorkspaceTemplateSpec(
 		}
 	}
 
-	preStartCommands := sets.String{}
-	postStartCommands := sets.String{}
-	preStopCommands := sets.String{}
-	postStopCommands := sets.String{}
+	var preStartCommands []string
+	var postStartCommands []string
+	var preStopCommands []string
+	var postStopCommands []string
 	for _, content := range allContents {
 		if content.Events != nil {
 			if result.Events == nil {
 				result.Events = &dw.Events{}
 			}
-			preStartCommands = preStartCommands.Union(sets.NewString(content.Events.PreStart...))
-			postStartCommands = postStartCommands.Union(sets.NewString(content.Events.PostStart...))
-			preStopCommands = preStopCommands.Union(sets.NewString(content.Events.PreStop...))
-			postStopCommands = postStopCommands.Union(sets.NewString(content.Events.PostStop...))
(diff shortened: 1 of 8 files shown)


michael-valdron (approved):
/lgtm

Built successfully and testing passed:
Screenshot From 2025-05-06 15-57-00
Screenshot From 2025-05-06 15-51-19
Screenshot From 2025-05-06 15-54-48

thepetk (approved):
/lgtm

rohanKanojia:
Thanks a lot for the review! Is it possible to merge it?

